PR #1, opened 2016-09-13
TANGO-678 Migrate to cmake

Ingvord:
In progress


taurel:
Igor,

Some comments:
- At ESRF, we are still using a lot of Debian 7 host where the default cmake release is 2.8. May be it's better to have a minimum cmake version required to 2.8

- I haven't seen the compiler option for c++11

- I think we need a way to specify where is omniORB or ZMQ

Ingvord:
Hi Emmanuel,

Sure we can set minimum version to 2.8;

C++11 will be added;

you mean some external variable, like --omiorb-lib-path ?

PR #265, opened 2016-09-26
remove zmq.hpp

Ingvord:


bourtemb:
It seems this should fix #266,  the request from Frederic Picca who noticed that tango zmq.hpp was superseeding the zmq.hpp from the system.

PR #278, opened 2016-10-07
Refactoring 1

Ingvord:
PR to move all the corba related code in a single module, i.e. directory src/corba


Ingvord:
Looks like this one is a dead end. Aborting.

PR #290, opened 2016-10-19
Refactoring readers writers lock

Ingvord:


include/tango/server/readers_writers_lock.h
```diff
@@ -34,83 +34,57 @@
...
+    condition_variable writers;
+    atomic_int totalReaders;
+    atomic_int totalWriters;
+public:
+    ReadersWritersLock(void) : totalReaders(0), totalWriters(0) {}
 
     void readerIn(void) {
-        thread::id threadId{this_thread::get_id()};
-
+        if(totalWriters.load() > 0 && !::isWriter)
```
Ingvord: Strangely enough but reading `::isWriter` in release mode always returns false, i.e. forces the thread which already has a WriterLock to wait forever. Looks like _::variable_ does not refer to accessing variables in the unnamed namespaces, but rather smth totally different, see, for instance, [this post on SO](http://stackoverflow.com/questions/357404/why-are-unnamed-namespaces-used-and-what-are-their-benefits)

Issue #301, opened 2016-10-19
Swapped origin and desc in some versions of Tango::Except::throw_exception.


tango-tickets-migrator:
In except.h, origin and desc are sometimes swapped when being copied to the DevErrorList entry. For example, lines 490-491 are
`errors[0].origin = CORBA::string_dup(desc);`
`errors[0].desc = CORBA::string_dup(origin.c_str());`
The same occurs also on lines 697, 844, 1262, 1406, 1481 and 1636.

Regards,
Josef

Reported by: jcibulka

Original Ticket: [tango-cs/bugs/822](https://sourceforge.net/p/tango-cs/bugs/822)

Ingvord:
Fixed in #305

PR #304, opened 2016-10-27
Fix bug #301 (SF822)

bourtemb:
Swapped origin and desc in some versions of Tango::Except::throw_exception.


bourtemb:
Closed because of wrong commit message...

Issue #235, opened 2016-04-14
API_AsynReplyNotArrived exception thrown when asyn call response is received in the last 20 ms before the timeout expiration


tango-tickets-migrator:
Hi,

The API_AsynReplyNotArrived exception is wrongly thrown when an asynchronous call with timeout (like command_inout_reply(long id, long call_timeout) response is received in the last 20 ms before the call timeout expiration.

Cheers,
Reynald

Reported by: bourtemb

Original Ticket: [tango-cs/bugs/793](https://sourceforge.net/p/tango-cs/bugs/793)

Ingvord:
Please do not create a separate issue for such simple cases (e.g. fix that requires single commit) - use PR directly. Otherwise we just have duplicate entities.

tango-tickets-migrator:
This issue was automatically migrated from a Sourceforge ticket.

PR #306, opened 2016-10-27
Fix bug #235(SF793)

bourtemb:
API_AsynReplyNotArrived exception was wrongly thrown when asyn call
response was received in the last 20ms before the call timeout expiration.

### DIFF
--- a/src/client/proxy_asyn.cpp
+++ b/src/client/proxy_asyn.cpp
@@ -574,18 +574,15 @@ DeviceData Connection::command_inout_reply(long id,long call_timeout)
 #endif
 		}
 
-		if (i == nb)
+		if(i == nb && !req.request->poll_response())
 		{
-			if (req.request->poll_response() == false)
-			{
-				TangoSys_OMemStream desc;
-				desc << "Device " << dev_name();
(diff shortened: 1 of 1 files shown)


Ingvord:
Please use "close #<smth>" in PR comment or commit to allow GitHub automatically close corresponding issues.

src/client/proxy_asyn.cpp
```diff
@@ -576,13 +576,16 @@ DeviceData Connection::command_inout_reply(long id,long call_timeout)
...
 		if (i == nb)
 		{
-			TangoSys_OMemStream desc;
-			desc << "Device " << dev_name();
-			desc << ": Reply for asynchronous call (id = " << id;
-			desc << ") is not yet arrived" << ends;
-			ApiAsynNotThereExcept::throw_exception(API_AsynReplyNotArrived,
-						       	       desc.str(),
-						               "Connection::command_inout_reply");
+			if (req.request->poll_response() == false)
```
Ingvord: Nested if can be moved to outer, i.e.

```#!c++

if(i == nb && !req.request->poll_response())

```

PR #308, opened 2016-11-09
Fix bug SF828

bourtemb:
(Crash when a client receives attribute configuration events with non empty
enum labels field)
Original ticket: https://sourceforge.net/p/tango-cs/bugs/828/

### DIFF
--- a/src/client/api_util.cpp
+++ b/src/client/api_util.cpp
@@ -1845,9 +1845,10 @@ AttributeInfoEx &AttributeInfoEx::operator=(const AttributeConfig_5 *att_5)
 		else
 			memorized = MEMORIZED_WRITE_INIT;
 	}
-	for (unsigned int j=0; j<att_5->enum_labels.length(); j++)
+	enum_labels.clear();
+	for (size_t j = 0, size = att_5->enum_labels.length(); j < size; ++j)
 	{
-		enum_labels[j] = att_5->enum_labels[j];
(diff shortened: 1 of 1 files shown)


src/client/api_util.cpp
```diff
@@ -1847,9 +1847,10 @@ AttributeInfoEx &AttributeInfoEx::operator=(const AttributeConfig_5 *att_5)
 		else
 			memorized = MEMORIZED_WRITE_INIT;
 	}
+	enum_labels.clear();
 	for (unsigned int j=0; j<att_5->enum_labels.length(); j++)
 	{
-		enum_labels[j] = att_5->enum_labels[j];
+		enum_labels.push_back(att_5->enum_labels[j].in());
```
bourtemb: Good idea but enum_labels is a vector<string> but att_5->enum_labels is a DevVarStringArray, not a vector<string>.

src/client/api_util.cpp
```diff
@@ -1847,9 +1847,10 @@ AttributeInfoEx &AttributeInfoEx::operator=(const AttributeConfig_5 *att_5)
 		else
 			memorized = MEMORIZED_WRITE_INIT;
 	}
+	enum_labels.clear();
 	for (unsigned int j=0; j<att_5->enum_labels.length(); j++)
```
bourtemb: Please note that there are many lines of code to change in order to follow this standard naming pattern. I changed this for loop line but I won't change all the other similar lines in this pull request.

PR #311, opened 2016-11-22
Update tango.pc.cmake

Ingvord:
Add default device installation directory variable

### DIFF
--- a/tango.pc.cmake
+++ b/tango.pc.cmake
@@ -2,6 +2,7 @@ prefix = @CMAKE_INSTALL_PREFIX@
 exec_prefix = ${prefix}
 includedir = ${prefix}/include
 libdir = ${prefix}/@CMAKE_INSTALL_LIBDIR@
+devices_dir = ${prefix}/@CMAKE_INSTALL_LIBDIR@/tango
 
 Name: @CMAKE_PROJECT_NAME@
 Description: TANGO client/server API library
--- a/tango.pc.cmake
+++ b/tango.pc.cmake
@@ -2,7 +2,7 @@ prefix = @CMAKE_INSTALL_PREFIX@
 exec_prefix = ${prefix}
 includedir = ${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
 libdir = ${prefix}/@CMAKE_INSTALL_LIBDIR@
-ds_default_dir = @DS_DEFAULT_DIR@
+tangodsdir = @TANGO_DEVICE_SERVER_PATH@
 
 Name: @CMAKE_PROJECT_NAME@
 Description: TANGO C++ client/server API library
--- a/tango.pc.cmake
+++ b/tango.pc.cmake
@@ -1,11 +1,11 @@
 prefix = @CMAKE_INSTALL_PREFIX@
 exec_prefix = ${prefix}
-includedir = ${prefix}/include
+includedir = ${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
 libdir = ${prefix}/@CMAKE_INSTALL_LIBDIR@
-devices_dir = ${prefix}/@CMAKE_INSTALL_LIBDIR@/tango
+ds_default_dir = @DS_DEFAULT_DIR@
 
 Name: @CMAKE_PROJECT_NAME@
-Description: TANGO client/server API library
+Description: TANGO C++ client/server API library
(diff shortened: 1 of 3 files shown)


tango.pc.cmake
```diff
@@ -2,10 +2,11 @@ prefix = @CMAKE_INSTALL_PREFIX@
 exec_prefix = ${prefix}
 includedir = ${prefix}/include
 libdir = ${prefix}/@CMAKE_INSTALL_LIBDIR@
+devices_dir = ${prefix}/@CMAKE_INSTALL_LIBDIR@/tango
```
bourtemb: If I understood correctly, this new variable will be used to know the default installation directory for device servers.
If this is the case, I would suggest to call it dserver_default_dir or ds_default_dir or something similar.
It would probably make sense to install them by default under ${prefix}/@CMAKE_INSTALL_BINDIR@ or ${prefix}/@CMAKE_INSTALL_BINDIR@/tango
But unless I misunderstood the purpose of this variable (very probable), I would suggest to avoid to install the device servers anywhere under ${prefix}/@CMAKE_INSTALL_**LIBDIR**@

Ingvord: We need more info from Frederic. I believe `${prefix}/@CMAKE_INSTALL_LIBDIR@/tango` is what is used by default. `dserver` sounds for me as a admin device... anyway lets wait for Frederic's comment

hardion: For your information the path where is installed the device server is different between deb and rpm. 
For the hundreds of rpm we have the default path is: /usr/bin

Will it be possible to easily change the default directory following the platform?

Ingvord:
Now default device servers dir can be set via cmake variable: `cmake .. -DDS_DEFAULT_DIR=/usr/bin`

If not set default value is `${CMAKE_INSTALL_FULL_LIBDIR}/tango/server`

picca:
Hello guyes,

In fact it is better if device_dir was renamed devicedir autotools convention is to append dir without the _
this way you can write

<target>dir = /usr/bin

And in the Makefile.am
<target>_bin = ...


The objectif of this request was to avoid polution of the /usr/bin directory with hundreds of binaries which should be started via the Starter device. (I took the git binaries has example, before tens of git-xxx) and now only one git binary command.

Another important point is to have an automatic way to know the device path expected by the installed tango, so make install works out of the box for a third party software.


I agree with vincent hardion that the default should be /usr/bin if you think that it is better.
So mauybe make it simple and provide only one CMAKE_DEVICE_PATH variable and do not use the LLIB directory.

It is then the repsonsability of the package maintainer to customise this.

It should be nice also to document this devicedir varaiable available from the pkg-config dir :)

Cheers

bourtemb:
Hello,
I still disagree with the name _devicedir_ because it is confusing. A device means something different in the Tango world and we should avoid to confuse people.
I think _dsdir_, _dsdefaultdir_ or _deviceserversdir_ would be less confusing in tango.pc file.
I agree with Frédéric about the fact that it would be nice to document about the availability of this variable.
I actually don't know whether there is already any documentation on this subject.
I have never seen any documentation related to tango.pc file.
Cheers

picca:
tangodeviceserversdir :)

Ingvord:
deafultdirfotallmytangodeviceserversinstalledusingthispackageconfigfile

Guys stop it!

Ingvord:
In fact everyone whom I know says "device" meaning tango server executable and "device server" usually refers to dserver, i.e. admin device.

Ingvord:
Other thing I did not quite understood - documenting this variable - means simple add a description of the .pc file to README, or there is some mechanism provided by pkg-config, like list all variables? Another question - maybe we should mark Requires as private, i.e. `Requires.private omniDynamic4 >= 4.1.6, omniCOS4, libzmq`

picca:
Look at here, It seems to me that the variables are prepended with the package name space in order to avoid clash between projects.

https://github.com/systemd/systemd/blob/dd8352659c9428b196706d04399eec106a8917ed/src/core/systemd.pc.in

Cheers

Ingvord:
OK, `tangodsdir` seems to be reasonable choice then.

picca:
just explaining for third party peoples how to obtain the tangodevicesdir ;)
via the pkg-config

pkg-config --variable=tangodevicesdir tango
/usr/lib/tango

on Debian system.

PR #325, opened 2016-12-07
Generates files from IDL only at compilation time and only if necessary

bourtemb:
(Before it was done when executing cmake, leading to tango.h which could be
patched several times and causing compilation errors).
The files generated from the IDL are now generated in CMAKE_CURRENT_BINARY_DIR
because they could depend on the version of omniORB used by the current config.
CMake minimum required version is now 3.0 instead of 2.8. (CMake was failing
with 2.8.11 version).
Added install_libtango custom target to be able to test the installation part
from CLion IDE.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.0)
 project(libtango)
 include(CTest)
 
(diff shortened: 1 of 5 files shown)


Ingvord:
Nice solution! Much cleaner. Definitely the same should be done for the master.

P.S. please try not to do more than one thing in a PR. For instance, install target could be done in a separate one.

bourtemb:
Right you are! I'll try to refrain from doing several things in a single PR. Sorry for that.
About doing the same on the master, I put this on my todo list (quite busy with other taks right now), unless you want to do it. ;-)

Ingvord:
Looks like due to this tests may now accidentally fail. As make is normally executed in parallel mode and travis always builds from scratch it sometimes can not find generated tango.h. A minor issue, but annoying.

bourtemb:
If tango.h can not be found during parallel make compilation, this means that some dependencies definitions are missing somewhere. During the build, make should not try to build something which requires tango.h if tango.h was not yet generated... We need to fix that indeed.

Issue #328, opened 2016-11-08
Crash in local call


tango-tickets-migrator:
Hello. We've stumbled upon another crash that happens with a local-call from a thread - not unlike my previous bug (823). I applied the patch to blackbox.cpp, but now I get a crash that does not go through the blackbox:

~~~
#0 0x00007ffff1e1ce62 in omni_thread::get_value (this=0x0, k=2) at threaddata.cc:71
#1 0x00007ffff3e9defa in Tango::DeviceImpl::get_client_ident (this=<optimized out>) at device.cpp:4416
#2 0x00007ffff3ea5255 in Tango::DeviceImpl::check_lock (this=this@entry=0x2007850, meth=meth@entry=0x7ffff40991a5 "command_inout4",
cmd=cmd@entry=0x7fff94005eb8 "Stop") at device.cpp:4820
#3 0x00007ffff3ed3ee2 in Tango::Device_4Impl::command_inout_4 (this=0x2007850, in_cmd=0x7fff94005eb8 "Stop", in_data=..., source=Tango::CACHE_DEV, cl_id=...)
at device_4.cpp:467
#4 0x00007ffff404f342 in _0RL_lcfn_6fe2f94a21a10053_a3000000 (cd=0x7fffa2ff8dc0, svnt=<optimized out>) at tangoSK.cpp:5383
#5 0x00007ffff38bd333 in doLocalCall (servant=<optimized out>, this=0x7fffa2ff8dc0) at ../../../../include/omniORB4/callDescriptor.h:145
#6 omni::omniOrbPOA::dispatch (this=<optimized out>, call_desc=..., id=0x26924b0) at poa.cc:1852
#7 0x00007ffff389e6ee in omniLocalIdentity::dispatch (this=0x26924b0, call_desc=...) at localIdentity.cc:145
------------SNIP--------------
#142 0x00007ffff78180db in start_thread () from /lib64/libpthread.so.0
#143 0x00007ffff7548e3d in clone () from /lib64/libc.so.6
~~~

Johan here https://sourceforge.net/p/tango-cs/bugs/814/#da32 seems to have the same problem.

Reported by: schneidemwe ( http://sf.net/u/schneidemwe )

Original Ticket: [tango-cs/bugs/827](https://sourceforge.net/p/tango-cs/bugs/827)

bourtemb:
Problem solved in Tango 9.2.5, even though the solution re-introduced a small memory leak.

Issue #302, opened 2016-10-24
Crash when executing a command from a thread on an in-process device


tango-tickets-migrator:
Using Tango 9.2.2:
Crash is a null-pointer access in Tango's blackbox.cpp:
omni_thread::value_t *ip = omni_thread::self()->get_value(key);

self() returns nullptr in this case.

To reproduce, let any command do something like this:

```
std::thread thread([this]() {
    std::string target = this->device_name;
    auto proxy = std::make_shared<Tango::DeviceProxy>(target);
    proxy->command_inout("status");
});

thread.join();
```

Reported by: schneidemwe

Original Ticket: [tango-cs/bugs/823](https://sourceforge.net/p/tango-cs/bugs/823)

bourtemb:
Fixed in Tango 9.2.5

Issue #247, opened 2016-08-16
Segmentation fault when reading attribute 


tango-tickets-migrator:
Another segmentation fault we found while upgrading to TANGO 9:

If a device reads an attribute from a device in _the same server_ (i.e. the same process) and from _within a thread_, the server segfaults. It does not happen for devices in other servers, nor if the read is not done from a thread.

The problem is easy to reproduce and has been observed in both a python device and a C++ device (included), so the issue should lie in libtango. 

We are using TANGO 9.2.2. We have not yet tried to reproduce with older v9 versions but this behavior was not present in v8.

```
Thread 10 "ThreadReadSegfa" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffeb7fe700 (LWP 4848)]
0x00007ffff6b33c22 in omni_thread::get_value(unsigned int) () from /usr/lib/libomnithread.so.3
(gdb) bt
#0  0x00007ffff6b33c22 in omni_thread::get_value(unsigned int) () from /usr/lib/libomnithread.so.3
#1  0x00007ffff78b3138 in Tango::BlackBox::insert_attr (this=0x69b490, names=..., cl_id=..., vers=vers@entry=5, 
    sour=sour@entry=Tango::CACHE_DEV) at blackbox.cpp:643
#2  0x00007ffff7916447 in Tango::Device_5Impl::read_attributes_5 (this=0x69db60, names=..., source=Tango::CACHE_DEV, 
    cl_id=...) at device_5.cpp:114
#3  0x00007ffff7aadcfb in _0RL_lcfn_6fe2f94a21a10053_84000000 (cd=0x7fffeb7fd080, svnt=<optimized out>)
    at tangoSK.cpp:6272
#4  0x00007ffff6e1a94f in omni::omniOrbPOA::dispatch(omniCallDescriptor&, omniLocalIdentity*) ()
   from /usr/lib/libomniORB4.so.1
#5  0x00007ffff6dff939 in omniLocalIdentity::dispatch(omniCallDescriptor&) () from /usr/lib/libomniORB4.so.1
#6  0x00007ffff6e0da65 in omniObjRef::_invoke(omniCallDescriptor&, bool) () from /usr/lib/libomniORB4.so.1
#7  0x00007ffff7aadfbf in Tango::_objref_Device_5::read_attributes_5 (this=this@entry=0x7fffd8000d10, names=..., 
    source=<optimized out>, cl_ident=...) at tangoSK.cpp:6298
[...]

PR #329, opened 2016-12-12
Revert changes introduced in svn r28853 on master

bourtemb:
This reintroduces a small memory leak but will avoid crashes reported in the
following tickets: SF814 (#247), SF823 (#302) and SF827 (#328)


Ingvord:
Hopefully this all gone once we switch to C++11 threads completely

PR #333, opened 2016-12-14
Improve AttributeProxy documentation for subscribe_event methods

### LINKED ISSUE #313: Improve AttributeProxy documentation for subscribe_event methods
The documentation of AttributeProxy subscribe_event methods should be improved in order to precise that the AttributeProxy object and the callback object passed when these methods are invoked should be alive until the events are no longer requested (until unsubscribe_event() is called) because only the pointer to these objects is stored and used by the event machinery.
This is already documented in DeviceProxy::subscribe_event() method but not yet in AttributeProxy::subscribe_event() methods.

bourtemb:
Fix #313

### DIFF
--- a/include/tango/client/AttributeProxy.h
+++ b/include/tango/client/AttributeProxy.h
@@ -464,13 +464,15 @@ namespace Tango {
 /**
  * Subscribe to attribute event
  *
- * The client call to subscribe for event reception in the pushmodel. The client implements a callbackmethod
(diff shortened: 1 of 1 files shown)


bourtemb:
Hum, I think there is a problem with the tests... This PR changed only comments. The tests should not fail.

Ingvord:
Yeah, there was segmentation fault in per_event test. Most probably threading issue... we have to keep our eyes on it open.

PR #334, opened 2016-12-15
Add cxx files to project

Ingvord:
Currently cxx test files are not tracked by CLion as they are not part of the cmake project. We need to fix this. Using cmake's [add_custom_command](https://cmake.org/cmake/help/v3.0/command/add_custom_command.html) may help, e.g. as in #326 .


Ingvord:
Test comment via UpSource

PR #335, opened 2016-12-15
Fix #312 and SF829: zmq 4.2.0 memory alignment

### LINKED ISSUE #312: Please adapt tango to zeromq 4.2.0
Hello guyes,

Here I put the information I have about this problem, here

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=844479

It seems to me that it will be difficult to keep zeromq < 4.2.0 for Debian 9.

Can you try to find a fix before the freeze of Debian ?

thanks

Fred

bourtemb:


bourtemb:
This is the equivalent of the following commit done on tango-9-lts, but for the master branch:
https://github.com/tango-controls/cppTango/pull/318/commits/5fde3d5cfe329b6576a972bd3f5d95b6aa43dde8

PR #336, opened 2016-12-15
Fix#314 master

bourtemb:


bourtemb:
This is the equivalent of the following commit from tango-9-lts branch for the master branch:
https://github.com/tango-controls/cppTango/pull/318/commits/e529990af23a117e1e8f9b71cad8382f4b9dfb5d

src/server/eventcmds.cpp
```diff
@@ -755,6 +755,37 @@ DevVarLongStringArray *DServer::zmq_event_subscription_change(const Tango::DevVa
         action = (*argin)[2];
         event = (*argin)[3];
 
+//
+// Check event type validity
```
Ingvord: Whenever there is a comment in the code describing what it is doing - extract a method.

src/server/eventcmds.cpp
```diff
@@ -755,6 +755,37 @@ DevVarLongStringArray *DServer::zmq_event_subscription_change(const Tango::DevVa
...
+        transform(check_event.begin(),check_event.end(),check_event.begin(),::tolower);
+
+        string::size_type pos_check = check_event.find(EVENT_COMPAT);
+        if (pos_check != string::npos)
+            check_event.erase(0,EVENT_COMPAT_IDL5_SIZE);
+
+        size_t nb_event_type = sizeof(EventName)/sizeof(char *);
+        bool found = false;
+
+        for (size_t loop = 0;loop < nb_event_type;loop++)
```
Ingvord: `loop` should be renamed to `i`

src/server/eventcmds.cpp
```diff
@@ -755,6 +755,37 @@ DevVarLongStringArray *DServer::zmq_event_subscription_change(const Tango::DevVa
...
+        {
+            if (strcmp(check_event.c_str(),EventName[loop]) == 0)
+            {
+                found = true;
+                break;
+            }
+        }
+
+        if (found == false)
```
Ingvord: Should be `if(!found)`

src/server/eventcmds.cpp
```diff
@@ -755,6 +755,37 @@ DevVarLongStringArray *DServer::zmq_event_subscription_change(const Tango::DevVa
...
+        {
+            stringstream ss;
+            ss << "The event type you sent (" << event << ") is not  valid event type";
+
+            Except::throw_exception(API_WrongNumberOfArgs,ss.str(),"DServer::zmq_event_subscription_change");
```
Ingvord: Can be simplified to 

```c++
Except::throw_exception(
    API_WrongNumberOfArgs,
    "The event type you sent (" + event +  ") is not  valid event type", //TODO name valid types
    "DServer::zmq_event_subscription_change");
```

bourtemb:
I'm sorry, I don't understand the following comment:

> Looks like it is an utility method that is useful for any C++ type enum. I would put it into a dedicated header file as a template utility function, smth like

```
template<typename Enum, typename EnumNames>
Enum enum_value_of(const std::string& name){
//TODO ...
}

```
Can you be more explicit, please? To what part of the code are you referring to for this comment?

Ingvord:
It was meant that the the added code should be extracted into a method. It is not urgent... just a remark for the future refactoring. 15.12.2016, 20:17, "Reynald Bourtembourg" <notifications@github.com>:I'm sorry, I don't understand the following comment:Looks like it is an utility method that is useful for any C++ type enum. I would put it into a dedicated header file as a template utility function, smth liketemplate<typename Enum, typename EnumNames>
Enum enum_value_of(const std::string& name){
//TODO ...
}

Can you be more explicit, please? To what part of the code are you referring to for this comment?—You are receiving this because you commented.Reply to this email directly, view it on GitHub, or mute the thread.   -- Ingvord

src/server/eventcmds.cpp
```diff
@@ -755,6 +755,37 @@ DevVarLongStringArray *DServer::zmq_event_subscription_change(const Tango::DevVa
...
         event = (*argin)[3];
 
+//
+// Check event type validity
+//
+// TODO Extract this as a method
+        string check_event = event;
+        transform(check_event.begin(),check_event.end(),check_event.begin(),::tolower);
+
+        string::size_type pos_check = check_event.find(EVENT_COMPAT);
```
Ingvord: `string::size_type` can be replaced with `auto`

PR #338, opened 2016-12-15
Extract method as discussed in #336

Ingvord:
subj.


bourtemb (commented):
It is very difficult to review this kind of changes since almost all the lines changed because you changed the coding style.
I think we should stick to one coding style. Usually, the convention is to re-use the coding style of the existing code in the project. One of the goals is precisely to simplify the review process and the maintenance because it's easier to compare the current version with older versions.
Of course we could decide to change the coding style in all the files of the master branch since there will be a major refactoring process for Tango 10.
Personally, I am not in favor of it because I will be involved in Tango 9 LTS maintenance. Having the same coding style in the 2 branches will simplify the maintenance of the project and the comparisons with older versions. 
Moreover, I personally prefer the existing coding style (Allman), as several of my colleagues as we discussed during a meeting at the ESRF.

Ingvord:
Sure, we need to define coding style. This can be forced via CLion, I know there is a possibility to create a special file which then can be used by everyone. But it seems as a lot of work for me, at least for now. Maybe after refactoring.

Anyway we must intend files with 4 spaces.

You can compare files ignoring white spaces and empty lines using CLion: rmb -> Git -> compare with branch -> ... 

See attachment - there are only two differences:

comparing_branches

bourtemb:
Thanks for the tip with CLion. CLion is actually very smart for generating this diff.
Clion seems to be smarter than GitHub and git diff command (unless I didn't find the right option to produce the same diff as the one we can see with CLion).
I'm still not convinced we should change the coding style since GitHub (and diff command, maybe) is not as smart as CLion.

Ingvord:
It is better to stick with CLion, because non-trivial things ain't intended to be reviewed via GitHub anyway (usually you want to clone it, run the tests, see the overall architecture etc).

Until we implement coding style policy for CLion I will stay (and suggest to do the same for everyone involved) with CLion's coding style (basically it is what you get when performing code arrangement with `ctrl+alt+L`), cauz' life is too short to spend it manually arranging lines :)

Issue #339, opened 2016-12-17
pointer-cast-size-mismatch attrgetsetprop.cpp:200


picca:
Hello,

while preparing the tango package of the 9.2.5~rc5, I got this error.

You can find other warning errors here, which are more packaging related.

https://qa.debian.org/bls/packages/t/tango.html

Cheers

Ingvord:
Is this urgent? I mean, is it possible to provide debian package with this error?

picca:
> Is this urgent? I mean, is it possible to provide debian package with this error?

I think that plenty of packages are distributed with bug :), so it is possibleto provide tango 9.2.5 with this bug.

Now I do not know if this can be fixed with a one liner for the final release of 9.2.5 in january.


Cheers


Fred

andygotz:
Needs to be done in LTS9 as well. If done then please close this issue. Thanks!

bourtemb:
Done on Tango 9 LTS too (PR #406).
Closing this issue.

PR #340, opened 2016-12-17
cppTango-#339: pointer-cast-size-mismatch attrgetsetprop.cpp:200

Ingvord:
progress #339: pointer-cast-size-mismatch attrgetsetprop.cpp:200


src/server/attrgetsetprop.cpp
```diff
@@ -198,7 +198,7 @@ void Attribute::get_properties(Tango::AttributeConfig_3 &conf)
 		conf.event_prop.per_event.period = Tango::string_dup((const char *)(DEFAULT_EVENT_PERIOD));
 	else
 	{
-		int per = (int)((double)event_period);
+		int per = event_period;
 		str << per;
```
bourtemb: We could even optimize by replacing lines 201 and 202 with:
`str << event_period;`

PR #341, opened 2016-12-19
Forwarded dynamic attributes implementation

jairomoldes:
Hi,

Using branch "tango-9-lts" of github cppTango I noticed that tango doesn't seem to properly manage the creation of forwarded attributes when they are dynamically created. In my opinion, there are 3 different scenarios in which forwarded attributes can be created:

1.- Static creation. This is what pogo does, for example. This seems to already be properly managed.
2.- Dynamic creation with root_attribute specified in constructor. In this case the root attribute used will be the one passed in the constructor (tango DB value, if defined, should be always ignored).
3.- Dynamic creation without root_attribute specified in constructor. In this case the root attribute will be the one defined in tango DB.

In cases 2 and 3 if root attribute is not correctly defined then the device will got to Alarm state and status will give a clue of what is wrong: all dynamic attributes will be created, but the ones with incorrect root attribute definition will throw an exception when read or written.

You can easily test the all above options using the attached TestFwd device server (ziped in "TestFwd.zip"). You will have to play with their root attribute definition (for example using jive) in order to reproduce all cases mentioned above. IT assumes that "sys/tg_test/1/double_scalar" is available. It defines 3 forwarded attributes:
- TestFwdAttr: this is a static attribute defined using pogo (it can be used for testing case 1 above).
- TestFwdDyn1: this is a dynamic attribute with root attribute defined in the constructor (it can be used for testing case 2 above).
- TestFwdDyn2: this is a dynamic attribute with root attribute not defined in the constructor (it can be used for testing case 3 above).

The patch seems to work fine in my case. I propose to merge it into tango-9-lts and master or use it as a first approach for developing a better patch (I'm far from a tango core expert, so I may have introduced other bugs while trying to solve this issue).

NOTE that it also fixes a bug I found when the dynamic forwarded attributes are created (this is the patch for "multiattribute.cpp" file): it looks like the write value is not correctly initialized and you can randomly get a coredump when reading them (you may need to restart the server several times to reproduce the error).

Regards,
Jairo Moldes

[TestFwd.zip](https://github.com/tango-controls/cppTango/files/661494/TestFwd.zip)

### DIFF
--- a/cppapi/server/device.cpp
+++ b/cppapi/server/device.cpp
@@ -3229,7 +3229,43 @@ void DeviceImpl::add_attribute(Tango::Attr *new_attr)
 
 	if (new_attr->is_fwd() == true)
 	{
-		dev_attr->add_fwd_attribute(device_name,device_class,i,new_attr);
+		// If forwarded attribute is dynamically created and was constructed without specifying
+		// the root_attribute parameter then we have to get its __root_att property from tango DB
+		// prior to calling validate_fwd_att()
+		vector<AttrProperty> dev_prop;
+		Tango::Util *tg = Tango::Util::instance();
+		Tango::FwdAttr *fwd_attr = static_cast<Tango::FwdAttr *> (new_attr);
+		if ((tg->_UseDb==true) && (fwd_attr->get_full_root_att()==RootAttNotDef))
+		{
+			Tango::DbData db_list;
+			db_list.push_back(DbDatum(new_attr->get_name()));
+			tg->get_database()->get_device_attribute_property(this->device_name,db_list,tg->get_db_cache());
+			for (unsigned int ind=0 ; ind<db_list.size() ; ind++)
+			{
+				if (db_list[ind].name == RootAttrPropName)
+				{
+					dev_prop.push_back(AttrProperty(db_list[ind].name, db_list[ind].value_string[0]));
+					break;
+				}
+			}
+		}
+		// Validate the __root_att property of the attribute
+		fwd_attr->validate_fwd_att(dev_prop, this->device_name);
+		// Register the root attribute configuration
+		try
+		{
+			fwd_attr->get_root_conf(this->device_name,this);
+		}
+		catch (...)
+		{
+			DeviceImpl::FwdWrongConf fwc;
+			fwc.att_name = attr_name;
+			fwc.full_root_att_name = fwd_attr->get_full_root_att();
+			fwc.fae = fwd_attr->get_err_kind();
+			fwd_att_wrong_conf.push_back(fwc);
+		}
+		// Add the attribute
+		dev_attr->add_fwd_attribute(device_name,device_class,i,fwd_attr);
 	}
 	else
 		dev_attr->add_attribute(device_name,device_class,i);
--- a/cppapi/server/fwdattrdesc.cpp
+++ b/cppapi/server/fwdattrdesc.cpp
@@ -180,7 +180,10 @@ bool FwdAttr::validate_fwd_att(vector<AttrProperty> &prop_list,const string &dev
 	}
 	catch (...) {}
 
-	if (root_att_db_defined == true)
+	//check if full_root_att is already set
+	if (full_root_att.size()!=0 and full_root_att.compare(RootAttNotDef)!=0)
+		;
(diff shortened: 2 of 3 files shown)


cppapi/server/device.cpp
```diff
@@ -3229,7 +3229,43 @@ void DeviceImpl::add_attribute(Tango::Attr *new_attr)
 
 	if (new_attr->is_fwd() == true)
 	{
-		dev_attr->add_fwd_attribute(device_name,device_class,i,new_attr);
+		// If forwarded attribute is dynamically created and was constructed without specifying
```
Ingvord: This code should be extracted into method `add_forwarded_attribute(Tango::Attr* fwd_attr)`

cppapi/server/fwdattrdesc.cpp
```diff
@@ -180,7 +180,10 @@ bool FwdAttr::validate_fwd_att(vector<AttrProperty> &prop_list,const string &dev
 	}
 	catch (...) {}
 
-	if (root_att_db_defined == true)
+	//check if full_root_att is already set
+	if (full_root_att.size()!=0 and full_root_att.compare(RootAttNotDef)!=0)
```
Ingvord: Never heard about `and` - is it macro?

Ingvord: This should be extracted into variable `bool is_full_root_att_set = ...`

cppapi/server/fwdattrdesc.cpp
```diff
@@ -180,7 +180,10 @@ bool FwdAttr::validate_fwd_att(vector<AttrProperty> &prop_list,const string &dev
...
+		;
+	else if (root_att_db_defined == true)
```
Ingvord: Should be `if(!is_full_root_att_set && root_att_db_defined)`

cppapi/server/device.cpp
```diff
@@ -3229,7 +3229,43 @@ void DeviceImpl::add_attribute(Tango::Attr *new_attr)
...
+		// the root_attribute parameter then we have to get its __root_att property from tango DB
+		// prior to calling validate_fwd_att()
+		vector<AttrProperty> dev_prop;
+		Tango::Util *tg = Tango::Util::instance();
+		Tango::FwdAttr *fwd_attr = static_cast<Tango::FwdAttr *> (new_attr);
```
Ingvord: Why `static_cast` is required here? Can this be done via polymorphism? Just wondering.

cppapi/server/fwdattrdesc.cpp
jairomoldes: Actually it looks like this is part of the C++ standard, but very rarely used:
http://stackoverflow.com/questions/4251699/is-it-okay-to-use-and-or-etc-instead-of

I can change it by && if you prefer.

Ingvord: Yes, please. This looks extremely odd.

cppapi/server/device.cpp
jairomoldes: I agree

cppapi/server/device.cpp
jairomoldes: A regular cast works. I'll change it for that. However I think that the cast is still necessary because I am calling FwdAttr specific methods that Attr is not aware of. Anyway if you have a better suggestion please let me know (I regularly program using python and I'm not a big expert on C++).

Ingvord: Please, use `dynamic_cast` in this case. This cast specifically designed for inherited structures, see [reference](http://en.cppreference.com/w/cpp/language/dynamic_cast).

cppapi/server/device.cpp
jairomoldes: I think that maybe its better to include this code in existing add_forwarded_attribute method

cppapi/server/fwdattrdesc.cpp
jairomoldes: ok

cppapi/server/fwdattrdesc.cpp
jairomoldes: ok

Ingvord:
I did not expect this PR to be closed. Strange...

PR #342, opened 2016-12-20
Forwarded dynamic attributes

jairomoldes:
Hi,

I rebuilt the code of the previous pull request. I hope its cleaner and clearer.
Using branch "tango-9-lts" of github cppTango I noticed that tango doesn't seem to properly manage the creation of forwarded attributes when they are dynamically created. In my opinion, there are 3 different scenarios in which forwarded attributes can be created:

1.- Static creation. This is what pogo does, for example. This seems to already be properly managed.
2.- Dynamic creation with root_attribute specified in constructor. In this case the root attribute used will be the one passed in the constructor (tango DB value, if defined, should be always ignored).
3.- Dynamic creation without root_attribute specified in constructor. In this case the root attribute will be the one defined in tango DB.

In cases 2 and 3 if root attribute is not correctly defined then the device will got to Alarm state and status will give a clue of what is wrong: all dynamic attributes will be created, but the ones with incorrect root attribute definition will throw an exception when read or written.

You can easily test the all above options using the attached TestFwd device server (gziped in "TestFwd.tgz"). You will have to play with their root attribute definition (for example using jive) in order to reproduce all cases mentioned above. If assusmes that "sys/tg_test/1/double_scalar" is available. It defines 3 forwarded attributes:
- TestFwdAttr: this is a static attribute defined using pogo (it can be used for testing case 1 above).
- TestFwdDyn1: this is a dynamic attribute with root attribute defined in the constructor (it can be used for testing case 2 above).
- TestFwdDyn2: this is a dynamic attribute with root attribute not defined in the constructor (it can be used for testing case 3 above).

The patch seems to work fine in my case. I propose to include it into tango-9-lts and master or use it as a first approach for developing a better patch (I'm far from a tango core expert, so I may have introduced other bugs while trying to solve this issue).

NOTE that it also fixes a bug I found when the dynamic forwarded attributes are created (this is last patch for "multiattribute.cpp" file): it looks like the write value is not correctly initialized and you can randomly get a coredump when reading them (you may need to restart the server several times to reproduce the error).

Regards,
Jairo Moldes

[TestFwd.zip](https://github.com/tango-controls/cppTango/files/663910/TestFwd.zip)

### DIFF
--- a/cppapi/server/fwdattrdesc.cpp
+++ b/cppapi/server/fwdattrdesc.cpp
@@ -182,8 +182,7 @@ bool FwdAttr::validate_fwd_att(vector<AttrProperty> &prop_list,const string &dev
 	catch (...) {}
 
 	//check if full_root_att is already set
-	if (full_root_att.size()!=0 && full_root_att.compare(RootAttNotDef)!=0)
-		is_full_root_att_set = true;
+	is_full_root_att_set = full_root_att.size()!=0 && full_root_att.compare(RootAttNotDef)!=0;
 
 	if(is_full_root_att_set)
 		; // root attribute was defined in constructor: do not overwrite it
--- a/cppapi/server/fwdattrdesc.cpp
+++ b/cppapi/server/fwdattrdesc.cpp
@@ -126,6 +126,7 @@ bool FwdAttr::validate_fwd_att(vector<AttrProperty> &prop_list,const string &dev
 
 	string root_att_db;
 	bool root_att_db_defined = false;
+	bool is_full_root_att_set = false;
 
     Util *tg = Util::instance();
     Database *db = tg->get_database();
@@ -180,7 +181,13 @@ bool FwdAttr::validate_fwd_att(vector<AttrProperty> &prop_list,const string &dev
 	}
 	catch (...) {}
 
-	if (root_att_db_defined == true)
+	//check if full_root_att is already set
+	if (full_root_att.size()!=0 && full_root_att.compare(RootAttNotDef)!=0)
+		is_full_root_att_set = true;
+
+	if(is_full_root_att_set)
+		; // root attribute was defined in constructor: do not overwrite it
+	else if (root_att_db_defined)
 		full_root_att = root_att_db;
     else
         full_root_att = RootAttNotDef;
--- a/cppapi/server/fwdattrdesc.cpp
+++ b/cppapi/server/fwdattrdesc.cpp
@@ -184,12 +184,12 @@ bool FwdAttr::validate_fwd_att(vector<AttrProperty> &prop_list,const string &dev
 	//check if full_root_att is already set
 	is_full_root_att_set = full_root_att.size()!=0 && full_root_att.compare(RootAttNotDef)!=0;
 
-	if(is_full_root_att_set)
-		; // root attribute was defined in constructor: do not overwrite it
-	else if (root_att_db_defined)
-		full_root_att = root_att_db;
-    else
-        full_root_att = RootAttNotDef;
+	if(!is_full_root_att_set) {
+		if (root_att_db_defined)
+			full_root_att = root_att_db;
+		else
+			full_root_att = RootAttNotDef;
+	}
 
 //
 // Check root att syntax and add TANGO_HOST info in root device name of not given
--- a/cppapi/server/multiattribute.cpp
+++ b/cppapi/server/multiattribute.cpp
@@ -884,6 +884,53 @@ void MultiAttribute::add_fwd_attribute(string &dev_name,DeviceClass *dev_class_p
 	add_default(prop_list,dev_name,attr.get_name(),attr.get_type());
 
 //
+// Validate and register the root attribute configuration
+//
+	if (new_attr->is_fwd() == true)
+	{
+		Tango::FwdAttr *fwd_attr = (Tango::FwdAttr *) new_attr;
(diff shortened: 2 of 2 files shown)


Ingvord:
Thanks. 

Please do not close existing PR next time -- continue to work with it: just push new commits into your branch, PR will be updated automatically.

This is a continuation of #341

cppapi/server/fwdattrdesc.cpp
```diff
@@ -180,7 +181,13 @@ bool FwdAttr::validate_fwd_att(vector<AttrProperty> &prop_list,const string &dev
 	}
 	catch (...) {}
 
-	if (root_att_db_defined == true)
+	//check if full_root_att is already set
+	if (full_root_att.size()!=0 && full_root_att.compare(RootAttNotDef)!=0)
```
Ingvord: I would do it like this:

```c++
is_full_root_att_set = full_root_att.size()!=0 && full_root_att.compare(RootAttNotDef)!=0;

if(!is_full_root_att_set && root_att_db_defined)
    full_root_att = root_att_db;
```

Ingvord: It does exactly the same -- I mean there is no `else` -- and much easier to read:

Instead of

```c++
if(is_full_root_att_set)
		; // root attribute was defined in constructor: do not overwrite it
else if (root_att_db_defined)
        full_root_att = root_att_db;
```

do

```c++
if(!is_full_root_att_set && root_att_db_defined)
    full_root_att = root_att_db;
```

 Because when you quickly look through the code, longer code requires much more efforts to understand what is going on. Believe me ;)

jairomoldes: Sorry, but I still don't get it. This is the original code before any modification:

```c++
if (root_att_db_defined == true)
    full_root_att = root_att_db;
else
    full_root_att = RootAttNotDef;
```

And this is what you are suggesting (correct me if I am wrong):
```c++
if(!is_full_root_att_set && root_att_db_defined)
    full_root_att = root_att_db;
else
    full_root_att = RootAttNotDef;
```
In this case if is_full_root_att_set is true then we would enter the else clause and hence we would modify the full_root_att variable

Ingvord: Ah, I did not notice the last `else`. You are right then: if `is_full_root_att_set == true` `full_root_att` is set to `RootAttNotDef`. In this case just wrap inner `if/else` with outer:

```c++
if(!is_full_root_att_set){
    if(root_att_db_defined)
        full_root_att = root_att_db;
    else
        full_root_att = RootAttNotDef;
}
```

The later I find better than empty statement.

jairomoldes: Ok. Done

bourtemb:
Hi,
Thanks you very much Jairo for this PR!
I played a bit with your FwdTest device server and I encountered problems when I declare several devices in the same device server instance.
I get the following exception in this case:

> Tango exception
> Severity = ERROR 
> Error reason = API_AttrNotFound
> Desc : Device rb/test/forward1 -> Attribute TestFwdDyn1 already exists for your device class but with other definition
> (data type, data format or data write type)
> Origin : DeviceImpl::add_attribute

It looks like there is more work to do to support this use case too.

cppapi/server/fwdattrdesc.cpp
```diff
@@ -180,10 +181,15 @@ bool FwdAttr::validate_fwd_att(vector<AttrProperty> &prop_list,const string &dev
...
 	catch (...) {}
 
-	if (root_att_db_defined == true)
-		full_root_att = root_att_db;
-    else
-        full_root_att = RootAttNotDef;
+	//check if full_root_att is already set
+	is_full_root_att_set = full_root_att.size()!=0 && full_root_att.compare(RootAttNotDef)!=0;
+
+	if(!is_full_root_att_set) {
```
bourtemb: This change is introducing a problem when the server is exporting several devices having a standard forwarded attribute (defined with POGO) in their class. If these forwarded attributes are defined to point to different root attributes (since pointing to the same root attribute in the same process is not supported by Tango), only the first created device works as expected. The other devices do not manage to create the forwarded attribute and report the following error in their status:
The device is in ALARM state.```

> 
> Forwarded attribute TestFwdAttr is not correctly configured! Root attribute name = tango://mytangohost:10000/my/root/device/myattr
> You can update it using the Jive tool
> Error: Root attribute already used in this device server process for attribute my/test/forward1/TestFwdAttr

This problem is mainly due to how Tango is managing dynamic attributes. The dynamic attributes are added to the class attributes list when they are created and if an attribute with the same name already exists in the current class attributes list, the already created Attr object is re-used.
So with the current code from this PR, when the standard forwarded attribute for the second device is created, DeviceImpl constructor will cast the Attr object which was already created for the first device and will call the validate_fwd_att() method but on the Attr object which was created for the first device. In this case, full_root_att is not empty and is pointing to the root attribute defined for the first device., leading to this error.

bourtemb:
As agreed on Friday during the Tango kernel teleconference, we will merge this PR because it is improving the situation a bit in the case where there is only one device exported by the device server (not taking into account the admin device).
Another PR will be necessary to fix the documentation in order to warn the users about this limitation.
In the case where the user uses the constructor with a root attribute defined by code, the root attribute value should be written in the DB to avoid any confusion. this will be done in another PR.

PR #347, opened 2017-01-10
added DevVarBooleanArray operators

grm84:


Ingvord:
@grm84 thank you for the contribution. It looks fine. We will need a test case though - I will try to implement it myself. Once it is done your PR will be merged.

PR #352, opened 2017-02-11
resolve #349: add missing event loop implementation in Windows builds

### LINKED ISSUE #349: server_set_event_loop() not being run in Windows
I have a tango server where the event loop runs just fine when compiled for Linux, but does not run at all in Windows.

When debugging the issue, it looks like the code in Util::server_run() surrounding the orb->run() call is very different depending on if _TG_WINDOWS_ is defined:

```
#ifdef _TG_WINDOWS_
   <snip>
   try
   {
      orb-run();
      server_cleanup();
   <snip>
#else 
   <snip>
   try
   {
      if (ev_loop_func != NULL)
      {
         <while loop that runs event loop and calls orb->perform_work()>
      }
      else
      {
         orb->run();
      }
      server_cleanup();
   <snip>
#endif
```

Is there a reason for not using the event loop functionality in windows? Would a cut/paste of the if(ev_loop_func) block fix this issue?

mike-a-smith:
Refactored the while loop that calls orb->perform_work() and the even loop function into a private method called Util::server_perform_work() and added calls to this method in the Windows #ifdef section

### DIFF
--- a/cppapi/server/utils.cpp
+++ b/cppapi/server/utils.cpp
@@ -1911,6 +1911,60 @@ void Util::server_init(TANGO_UNUSED(bool with_window))
 //+-----------------------------------------------------------------------------------------------------------------
 //
 // method :
+//		Util::server_perform_work()
+//
+// description :
+//		To call orb->run() or call orb->perform_work() and user-defined ev_loop_func() in a while loop
+//      if user has setup the event loop function
+//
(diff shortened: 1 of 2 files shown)


Ingvord:
Looks good to me. It would be nice to have a test case once we have CI for Windows.

Issue #351, opened 2017-02-10
Support DevEnum in Commands


josefcibulka:
Support for DevEnum for Command parameters is incomplete, making it unusable. It is possible to choose DevEnum as the type of input and output in Pogo, but there is no way to set the list of enum labels. This is not only a problem of Pogo, but I do not even see any equivalent of Tango::Attribute::enum_labels in Tango::Command.

PS: I would also like to see #257 implemented and since it also requires changes in the Tango::Command class, I think it would make sense to do them at the same time.

Ingvord:
@josefcibulka we have merged #357 into the master branch  which resolves this issue. Could you please test it either building from source or using our debian repo: [link](https://bintray.com/tango-controls/debian/cppTango/10.0.1)? If you confirm that it is OK we will close the issue. Thanks in advance.

josefcibulka:
@Ingvord  Apologies for the long delay. I confirm that it works well for me. Am I correct that the latest versions of Pogo and ATK panel do not support it yet? Should I open a ticket in ATK?

Ingvord:
@josefcibulka thanks for the positive response. Yes, please open issues in the corresponding projects.

PR #357, opened 2017-03-12
PR for #351

Ingvord:
PR for #351


include/tango/client/device/device_command_info.hxx
```diff
@@ -0,0 +1,82 @@
+//
```
bourtemb: When you create a new file, try to keep the standard Tango comments at the top of the file: 
```
//
// This file is part of Tango.
// etc.
```

Ingvord:
Unfortunately we can not just add new functions to the existing IDL interfaces (aka Device_5), because this produces the following error in case of _new_ client talks to _old_ server:

```
Tango exception
Severity = ERROR
Error reason = API_CorbaException
Desc : BAD_OPERATION CORBA system exception: BAD_OPERATION_UnRecognisedOperationName
Origin : DeviceProxy::command_query()


Tango exception
Severity = ERROR 
Error reason = API_CommunicationFailed
Desc : Failed to execute command_query on device test/debian8/10
Origin : DeviceProxy::command_query()
```

here client calls for `Device_5::command_query_5()`

This means that everytime we add a new functionality we have to create a new interface... which will blow out tens of them :( 

We have to take this into account when implementing new Tango protocol, i.e. automatically fall back to existing function on the server side and client must handle such situation e.g. provide default behavior, ignore etc

include/tango/client/device/device_command_info.hxx
Ingvord: OK,

CLion can help us with this, there is a feature to automatically add copyright. We just need to create a template and put it into the repo. So everyone can use this template in the future.

Ingvord: Project code style has been added to the master branch, see ea387e6. Also a new wiki page has been created: [Code style guidelines](https://github.com/tango-controls/cppTango/wiki/Code-style-guidelines)

Ingvord:
@bourtemb: PR is ready for review

.idea/fileTemplates/includes/C File Header.h
```diff
@@ -0,0 +1,28 @@
+#if ($HEADER_COMMENTS)
```
bourtemb: I think this change is coming from another PR, but it would be a good idea to avoid file names containing spaces if possible.

Ingvord: Well, it is internal stuff for CLion. I do not think we can rename it.

src/server/eventsupplier.cpp
```diff
@@ -141,9 +141,11 @@ SendEventType EventSupplier::detect_and_push_events(DeviceImpl *device_impl,stru
...
 		switch (*ite)
 		{
-			case 5:
-			if (change5_subscription >= EVENT_RESUBSCRIBE_PERIOD)
-				attr.remove_client_lib(5,string(EventName[CHANGE_EVENT]));
+            //TODO extract class hierarchy based on version!!!
+            case 6:
+            case 5:
+                if (change5_subscription >= EVENT_RESUBSCRIBE_PERIOD)
+                    attr.remove_client_lib(5, string(EventName[CHANGE_EVENT]));
```
bourtemb: I think we might encounter issues when *ite == 6 here...
I would replace this line with:
`attr.remove_client_lib(*ite, string(EventName[CHANGE_EVENT]));`

src/server/eventsupplier.cpp
```diff
@@ -177,9 +179,11 @@ SendEventType EventSupplier::detect_and_push_events(DeviceImpl *device_impl,stru
...
 		switch (*ite)
 		{
-			case 5:
-			if (periodic5_subscription >= EVENT_RESUBSCRIBE_PERIOD)
-				attr.remove_client_lib(5,string(EventName[PERIODIC_EVENT]));
+            //TODO extract class hierarchy based on version!!!
+            case 6:
+            case 5:
+                if (periodic5_subscription >= EVENT_RESUBSCRIBE_PERIOD)
+                    attr.remove_client_lib(5, string(EventName[PERIODIC_EVENT]));
```
bourtemb: I think we might encounter issues when *ite == 6 here...
I would replace this line with:
`attr.remove_client_lib(*ite, string(EventName[PERIODIC_EVENT]));`

src/server/eventsupplier.cpp
```diff
@@ -213,9 +217,11 @@ SendEventType EventSupplier::detect_and_push_events(DeviceImpl *device_impl,stru
...
 		switch (*ite)
 		{
-			case 5:
-			if (archive5_subscription >= EVENT_RESUBSCRIBE_PERIOD)
-				attr.remove_client_lib(5,string(EventName[ARCHIVE_EVENT]));
+            //TODO extract class hierarchy based on version!!!
+            case 6:
+            case 5:
+                if (archive5_subscription >= EVENT_RESUBSCRIBE_PERIOD)
+                    attr.remove_client_lib(5, string(EventName[ARCHIVE_EVENT]));
```
bourtemb: I think we might encounter issues when *ite == 6 here...
I would replace this line with:
`attr.remove_client_lib(*ite, string(EventName[ARCHIVE_EVENT]));`

taurel:
Hi guys,

I don't think I will have time to look in details to this PR before my next week of holidays (from 22 to 26/05). Nevertheless, I had a look at the test file. I think some tests (features?) are missing:
- I don't see any insertion / extraction into DeviceData instance using a real C++ enum
(Which by the way is also missing in the enum att test...)
- I don't see any test of command polling for command with void in type but with enum as output type
- I don't see any use of enum with dynamic labels.
- I don't see test for dynamic command with enum type

Cheers

Manu

Ingvord:
@taurel Thanks for the input! 

I will add the test cases mentioned, but I do not get what is dynamic enum labels. For me enum is a compile time constant. Is there any example of this for attributes?

taurel:
Hi Igor,

_I will add the test cases mentioned, but I do not get what is dynamic enum labels. For me enum is a compile time constant. Is there any example of this for attributes?_
Look at Tango book chapter 7.2.1.2. This feature was a requirement when the enumerated attributes
were implemented. I must admit that I have already used it in some of the Tango class I recently wrote

_BTW I am not sure about test for DeviceData insertion/extraction - correct me if I am wrong, but if client wants to use native C++ enum with DeviceData this means to provide compile time enum constant on the client side i.e. hard-coded in the client code... now server has been changed via pogo, ooops - client inserts wrong data..._
Agree. Insertion/Extraction into/from DeviceData using native C++ enum requires no change in
enum definition but for some device, it happens (Different scales of a measurement device for instance)

_Concerning polling - command with DevEnum output type simply returns short. We can add this test, but it has no value, as polling of short is already tested (maybe not, but it works, otherwise we would have tons bug reports)_
Take care, as far as I remember, in order to properly store data in the polling buffer, there is a kind of
switch in function of the command output data type. Command returning DevEnum is a new
output data type. Better to test if this works as well

Good luck

Manu

BTW, did you also see my comment on the IDL change ?

Ingvord:
@taurel:concerning dynamic labels the numeration of the documentation is deprecated, but I guess we are talking about this feature [setting-the-labels-without-enumeration-compile-time-knowledge](http://tango-controls.readthedocs.io/en/latest/development/device-api/enumerated-attribute.html#setting-the-labels-without-enumeration-compile-time-knowledge). I would not add this within this PR but rather wait for a request (if any). Because this particular PR is to fix #351 i.e. provide a way to set/read labels for a command with DevEnum as argin/argout. 

In overall IMHO labels MUST be compile time constant otherwise we will never be able to implement them using native enums.

Ingvord:
@taurel do you have any more comments? Otherwise I will merge this PR -  it would be nice to have it in the master before Tango meeting in Firenze next week.

taurel:
Hi Igor,

I do not really understand how you manage command with enumeration when you want to set the strings
associated to the "real C++ enum". In the test DS in the EnumLabelsDynCommand, you call the Command class get_ext() method.
I am not sure it is the best solution. For me these extension class should be completely hidden for the user. Otherwise, what will happen when the data member(s) temporary added in the extension class are
moved to the official class? The user code won't work anymore.
May be a call a call in the Command class to retrieve these vector of strings is more "safe".

Ingvord:
@taurel access to in/out_enum_labels has been encapsulated in  3846d3d There is no `get_` methods as in this PR enum labels are considered constant (client sets the labels ones and forgets about it)

I was thinking about using native enums and I am not sure if it is possible in C++ [marshaling/unmarshaling; even setting a field on the server side will require some kind of a generic pointer] (no problems with Java though %) ) Anyway this discussion is out of the scope of this PR.

taurel:
Hi Igor,

About using native C++, isn't it possible to have for command something similar to what already
exist for attribute. I am thinking of what is explained in doc chapter 7.2.1.3. I am still
talking in old Tango book chapter numbering because this is what I find if I look in the web server in the
doc page.
As far as I remember, this was done using template method.
The same usage of native C++ enum is also supported for insertion/extraction iinto/from DeviceAttribute
object. Therefore, I think it should also be available for DeviceData object

Cheers

Emmanuel

Ingvord:
We can do it in a dedicated PR later

Ingvord:
@taurel if you do not have any further comments I will merge this PR

taurel:
Go ahead

Ingvord:
@taurel thanks a lot for your input and advice on native enums. This is very useful!!!

include/tango/server/command.h
```diff
@@ -220,9 +222,10 @@ class Command
  * The object desctructor.
  */
 #ifdef HAS_UNIQUE_PTR
-    virtual ~Command() {}
+    virtual ~Command()
+    {}
 #else
-	virtual ~Command() {delete ext;}
+    virtual ~Command() {delete ext;}
```
bourtemb: Minor improvement: ext is now always a shared_ptr since this PR (See line 1275 in this file and line 1211 in the previous version), so this #ifdef could be removed (the #else part could be removed).

Ingvord: OK, thanks for the tip. Will do it in the master since this PR is already merged

bourtemb: Actually, this was found thanks to @chedburgh who ran a tool to check cppTango code quality.
Thanks Stuart! :-)

include/tango/server/device_6.h
```diff
@@ -0,0 +1,75 @@
...
+                                                                                               desc) {
+            idl_version = 6;
+        }
+
+        Device_6Impl(DeviceClass *device_class, string &dev_name, string &desc, DevState dev_state, string &dev_status)
+                : Device_5Impl(device_class, dev_name, desc, dev_state, dev_status) {
+            idl_version = 6;
+        }
+
+        Device_6Impl(DeviceClass *device_class, const char *dev_name, const char *desc, DevState dev_state,
```
bourtemb: In Device_5Impl version, there are some default values defined for desc, dev_state and dev_status arguments.
This is important if you want to be backward compatible with servers generated with Pogo which generates code like the following for a Tango class named TestCmds:

> TestCmds::TestCmds(Tango::DeviceClass *cl, const char *s, const char *d)
>  : TANGO_BASE_CLASS(cl, s, d)
> {
> 	/*----- PROTECTED REGION ID(TestCmds::constructor_3) ENABLED START -----*/
> 	init_device();
> 	
> 	/*----- PROTECTED REGION END -----*/	//	TestCmds::constructor_3
> }

Ingvord: Got it! Thanks for the input!

Issue #360, opened 2017-04-12
NaN value in array property


taurel:
I have a class property that I retrieve as a vector of double. If in DB, one of the array value is set to
NaN (or nan), the DbDatum class extractor method convert the NaN to zero and idem for all the values
in the array after the NaN.
I guess the same is true for other kind of properties (device)

PR #362, opened 2017-04-13
Add support for NaN and inf in device properties (#360)

bourtemb:


bourtemb:
A test should be added to test this feature.

PR #366, opened 2017-05-05
FeatureRequest on improving exception creation

Ingvord:
Currently developer must provide an origin of the exception in the code manually:

```c++

ApiCommExcept::re_throw_exception(one,
    (const char*)"API_CommunicationFailed",
    desc.str(),
    (const char*)"DeviceProxy::command_list_query()");// NOTE THIS LINE

```

We can automatize this using `__func__`, `__LINE__` and `__FILE__`. See for instance [this post](http://stackoverflow.com/questions/597078/file-line-and-function-usage-in-c) on SO


t-b:
Sounds like a good idea. If we do that I would also appreciate overloads accepting std::string.

t-b:
According to [1] I think using __PRETTY_FUNCTION__ for gcc should be prefered over __func__ as it has the namespace, class and function parameters as well. MSVC seems to not have something as good.

Unfortunately we would have to do this with a macro though.

[1]: https://stackoverflow.com/a/4384825

Ingvord:
Well, I was thinking about to use this library: https://github.com/boostorg/stacktrace

But as we prefer not to depend on 3rd party libraries we will have to fetch the source code and include into our code, or just copy-and-paste the main idea...

t-b:
@Ingvord Nice library! But a big invasive for my taste. You could make depending on it optional. Just grabbing the source and including it is usally a bad idea.

PR #367, opened 2017-05-07
Fix misleading-indentation warning for GCC 6.0 and higher

t-b:
A more future proof fix would probably be to add a .clang-format definition and a `make format` target calling `clang-format`.

### DIFF
--- a/include/tango/server/attribute.tpp
+++ b/include/tango/server/attribute.tpp
@@ -219,10 +219,10 @@ void Attribute::set_min_alarm(const T &new_min_alarm)
 // If the server is in its starting phase, give a NULL ptr to the AutoLock object
 //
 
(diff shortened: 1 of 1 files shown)


Ingvord:
Is this 4 spaces indentation?

t-b:
@Ingvord It's tab like the surrounding code.

Ingvord:
@t-b we are migrating to 4 spaces. Could you please confirm that this is OK for GCC 6.0 as well?

t-b:
@Ingvord Yes still works when I convert the whole file via `sed -i  's/\t/    /g' attribute.tpp` to four spaces..

t-b:
@Ingvord thanks.

PR #376, opened 2017-07-17
Fix warnings reported by gcc 6.3 related to misleading indentations.

bourtemb:
The files where these warnings were reported have been formatted by CLion using
the cppTango CLion code style already in use on the master branch.
Added CLion code style settings file based on Allman Braces code style (#364).

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -62,6 +62,10 @@
           <option name="BINARY_OPERATION_SIGN_ON_NEXT_LINE" value="true" />
           <option name="FOR_STATEMENT_WRAP" value="1" />
           <option name="ASSIGNMENT_WRAP" value="1" />
+          <option name="IF_BRACE_FORCE" value="3" />
+          <option name="DOWHILE_BRACE_FORCE" value="3" />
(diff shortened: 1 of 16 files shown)


bourtemb:
This PR adds a CLion project code style settings file.
The file used in the master branch was used as starting point.
This file has been modified in order to force braces in _if/while/do while/for_ statements since this is what Codacy recommends.
I noticed CLion's _Reformat Code_ feature does not work in conditionally non-compiled code (#ifdef code when condition is false) but Codacy also complains about issues in the conditionally non-compiled code, which is good.

For the record, Travis CI tests didn't pass the first time for the DEBUG version.
It complained about an assertion failed on new_tests/cxx_fwd_att.cpp:749 (cb.cb_err == 1).
I restarted the test and it passed successfully the second time.

It seems Codacy is reporting some false positive issues in devapi_data.cpp. It says some conditions are always true. In this specific case, this is not true, otherwise this would mean all the DeviceData extraction operators are not working when extracting to a vector.

Ingvord:
I would ignore some of the braces for instance in this case:

```C++

if(...) return;// or return some simple value
```

I usually write this in one line as above.

PR #377, opened 2017-07-17
Applying ThreadSanitizer

Ingvord:
[ThreadSanitizer](https://github.com/google/sanitizers/wiki/ThreadSanitizerCppManual) seems to be a usefull thing. It would be nice to integrate it with cppTango


bourtemb:
Interesting tool.
I think we should run the tests in DEBUG mode with it but we should not compile the RELEASE version with this flag since it has an impact on the performances

> The cost of race detection varies by program, but for a typical program, memory usage may increase by 5-10x and execution time by 2-20x.

Ingvord:
```
FATAL: ThreadSanitizer can not mmap the shadow memory (something is mapped at 0x55d514e1c000 < 0x7cf000000000)
FATAL: Make sure to compile with -fPIE and to link with -pie.
```

Even though -fPIE and -pie are there...

bourtemb:
I think the solution is on the ThreadSanitizer FAQ (1st question):

> - Q:  When I run the program, it says: FATAL: ThreadSanitizer can not mmap the shadow memory (something is mapped at 0x555555554000 < 0x7cf000000000). What to do? You need to enable ASLR:
> 
> $ echo 2 >/proc/sys/kernel/randomize_va_space
> 

Did you do that?

Ingvord:
Sure. In my case (Debian 8) I have permission denied even under root user.

Ingvord:
OK I am able to enable ThreadSanitizer on my kubuntu 17.04

```
$ uname -a
Linux FUJITSU 4.10.0-28-generic #32-Ubuntu SMP Fri Jun 30 05:32:18 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

Following [this debian bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=796246#33) and using `-no-pie` link option (see aee3fd5).

But this does not work on Debian 8 i.e. we can not integrate it with CI.

So I simply attach test reports here and close this PR. These reports must be considered when we start the refactoring:


[ctest.txt](https://github.com/tango-controls/cppTango/files/1185877/ctest.txt)

[DevTest_test.txt](https://github.com/tango-controls/cppTango/files/1185879/DevTest_test.txt)

[DevTest_test2.txt](https://github.com/tango-controls/cppTango/files/1185880/DevTest_test2.txt)

[FwdTest_test.txt](https://github.com/tango-controls/cppTango/files/1185881/FwdTest_test.txt)

PR #378, opened 2017-07-18
Implement commands with PipeBlob in/out

Ingvord:
Start


Ingvord:
Looks like it is enough to add corresponding methods to DeviceData

src/client/devapi_data.cpp
```diff
@@ -341,6 +341,12 @@ int DeviceData::get_type()
 
             case CORBA::tk_struct:
                 tc_field = tc->member_type(0);
+                if(tc_field->kind() == tk_string)
+                {
+                    // The first field in a DevPipeBlob structure is a string (name field)
```
Ingvord: Magic :)

Ingvord:
Ready for review!

@bourtemb , @taurel could you please provide your input?

src/client/devapi_base.cpp
```diff
@@ -4480,234 +5146,271 @@ namespace Tango {
...
+    if (tmp_ptr == Tango_nullptr)
+    {
+        Except::throw_exception(API_PipeNoDataElement, "No data in pipe!", "DeviceProxy::write_pipe()");
     }
 
+    DevULong max, len;
+    max = tmp_ptr->maximum();
+    len = tmp_ptr->length();
+    result.data_blob.blob_data.replace(max, len, tmp_ptr->get_buffer((DevBoolean) true), true);
+    delete tmp_ptr;
```
bourtemb: Are you sure it is correct to delete tmp_ptr before `dev->write_pipe_5(pipe_value_5, ci);` statement?
Please be extra careful with the meaning of 2 boolean parameters passed in:
`result.data_blob.blob_data.replace(max, len, tmp_ptr->get_buffer((DevBoolean) true), true);` statement because there are some memory ownership transfer involved here.

.travis.yml
```diff
@@ -19,7 +19,8 @@ services:
 before_install:
   - docker pull tangocs/mysql:9.2.2
   - docker pull tangocs/tango-cs:latest
-  - git clone https://${CI_USER_TOKEN}@github.com/tango-controls/tango-idl.git idl
+  - git clone https://github.com/JoakimSoderberg/coveralls-cmake.git
+  - git clone -b cppTango-378 https://${CI_USER_TOKEN}@github.com/tango-controls/tango-idl.git idl
```
bourtemb: I understood this is no longer necessary since the IDL didn't change.

Ingvord: Right! There is no change in cppTango-378 branch in tango-idl. So this is irrelevant and does not affect the code base

src/client/devapi_base.cpp
Ingvord: Well, honestly speaking this code is copy-n-paste from reading/writing pipes... so I guess (99% sure) it is ok. Otherwise tests must fail

bourtemb: This is not only copy and paste. The code was refactored here (createDevPipeData method did not exist before). delete tmp_ptr does not occur at the same time as it used to be. The previous code strongly suggests there was a trick involved to avoid unnecessary copies. If you can, please have a look at the book about CORBA I was suggesting before. There is a section about sequences and the replace() and get_buffer() method, as well as the meaning of their parameters. I would also recommend testing this with valgrind to check eventual memory problems. Maybe this code is correct, but the way this part was implemented is too unusual and I have the feeling there was a good reason for that.

bourtemb: Actually it seems to be fine to delete _tmp_ptr_ just after _tmp_ptr->get_buffer(true)_ in this case.
The parameter in _get_buffer()_ is named _orphan_. If _orphan_ is true, _tmp_ptr_ is no longer responsible for the memory which was previously allocated for its buffer sequence. _tmp_ptr->get_buffer(true)_ returns a pointer to the data which was previously allocated for _tmp_ptr_ buffer sequence and resets _tmp_ptr_ sequence buffer (pointer set to 0 and sequence length reset to 0 without deallocating the previously allocated memory).
_result.data_blob.blob_data_ is now responsible for deallocating the memory which was previously belonging to _tmp_ptr_.
The last parameter from replace method (_release_ parameter) is important here too. When set to true as here, it indicates that the sequence result.data_blob.blob_data will take care of deallocating the memory of its buffer.
It looks like the previous code was written that way (deleting tmp_ptr later) just to be on the safe side. We can wait for @taurel's review to be sure.

test/cpp_test_suite/new_tests/cxx_cmd_types.cpp
```diff
@@ -879,24 +879,88 @@ class CmdTypesTestSuite: public CxxTest::TestSuite
...
-            dout.extract(&received);
+            DevLong long_data = 123;
+            TS_ASSERT_THROWS_NOTHING(in["long"] << long_data);
+            DevDouble double_data = 3.14;
+            TS_ASSERT_THROWS_NOTHING(in["double"] << double_data);
+            string data_str = "String value";
+            TS_ASSERT_THROWS_NOTHING(in["string"] << data_str);
+            vector<double> double_array = {0.1, -0.02, 0.003, -0.0004};
+            TS_ASSERT_THROWS_NOTHING(in["double array"] << double_array);
+            TS_ASSERT_THROWS_NOTHING(din << in);
```
bourtemb: If we do something like the following here, between the insert and the command_inout:
```
DevDouble double_data2 = 789.123;
in["double"] << double_data2;
```
Then, with the previous version of DeviceData::insert, it crashes.
With the [new version](https://github.com/tango-controls/cppTango/pull/378/commits/2ceeb7edeb841c7a06d823971b880cbbf3b0436c), the DeviceData will point to the modified DevicePipeBlob, meaning the data element named "double" will contain the new value, not the one which was in DevicePipeBlob object when the insert occurred.
We can discuss whether this is good or bad. If we stay with this solution which is similar to what is done for DevEncoded, we need to document this well and warn the users about this pitfall.
I think there will be a crash if the DevicePipeBlob object is deleted before the command_inout occurs. This should be very well documented too if we stay with this approach.
If we go for the more intuitive approach which would consist in copying everything from the DevicePipeBlob object when the insert into a DeviceData occurs, we will clearly degrade performances if the blob is big.

Ingvord: I would first stay with more intuitive approach and only if necessary (users report performance issues + benchmark test) will try to optimize it. 

In fact what is need here is builder pattern: first user modifies DevPipeBlobBuilder with the data then freezes it (acquiring immutable DevicePipeBlob) and pass it to command_inout OR we can perform defensive copying inside command_inout

Ingvord:
@taurel Could you please review this PR?

Ingvord:
> Hi Igor,
> 
> As far as I remember, there were actually at least 2 remarks:
> It was very difficult to review the PR because there was no easy way to differentiate changes coming from the implementation of this features and changes coming from other commits fixing bugs or implementing other features. 
> So I advised to do an interactive rebase with only the commits relevant to this PR so we could review this PR in good conditions.
> There was a concerned raised by Emmanuel on the potential performance problems with using CORBA::Any for the transfer of DevPipeBlob data.
> Kind regards,
> Reynald

1 - I do not see any non-relevant commits in the history; maybe this was due to non-rebased state of the branch at the time review was performed. Anyway I am not very strong at git to do interactive rebase as I always mess it
2 - I would postpone performance consideration till the time we have reasonable code base quality so we can observer all other (maybe much more important) issues

Taking 1,2 into account I think it is OK to merge this PR now

bourtemb:
Did you notice that Travis CI is broken because it is trying to clone cppTango-378 branch from tango-idl?
As a minimum, I think we should check Travis CI status before merging.
Sorry for not finding the time yet to review this PR under good conditions after you rebased it correctly.

Ingvord:
@bourtemb this is already fixed in master: https://travis-ci.org/tango-controls/cppTango/builds/289485564

It is now failing due to https://travis-ci.org/tango-controls/cppTango/builds/289511969

cherry-pick of b719402

bourtemb:
> @bourtemb this is already fixed in master: https://travis-ci.org/tango-controls/cppTango/builds/289485564

Good! :-)

> It is now failing due to https://travis-ci.org/tango-controls/cppTango/builds/289511969
> cherry-pick of b719402

Not good :-(

Ingvord:
> Not good :-(

Yeah, I have reverted this commit and will create an issue

PR #383, opened 2017-07-24
Fix shift-negative-value compilation warnings reported by g++ 6.3

bourtemb:


bourtemb:
[This article](http://www.gamasutra.com/blogs/AndreyKarpov/20120416/168596/Wade_not_in_unknown_C_waters_Part_three.php?print=1) was actually quite interesting on this subject, even though the solution they proposed didn't compile on g++ 6.3. I had to add a cast to (int).

But this is suggesting that this code was originally inspired by libjpeg or a similar project... I think it would be fair to add a note in the distribution and documentation to acknowledge their work. Unless I missed it, it looks like this is not done in the latest versions of the Tango distribution.

Ingvord:
@bourtemb thanks for the very interesting article. In general I would say we have to avoid any optimizations in code (including non-trivial algos)(and especially not proofed by the tests) as usually performance bottlenecks are in <10% of code. Rather we have to focus on modularity and modifiability. Achieving this will give a clear way to test each component independently including performance testing. So we have to refactor, refactor and refactor :)

Talking about bit shifts there are also quite a few puzzlers described in [this book](http://www.javapuzzlers.com/). So every Java developer must read this book (@JeanLucPons). General things such as bitwise shifts or double precision computations are common for both languages and can be interesting to every one (@bourtemb).

PR #385, opened 2017-07-28
improve event subscription

Ingvord:
Currently each client and server guesses about the channel to use for the event subscription. 

In this PR we want to implement handshaking mechanism in which client asks server for the channel.

As this will be back ported into tango-9-lts minimal code refactorings must be applied.

### DIFF
--- a/src/common/admin/commands/zmq_event_subscription_change_response.h
+++ b/src/common/admin/commands/zmq_event_subscription_change_response.h
@@ -49,7 +49,7 @@ namespace commands
  * This structure represents AdminDevice::ZmqSubscriptionChangeResponse
  *
  */
-struct ZmqChangeSubscriptionResponse
+struct ZmqSubscriptionChangeResponse
 {
     int tango_lib_release;
     int dev_idl_version;
@@ -65,7 +65,7 @@ struct ZmqChangeSubscriptionResponse
 
     std::vector<std::pair<std::string, std::string>> alternative_endpoints;
 
-    ZmqChangeSubscriptionResponse(int tango_lib_release,
+    ZmqSubscriptionChangeResponse(int tango_lib_release,
                                   int dev_idl_version,
                                   int zmq_sub_event_hwm,
                                   int rate,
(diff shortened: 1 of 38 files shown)


Ingvord:
Looks like cppTango does not set ZMQ topic filter: [see my gist](https://gist.github.com/Ingvord/b7253cdf56d37d4d98a85809c233a6e3)

Filtering is performed manually through some non-trivial code, like:

```cpp
        string local_callback_key(device_name);

	string::size_type pos;
	if ((pos = local_callback_key.find('#')) == string::npos)
	{
		if (inter_event == true)
			local_callback_key = local_callback_key + "." + event_name;
		else
			local_callback_key = local_callback_key + "/" +obj_name_lower + "." + event_name;
	}
	else
	{
		local_callback_key.erase(pos);
		if (inter_event == true)
			local_callback_key = local_callback_key + MODIFIER_DBASE_NO + '.' + event_name;
		else
			local_callback_key = local_callback_key + "/" + obj_name_lower + MODIFIER_DBASE_NO + '.' + event_name;
	}
```

I guess this was designed to prevent socket being created for every subscription (I wonder if ZMQ reuses sockets underneath)

Ingvord:
a few hundred lines below:
```cpp
//
// Subscribe to the new event
//

            event_sub_sock->setsockopt(ZMQ_SUBSCRIBE,event_name,::strlen(event_name));
```

so ZMQ topics are actually used...

src/common/admin/commands/zmq_event_subscription_change_response.h
```diff
@@ -0,0 +1,162 @@
...
+
+namespace org
+{
+namespace tango
+{
+namespace common
+{
+namespace admin
+{
+namespace commands
```
bourtemb: Hello!
This part looks very Java-ish to me ;-)...
Do we really need so many nested namespaces? I would drop some of them to improve readability of the code.

Ingvord: Maybe, you are right... perhaps `org::tango::common` would be enough.

P.S. once we switch to C++17 we can replace this with `org::tango::common::admin::commands`

bourtemb: I would even remove the 'org' part. I don't think we need it.

Ingvord: OK, I will remove `org` but leave the rest to be consistent with the folder structure. We will see how this evolve and maybe change in the future.

Ingvord:
I do not like where it goes. So I close this one and start another #404

PR #387, opened 2017-07-30
place opening bracket in the end of line

Ingvord:
I suggest to change the code style to place opening bracket in the end of line. This helps to investigate code i.e. when I put the cursor onto the ending bracket CLion popups the code line with the opening one:

code_highlight

Here I see the full if statement. Otherwise only the bracket is shown:

code_highlight1

As our code has tons of very long if/switch/classes/functions this is critical.

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -7,9 +7,6 @@
         <Objective-C>
           <option name="INDENT_NAMESPACE_MEMBERS" value="0" />
           <option name="KEEP_STRUCTURES_IN_ONE_LINE" value="true" />
-          <option name="NAMESPACE_BRACE_PLACEMENT" value="2" />
-          <option name="FUNCTION_BRACE_PLACEMENT" value="2" />
-          <option name="BLOCK_BRACE_PLACEMENT" value="2" />
           <option name="FUNCTION_PARAMETERS_WRAP" value="5" />
           <option name="FUNCTION_CALL_ARGUMENTS_WRAP" value="5" />
           <option name="TEMPLATE_CALL_ARGUMENTS_WRAP" value="5" />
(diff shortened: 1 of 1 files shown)


bourtemb:
One can use Ctrl+Shift+M shortcut to toggle between code block start or end. You can see the if statement in this case.
This is described in this web page as some other tricks related to brackets: https://www.jetbrains.com/help/clion/navigating-to-braces.html
I would prefer to stick with the current style.

bourtemb:
Ooops... Didn't want to close the discussion... Wrong button. Sorry.

Ingvord:
Yeah, but then you literally jump to the opening bracket position - which is not that convenient comparing to just glance onto popup

bourtemb:
It is a bit less convenient I agree. Typing the shortcut again bring you back to where you were.
I think this is something they should improve in CLion.  When you switch to the beginning of the block with Ctrl+Shift+M, it switches to the line where the if statement is, so CLion knows where is this line.
CLion is already able to display several lines in the preview Window (They do it with the Context info option Alt+Q). So they should be able to improve that.
I would rather ask for an improvement in CLion than change our code style because the current version of the IDE is not optimal.

Ingvord:
But why do you prefer bracket on a new line in the first place?

bourtemb:
Well, we are entering into a religious debate here.
I prefer because:
- This is the code style which was originally used/chosen by the project and I think it is important to be consistent in a project.
- It makes it easier to see the differences between the new versions and the old versions of the files with standard tools (git diff, diff) if we don't change too much the code style. I know CLion can help on that but I also like to use other tools from a console.
- I think you distinguish better the code blocks with this style
- This is the code style I am also used to program with. I understand your point of view... You are used to another style so you would like to switch to it too.

Ingvord:
Well, I have just found out that IDEA actually works well in this case:
,
idea_code_highlight


 so I agree this can be improved in CLion itself, here is the issue - [link](https://youtrack.jetbrains.com/issue/CPP-10367) 

Could you please can vote for it to speed up the process?

I will close this PR

bourtemb:
Thanks for creating the issue on Jetbrains Youtrack. I voted for it.

Issue #392, opened 2017-09-02
DeviceAttribute constructor fails for short datatype (Regression from Tango 8)


t-b:
The following code 

```cpp
    std::string attr = "myname";
    short data = -111;
    Tango::DeviceAttribute da1;
    da1.set_name(attr);
    da1 << data;
    std::cout << "type: " << da1.get_type() << std::endl;

    Tango::DeviceAttribute da2(attr, data);
    std::cout << "type: " << da2.get_type() << std::endl;
```
gives with tango 9.2.5a

```
type: 100
type: 914331536
```

So the the constructor with signed short does not work at all, but the first method gives an unknown type. (DATA_TYPE_UNKNOWN is 100).

bourtemb:
Good catch, Thomas!
Thanks for the bug report.
It seems this bug was introduced when DevEnum support was added in Tango 9.
It looks like DeviceAttribute `data_type` field is not initialized in this case.
The initialization is now necessary for the get_type() method to work as expected.

Ingvord:
Resolved in 3721f6e

t-b:
@bourtemb @Ingvord Thanks.

PR #393, opened 2017-09-08
Fix DeviceAttribute constructor for short and enum data types (#392)

bourtemb:
Add tests for DeviceAttribute constructors and tests for insertion operators for short and enum types.
DeviceAttribute::get_type() no longer throws an exception when DeviceAttribute object is empty.
It now returns DATA_TYPE_UNKNOWN in this case.

### DIFF
--- a/cppapi/client/devapi_attr.cpp
+++ b/cppapi/client/devapi_attr.cpp
@@ -2069,9 +2069,18 @@ bool DeviceAttribute::is_empty()
 int DeviceAttribute::get_type()
 {
 	int da_type;
+	bool da_is_empty;
 
-	if (is_empty() == true)
-		return -1;
+	// reset is_empty exception flag to avoid throwing an exception
+	// during is_empty() call
+	bitset<DeviceAttribute::numFlags> bs = exceptions();
+	reset_exceptions(DeviceAttribute::isempty_flag);
+	da_is_empty = is_empty();
+	// restore is_empty exception flag
+	exceptions(bs);
+
+	if (da_is_empty)
+		return DATA_TYPE_UNKNOWN;
 	else
 	{
 		if (LongSeq.operator->() != NULL)
@@ -2103,7 +2112,7 @@ int DeviceAttribute::get_type()
 		else if ((StateSeq.operator->() != NULL) || (d_state_filled == true))
 			da_type = Tango::DEV_STATE;
         else
-            da_type = -1;
+            da_type = DATA_TYPE_UNKNOWN;
 	}
 
 	return da_type;
--- a/cppapi/client/devapi_attr.cpp
+++ b/cppapi/client/devapi_attr.cpp
@@ -1118,6 +1118,7 @@ DeviceAttribute::DeviceAttribute(string& new_name, vector<short> &datum):ext(new
 	exceptions_flags.set(isempty_flag);
 	ShortSeq = new(DevVarShortArray);
 	ShortSeq.inout() << datum;
+	data_type = DEV_SHORT;
 }
 
 DeviceAttribute::DeviceAttribute(const char *new_name, vector<short> &datum):ext(new DeviceAttributeExt)
@@ -1134,6 +1135,7 @@ DeviceAttribute::DeviceAttribute(const char *new_name, vector<short> &datum):ext
 	exceptions_flags.set(isempty_flag);
 	ShortSeq = new(DevVarShortArray);
 	ShortSeq.inout() << datum;
+	data_type = DEV_SHORT;
 }
 
 DeviceAttribute::DeviceAttribute(string& new_name, vector<short> &datum,int x,int y):ext(new DeviceAttributeExt)
@@ -1150,6 +1152,7 @@ DeviceAttribute::DeviceAttribute(string& new_name, vector<short> &datum,int x,in
 	exceptions_flags.set(isempty_flag);
 	ShortSeq = new(DevVarShortArray);
 	ShortSeq.inout() << datum;
+	data_type = DEV_SHORT;
 }
 
 DeviceAttribute::DeviceAttribute(const char *new_name, vector<short> &datum,int x,int y):ext(new DeviceAttributeExt)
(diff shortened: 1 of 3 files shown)


cppapi/client/devapi_attr.cpp
```diff
@@ -563,6 +563,7 @@ DeviceAttribute::DeviceAttribute(string &new_name, short datum):ext(new DeviceAt
 	ShortSeq = new(DevVarShortArray);
 	ShortSeq->length(1);
 	ShortSeq[0] = datum;
+	data_type = DEV_SHORT;
```
Ingvord: Looks to me that the whole constructor can be delegated to the *DeviceAttribute(const cha*)*:

```c++
DeviceAttribute::DeviceAttribute(string &new_name, short datum):ext(new DeviceAttributeExt)
{
    this(new_name.c_str(), datum);
}
```

Is this available in C++? Otherwise a helper method may be introduced.

cppapi/client/devapi_attr.cpp
```diff
@@ -1116,6 +1118,7 @@ DeviceAttribute::DeviceAttribute(string& new_name, vector<short> &datum):ext(new
 	exceptions_flags.set(isempty_flag);
 	ShortSeq = new(DevVarShortArray);
 	ShortSeq.inout() << datum;
+	data_type = DEV_SHORT;
```
Ingvord: Looks to me that the whole constructor can be delegated to the *DeviceAttribute(const cha*)*:

```c++
DeviceAttribute::DeviceAttribute(string &new_name, short datum):ext(new DeviceAttributeExt)
{
    this(new_name.c_str(), datum);
}
```

Is this available in C++? Otherwise a helper method may be introduced.

cppapi/client/devapi_attr.cpp
```diff
@@ -2097,7 +2112,7 @@ int DeviceAttribute::get_type()
 		else if ((StateSeq.operator->() != NULL) || (d_state_filled == true))
 			da_type = Tango::DEV_STATE;
         else
-            da_type = -1;
+            da_type = DATA_TYPE_UNKNOWN;
```
Ingvord: **return** maybe inlined here and above in if-else-if-...

cppapi/client/devapi_attr.tpp
```diff
@@ -77,9 +77,12 @@ void DeviceAttribute::base_val(T _val)
 	d_state_filled = false;
 	exceptions_flags.set(failed_flag);
 	exceptions_flags.set(isempty_flag);
+	// This constructor is supposed to be used only for enum types since all
+	// the constructors for the other standard Tango types have a specialization defined
```
Ingvord: In this case we should remove template from here and add standard overloading. Should be clear then, should not be?

cppapi/client/devapi_attr.tpp
```diff
@@ -397,7 +408,7 @@ void DeviceAttribute::insert(vector<T> &_datum,int _x,int _y)
 }
 
 template <typename T>
-bool DeviceAttribute::template_type_check(T &_datum)
+bool DeviceAttribute::template_type_check(T &TANGO_UNUSED(_datum))
```
Ingvord: It is better to remove unused parameters from the code. In CLion use CTRL+F6 when cursor is on the method name to refactor method's signature.

cppapi/client/devapi_attr.cpp
t-b: @Ingvord You can do that in C++11 only.

And there it must be writtten as

```cpp
DeviceAttribute::DeviceAttribute(string &new_name, short datum)
:
ext(new DeviceAttributeExt),
DeviceAttribute(new_name.c_str(), datum)
{}
```

Ingvord: @t-b thanks for the hint!

cppapi/client/devapi_attr.tpp
bourtemb: How would you support the case where _val is of type enum Color, or enum MyEnum?
This is why this method is here.

cppapi/client/devapi_attr.tpp
bourtemb: In this case, the parameter is actually useful in a way because his type is used in the method.

cppapi/client/devapi_attr.cpp
bourtemb: Indeed, this works only in C++11. So I think we could do that for Tango 10.
In any case, I agree with Igor we could refactor using a method which would initialize most of the members of this class.
There is a lot of code duplication in this file in all the constructors.
I will propose another Pull Request for this refactoring in order to avoid to pollute this one.
I will propose it just after this PR has been merged.

tiagocoutinho:
This PR seems to have introduced a backward incompatibility.
We have complains in PyTango from taurus users: tango-controls/pytango#271

In PyTango we hid the problem with PR tango-controls/pytango#273 but we are wondering if this is correct.

Would you consider reverting the behaviour to the original or do we live with this?

Thanks for having a look more than 1.5 years later :-)

bourtemb:
@tiagocoutinho, what was done in tango-controls/pytango#273 looks correct to me.
I think this is the way to go.

This backwards compatibility is documented in the release notes and changelog.
It is true that the case when the DeviceAttribute::isempty_flag exception flag is reset before the call is currently not correctly documented. In this case, is_empty was not throwing an exception when invoked from get_type()) and get_type() was returning -1 (when there was no value or invalid data. This -1 return value was not documented anywhere but PyTango knew the trick) and get_type() is now returning DATA_TYPE_UNKNOWN in this specific case.
So now, get_type() never returns -1.

Actually, I think ideally get_type() should not even return an int, but should return an enum CmdArgType but I don't think we will change that in Tango 9 LTS versions.

tiagocoutinho:
Thank you soo much @bourtemb for the complete answer.
It is clear too me know. 
PS: I wish you a profitable tango meeting.

PR #394, opened 2017-09-11
Refactor DeviceAttribute code

bourtemb:


cppapi/client/DeviceAttribute.h
```diff
@@ -1301,6 +1301,7 @@ protected :
 	bool check_for_data();
 	bool check_wrong_type_exception();
 	int  check_set_value_size(int seq_length);
+	void init_common_class_members(const char * name,int dim_x,int dim_y);
```
t-b: @bourtemb Why not make it private?

bourtemb: I have nothing against making it private.
DeviceAttributeHistory is the only class inheriting from DeviceAttribute and this class does not need to use this new method and the other classes inheriting from it in the future will probably not need it either because they ill probably call DeviceAttribute constructors directly in their constructors...
We can make it private.

cppapi/client/DeviceAttribute.h
```diff
@@ -1318,6 +1318,9 @@ protected :
 #else
 	DeviceAttributeExt	            *ext;		// Class extension
 #endif
+
+private:
+    void init_common_class_members(const char * name,int dim_x,int dim_y);
```
Ingvord: I would call it simply *init*. As it is so widely used it will never give any ambiguous and much shorter.

cppapi/client/devapi_attr.cpp
```diff
@@ -542,46 +542,44 @@ DeviceAttribute & DeviceAttribute::operator=(DeviceAttribute &&rval)
...
-	data_format = Tango::FMT_UNKNOWN;
-	d_state_filled = false;
-	exceptions_flags.set(failed_flag);
-	exceptions_flags.set(isempty_flag);
-	ShortSeq = new(DevVarShortArray);
-	ShortSeq->length(1);
-	ShortSeq[0] = datum;
-	data_type = DEV_SHORT;
+    init_common_class_members(new_name.c_str());
+    ShortSeq = new(DevVarShortArray);
```
Ingvord: Is ShortSeq a member? then it must be renamed to *short_seq*. I think we need a holder class here:

```c++
sequence = new HolderClass(datum)
```

*HolderClass* can be templatized in this case: 

```c++
temlate<typename T>
class Holder {
T value; //actual sequence

public:
  template<typename T. typename = std::enable_if<short>>
  HolderClass(short datum){
     value = new ShortSeq(1);
     value.length(1);
     value[0] = datum;
   }
}

```
NOTE this is just a C++-like pseudo code (mostly inspired by Java generics)

Probably this task is for another PR

cppapi/client/DeviceAttribute.h
bourtemb: I called it like that since it does not initialize everything.
ShortSeq is indeed a member.
I agree with you it would make sense to rename it to short_seq, as well as all the other similar members (DoubleSeq, ...).
My CLion instance does not seem to help me much in this task ("Refactor -> Rename" renames it only in DeviceAttribute.h. This is not very helpful), I don't know why...

About the Holder class, this is something that could be investigated in the future but I think there are more urgent tasks right now.

PR #398, opened 2017-09-18
Fix bug leading occasionally Tango 9 DS to segmentation fault

bourtemb:
when an IDLv4 client had subscribed to an attribute and the attribute became
INVALID (quality factor).


cppapi/server/attribute.cpp
```diff
@@ -3561,10 +3558,7 @@ void Attribute::AttributeValue_4_2_AttributeValue_5(const Tango::AttributeValue_
 // First pass the data from one union to another WITHOUT copying them
 //
 
-	if (att_4->quality != Tango::ATTR_INVALID)
```
Ingvord: Less *if*-s - perfect!!!

PR #400, opened 2017-09-26
Fix bug in event name when pushing an error event to IDL5 clients.

bourtemb:
The events were not sent by ZMQ in this case because the name didn't match the
attribute name the client had susbcribed to.
This bug was triggering some API_MissedEvents error events on the client side.
(Missed some events! Zmq queue has reached HWM?)


bourtemb:
A new unit test should be added to test this case.

PR #403, opened 2017-10-23
Attribute: Don't treat an integer as an char*

t-b:
Currently the compiler outputs

attrgetsetprop.cpp: In member function ‘void Tango::Attribute::get_properties(Tango::AttributeConfig_3&)’:
attrgetsetprop.cpp:200:91: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
   conf.event_prop.per_event.period = CORBA::string_dup((const char *)(DEFAULT_EVENT_PERIOD));

and that is justified as DEFAULT_EVENT_PERIOD is an integer.

Using the plain output operator into the stream is the preferred way.


t-b:
@bourtemb Sure, sorry for the messup.

Issue #266, opened 2016-09-03
Please remove zmq.hpp from tango sources


tango-tickets-migrator:
Hello,

tango zmq.hpp superseed the zmq.hpp from the system.

This is bad, so please do not distribut this embeded copy of zmq.hpp.

Cheers

Frederic

Reported by: piccaf

Original Ticket: [tango-cs/feature-requests/125](https://sourceforge.net/p/tango-cs/feature-requests/125)

PR #412, opened 2017-11-06
Tango 9 lts remove zmq hpp

bourtemb:
Remove zmq.hpp from tango-9-lts (#266)
Add CPPZMQ_BASE CMake optional variable


t-b:
@bourtemb Is cppzmq now a required dependecy? If yes cmake should also complain if it could not be found or?

bourtemb:
Well, zmq.hpp file is required.
On Debian it is distributed with zmq Debian packages.
On some other platforms, it might be independent.
Maybe we should add a [CHECK_INCLUDE_FILE](https://cmake.org/cmake/help/v3.0/module/CheckIncludeFile.html) line to ensure it is available on the system and give a nice message explaining this requirement if the file is not found....

t-b:
> Maybe we should add a CHECK_INCLUDE_FILE line to ensure it is available on the system and give a 
> nice message explaining this requirement if the file is not found....

yes that sounds good!

t-b:
@bourtemb THanks, I've just tested it here locally and it works.

bourtemb:
@t-b Thank you very much for your feedback and for having taken the time to test it.

bourtemb:
Closing this PR since it is now handled by PR #421

PR #413, opened 2017-11-08
Tango 9 lts - Windows fixes

NexeyaSGara:
This pull request add the Windows CMakelist support for the lts branch.

We will use Appveyor for the building process with the CMakeList


andygotz:
Does this mean you got CI of LTS working on Windows? Chapeau!

cpp_test_suite/new_tests/cxx_nan_inf_in_prop.cpp
```diff
@@ -5,6 +5,15 @@
...
 #include <iostream>
+#ifdef _WIN32
+#include <limits>
+#ifndef isinf
+#define isinf(x) ((x)!=(x))
+#endif
+#ifndef isnan 
+#define isnan(x) ((x)!=(x)) 
+#endif
+#endif
```
t-b: First of all NaN and infs are not the same.

Second recent enough VS versions have std::isnan and std::isinf defined (since C++11), so no need for supplying them for these versions.

For older VS versions I would use something like [1].

[1]: https://stackoverflow.com/a/20723890

log4tango/tests/clock.hh
```diff
@@ -34,7 +34,11 @@
     typedef long usec_t;    /* number of microseconds since 1970/01/01 */
 #   define INT64_CONSTANT(val)  (val##L)
 #else
-    typedef int64_t usec_t;
+    #ifdef _TG_WINDOWS_
+        typedef uint64_t usec_t;
+    #else
+        typedef __int64 usec_t;
+    #endif
```
t-b: Why are you changing the non windows type here form int64_t to __int64? Most POSIX, aka non windows, platforms support C99 so you can use `int64_t` [1]. It also looks really odd to use different signedness in both cases as well.

[1]: https://stackoverflow.com/a/6013417

cppapi/server/idl/CMakeLists.txt
```diff
@@ -8,16 +8,38 @@ message("Generate tango.h, tangoSK.cpp and tangoDynSK.cpp from idl")
...
+ set_source_files_properties(tangoDynSK.cpp PROPERTIES COMPILE_DEFINITIONS _OMNIORB_STUB_DLL)
+ set_source_files_properties(tangoDynSK.cpp PROPERTIES COMPILE_DEFINITIONS __NT__)
+ set_source_files_properties(tangoSK.cpp PROPERTIES COMPILE_DEFINITIONS _OMNIORB_STUB_DLL)
+ set_source_files_properties(tangoSK.cpp PROPERTIES COMPILE_DEFINITIONS __NT__)
+        if(${CMAKE_SIZEOF_VOID_P} EQUAL 8)
+execute_process(COMMAND ${OMNIIDL_PATH}/omniidl.exe -I${IDL_PKG_INCLUDE_DIRS} -bcxx -Wbh=.h -Wbs=SK.cpp -Wbd=DynSK.cpp -Wba ${IDL_PKG_INCLUDE_DIRS}/tango.idl
+                WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+                RESULT_VARIABLE FAILED)
+        else()
+execute_process(COMMAND ${OMNIIDL_PATH}/omniidl.exe -I${IDL_PKG_INCLUDE_DIRS} -bcxx -Wbh=.h -Wbs=SK.cpp -Wbd=DynSK.cpp -Wba ${IDL_PKG_INCLUDE_DIRS}/tango.idl
```
bourtemb: It looks like there is no difference between what is executed here when (${CMAKE_SIZEOF_VOID_P} EQUAL 8) and in the else.
It looks like this if could be removed and this code simplified.

NexeyaSGara: Surely.

I prepared this in case the 64 omniidl could be different.

CMakeLists.txt
```diff
@@ -29,28 +31,44 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/log4tango/include)
 #source code
 add_subdirectory("log4tango")
 add_subdirectory("cppapi")
-add_subdirectory("cpp_test_suite")
+#add_subdirectory("cpp_test_suite")
```
bourtemb: We can tolerate the tests to be deactivated on Windows for the moment, but they should be compiled on Linux and they should be executed on Travis CI

configure/CMakeLists.txt
```diff
@@ -1,23 +1,37 @@
...
 if(NOT CMAKE_BUILD_TYPE)
-    message("No build type specified - default is DEBUG")
-    set(CMAKE_BUILD_TYPE DEBUG)
+    message("No build type specified - default is RELEASE")
+    set(CMAKE_BUILD_TYPE RELEASE)
 endif()
 
 if(IDL_BASE)
-    set(IDL_PKG_INCLUDE_DIRS ${IDL_BASE}/include)
+    message("Using IDL_BASEIDL_BASE=$ENV{IDL_BASE}")
```
bourtemb: This is breaking something on Linux.
The user should be able to pass -DIDL_BASE=<IDL_BASE_folder> to cmake without having to specify the IDL_BASE environment variable.
I would suggest to fix the typo in the message and also discourage the use of $ENV...
Can't we use ${IDL_BASE} as it used to be instead of $ENV{IDL_BASE}?

The same applies for OMNI_BASE and ZMQ_BASE below.

appveyor.yml
```diff
@@ -0,0 +1,247 @@
...
+  # Zmq
+  - cmd: cd "C:\projects\"
+  - cmd: md zeromq
+  - cmd: cd "C:\projects\"
+  - if %ARCH%==win32-msvc9 appveyor DownloadFile https://ci.appveyor.com/api/buildjobs/kif9t1ni9dorv7ai/artifacts/zmq-4.0.5_%ARCH%.zip
+  - if %ARCH%==x64-msvc9 appveyor DownloadFile https://ci.appveyor.com/api/buildjobs/v08rh5ojx3uljbm0/artifacts/zmq-4.0.5_%ARCH%.zip
+  - if %ARCH%==win32-msvc10 appveyor DownloadFile https://ci.appveyor.com/api/buildjobs/bfw1w51dmwb4t1s8/artifacts/zmq-4.0.5_%ARCH%.zip
+  - if %ARCH%==x64-msvc10 appveyor DownloadFile https://ci.appveyor.com/api/buildjobs/dv6koti648n78wru/artifacts/zmq-4.0.5_%ARCH%.zip
+  - if %ARCH%==x64-msvc12 appveyor DownloadFile https://ci.appveyor.com/api/buildjobs/92i5i0466vrsf8uf/artifacts/zmq-4.0.5_%ARCH%.zip
+  - if %ARCH%==x64-msvc14 appveyor DownloadFile https://ci.appveyor.com/api/buildjobs/jlsd4d38c425kumq/artifacts/zmq-4.0.5_%ARCH%.zip
```
bourtemb: As we said during the last Tango kernel teleconference, the repositories to build ZMQ and omniORB should be on GitHub tango-controls organization so we could have full control on how to build these artifacts.
This can be done eventually after accepting this PR. But another PR will be necesseray in order to update all these download links.

configure/CMakeLists.txt
```diff
@@ -1,23 +1,37 @@
 include(FindPkgConfig)
 
 if(NOT CMAKE_BUILD_TYPE)
-    message("No build type specified - default is DEBUG")
-    set(CMAKE_BUILD_TYPE DEBUG)
+    message("No build type specified - default is RELEASE")
+    set(CMAKE_BUILD_TYPE RELEASE)
```
bourtemb: I don't mind changing the default build type but I think we should rather use Release and Debug words everywhere instead of RELEASE and DEBUG because they are the standard ones as described on CMAKE_BUILD_TYPE documentation (https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html).
We have to be aware that STREQUAL and MATCHES are case sensitive in CMake.
If we agree on doing this change, we will need to update line 148 from configure/CMakeLists.txt too and replace
`if(CMAKE_BUILD_TYPE MATCHES RELEASE)`
with
`if(CMAKE_BUILD_TYPE MATCHES Release)`
I think we should also update the README.md and INSTALL.md files to be consistent (Replace RELEASE|DEBUG with Release|Debug in the CMake command example.

configure/CMakeLists.txt
```diff
@@ -31,30 +45,102 @@ else(OMNI_BASE)
...
+    set(PTHREAD_WIN_PKG_INCLUDE_DIRS $ENV{PTHREAD_WIN}/include)
+    
+    if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+        set(PTHREAD_WIN_PKG_LIBRARIES_DYN "pthreadVC2d.lib")
+        set(PTHREAD_WIN_PKG_LIBRARIES_STA "pthreadVC2-sd.lib")
+    else()
+        set(PTHREAD_WIN_PKG_LIBRARIES_DYN "pthreadVC2.lib")
+        set(PTHREAD_WIN_PKG_LIBRARIES_STA "pthreadVC2-s.lib")
+    endif()
+else(IDL_BASE)
```
bourtemb: This else looks like a copy/paste error and actually breaks something on Linux in some cases

log4tango/tests/clock.hh
bourtemb: This change is breaking the Linux compilation.
If one looks at the logs of Travis CI builds, one can see there are errors when compiling this file due to this change.

cppapi/server/utils.cpp
```diff
@@ -1919,44 +1924,83 @@ void Util::server_init(TANGO_UNUSED(bool with_window))
 void Util::server_perform_work()
 {
 	if (ev_loop_func != NULL)
+
```
bourtemb: All these new empty lines should be removed on this file

cppapi/server/tango_config.h
```diff
@@ -230,6 +230,8 @@
 	#pragma warning(disable : 4355)
 	#pragma warning(disable : 4715)
     	#pragma warning(disable : 4786)
+    	#pragma warning(disable : 4267)
```
bourtemb: Could you please add a comment here and explain what warnings are disabled by these 2 new lines, please?

cppapi/server/idl/CMakeLists.txt
bourtemb: Since it appears this is not different, could you please remove this if(${CMAKE_SIZEOF_VOID_P} EQUAL 8)?
Thanks! :-)

appveyor.yml
NexeyaSGara: I think so, for the moment, theses artefacts are OK. We will just update them when you will have all setted up

CMakeLists.txt
NexeyaSGara: Done added a if not win32

configure/CMakeLists.txt
NexeyaSGara: No problem for me, I will need to update more files since I'm using this "if" in some places

configure/CMakeLists.txt
NexeyaSGara: clearly a bad copy

appveyor.yml
bourtemb: I get this CMake warning with your change:
CMake Warning (dev) in CMakeLists.txt:
  A logical block opening on the line
    /users/bourtemb/GIT/tango-controls/cppTango4Windows/CMakeLists.txt:34 (if)
  closes on the line
    /users/bourtemb/GIT/tango-controls/cppTango4Windows/CMakeLists.txt:36 (endif)
  with mis-matching arguments.

endif(WIN32) should be replaced with endif(NOT WIN32) in this case.

cppapi/server/tango_config.h
NexeyaSGara: 4267 --> var' : conversion from 'size_t' to 'type', possible loss of data --> lot of them
4244 --> conversion' conversion from 'type1' to 'type2', possible loss of data --> lot of them

appveyor.yml
NexeyaSGara: ok

cppapi/server/tango_config.h
bourtemb: Sorry, my comment was not clear enough... I meant could you add comments in tango_config.h on these new #pragma warning(disable...) lines?

cpp_test_suite/new_tests/cxx_nan_inf_in_prop.cpp
```diff
@@ -205,12 +214,12 @@ class NanInfInPropSuite: public CxxTest::TestSuite
...
 		Tango::DevFloat float_val;
 		TS_ASSERT(db_data_r[0] >> float_val);
-		TS_ASSERT(std::isnan(float_val));
+		TS_ASSERT(isnan(float_val));
 
 		// Extract as double:
 		Tango::DevDouble double_val;
 		TS_ASSERT(db_data_r[0] >> double_val);
-		TS_ASSERT(std::isnan(double_val));
+		TS_ASSERT(isnan(double_val));
```
bourtemb: This change is breaking the travis CI build with the following error:

/home/tango/src/cpp_test_suite/cxxtest/include/../../new_tests/cxx_nan_inf_in_prop.cpp:222:29: error: call of overloaded 'isnan(Tango::DevDouble&)' is ambiguous

NexeyaSGara: since it's on the test part, I can revert and we will see test after

bourtemb: Nice try! ;-) but this line was not the only one like that... There are many errors similar to this one.

NexeyaSGara: I changed all in the file ;)

PR #417, opened 2017-12-05
Enable Travis debian7

bourtemb:
Enable Travis CI tests for Debian 7.
CMake v3.10 is downloaded and installed using the installer script provided by CMake.org


bourtemb:
I would be in favour of adding Debian 9 as well, yes! :-)

Ingvord:
@bourtemb , I think we can merge this now and add Debian 9 in a separate PR

PR #421, opened 2017-12-07
Tango 9 lts remove zmq hpp

bourtemb:
Remove zmq.hpp from tango-9-lts (#266)
Add CPPZMQ_BASE optional CMake variable

This Pull Request is the same as #412 but it seems there is bug in Github which makes it think there are still some conflicts, even after having resolved them with the web interface!


bourtemb:
On Debian 7 and Debian 9, the tests are failing with the versions from the official repos, as you can see in the results of commit [4d53b31](https://github.com/tango-controls/cppTango/pull/421/commits/4d53b3141665ddf6b756fdb818163ca7fc83c2fb).
On Debian 8, you are right, we could use directly what we get from the official repo.

PR #404, opened 2017-10-25
Yet another attempt to improve event subscription with minimal changes to the existing code base

Ingvord:
see #315, #385

Issue #359, opened 2017-04-11
Change event subscription blind to change events right after device server restart


ghost:
[moved here from https://sourceforge.net/p/tango-cs/bugs/834/]

**Background**
If a device server is restarted, all connected event subscribers will try to reconnect to it after a timeout. Once the client has reconnected, it will receive the latest value of the attribute. So far, so good.

**Expected behaviour:**
After the connection has been reestablished, from there on the event subscriber should receive every attribute value change that occurs (e.g. a change event on the State attribute).

**Actual behaviour:**
After the event subscriber has reconnected to the device server, there is a time window in which the event subscriber will not be informed about attribute changes.
E.g. when subscribing to change events on the State attribute, any State changes won't be reported to the event subscribers during that time window. The length of the time window seems to depend on the polling period. A polling period of 10s on State makes the event channel blind to State changes for at least a couple of seconds.

In other words, the event subscriber gets unsynchronized with the device server. And it will remain unsynchronized for potentially a long time, until another attribute change on the server is prompted.

**Steps to reproduce:**
If the attribute is configured with a large polling period (e.g. the default 3000ms), the bug is easy and reliably to reproduce.
1. Tested with Tango 9.2.5a, OmniORB 4.2.1, libzmq v4.1.4, PyTango 4.2.1 on Ubuntu 16.04 [amd64]
2. start TangoTest device server ( (Mthreaded_impl: false; State polling enabled and set to 10s)

`Ready to accept request`

3. start attached event_bug_demo_client.py <device>
[this script connects to State change events and prints out the events]

`State:  UNKNOWN`

4. Quit and restart TangoTest
wait until the event subscriber has reestablished the connection

```
Error state: Event channel is not responding anymore, maybe the server or event system is down
State:  UNKNOWN
```
5. Now quickly execute attached event_bug_demo_switch_state.py <device>
[this will call SwitchStates on TangoTest, causing State to go from UNKNOWN to RUNNING]

If done quickly enough, the event_bug_demo_client.py script will stay stuck at UNKNOWN forever.

Please let me know if there is anything further I can provide to help analyzing this issue.

forouherCycle:
And here are the python test scripts mentioned above:

[python_test_scripts.zip](https://github.com/tango-controls/cppTango/files/912264/python_test_scripts.zip)

Note: I've created the ticket here because it looks to me like a bug in the C++ Tango device server.

bourtemb:
Thank you very much for re-creating the ticket here.
[...]

PR #423, opened 2017-12-14
Try to fix event system in Tango-9-LTS

Ingvord:
As discussed in #404


Ingvord:
@bourtemb , looks like your were [right](https://github.com/tango-controls/cppTango/pull/404#issuecomment-351440420). Life is really easier now :)

I have tested double event after re-connection (spectrum and scalar) and it works as expected

What we also need to do:

- [ ] is to test X-version case (pre-936 clients should work as before, i.e. do not work in case netalias etc)
- [ ] add integration test for scalar case or proof that attribute type does not affect the behavior

CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ include(CTest)
 enable_testing()
 #need to define the version of the library
 set(MAJOR_VERSION "9")
-set(MINOR_VERSION "2")
+set(MINOR_VERSION "3")
```
bourtemb: Going from 9.2.6 to 9.3.6?

cppapi/client/event.cpp
```diff
@@ -1823,6 +1832,42 @@ int EventConsumer::connect_event(DeviceProxy *device,
...
+	if (tango_lib_ver >= 936)
+	{
+		received_from_admin.channel_name = (dvlsa->svalue[dvlsa->svalue.length() - 1]);
+	}
+	else
+	{
+		received_from_admin.channel_name = std::move(adm_name);
+	}
+
+	assert(not(received_from_admin.event_name.empty()));
```
bourtemb: We should not follow Codacy recommendations here (or maybe we should find a way to tune them so it does not advise us to use this keyword).
"not"  keyword is not recognized by Visual C++ compilers.
This is why the appveyor build is broken.

CMakeLists.txt
Ingvord: Yep, according to [semantic versioning](https://semver.org/) I would consider this change as _MINOR version when you add functionality in a backwards-compatible manner_

Ingvord: Or do you mean we should drop PATCH_VERSION to `0`?

Ingvord:
I have tested the X-version configurations:

|                  | New Server | Old Server
| -------------- | ---------------- | --------------
| New Client |        V           | Events - OK; heartbeat - NOK
| Old Client  |          V          |          V

bourtemb:
Here are the results of my tests with new clients and new server: 
[EventsTests-reynald-tests-19Dec2017.zip](https://github.com/tango-controls/cppTango/files/1572420/EventsTests-reynald-tests-19Dec2017.zip)
I tested lines 2-9 and 17-24.
As you will see in the spreadsheet, there are still some cases to fix: 
* when TANGO_HOST=localhost on the client side and the subscription is done with an attribute name which is not an fully qualified attribute name (FQAN).
* when the TANGO_HOST on the server side is set to 127.0.0.1:<port-number>, there are some problems when some remote clients (running on another host) are trying to reconnect. I have observed some heartbeat errors in this case and sometimes some crashes of the client (in cppapi/client/eventkeepalive.cpp:187 . It seems ipos->second.adm_device_proxy is uninitialized or null in this case).
* when the TANGO_HOST on the client or FQAN used by the client is using <netalias-without-domain-name>:<port-number>, the name in the events is not correct in this case.

All the tests have been done while playing with TangoTest's float_scalar attribute.

During my tests, I also saw bug #359. In my case float_scalar attribute was polled at 10s.
The bug occurred when restarting the server and immediately after, setting float_scalar to 1000 for instance. I saw the client reconnecting with the synchronous call and displaying the following attribute values:
* Element number[0] = 0     (float_scalar read value)
* Element number[1] = 0     (float_scalar write value)
I saw a gap of about 20 seconds between the timestamp of this event and the timestamp of the next event received.
I was also reading this attribute with jive's device panel in the meantime with source = CACHE_DEVICE and I clearly saw that there was one value which was missing in the events sent (one polling loop).

I also saw the heartbeat issue between a new client and an old server.

From my experience, I know there are also some reconnection issues when a server instance is renamed (so the admin device name changes) or when a server is upgraded from an old IDL version to a new one.
I don't whether we could take this opportunity to find a solution for these cases too but we should keep this in mind because there might be simple solutions to solve these issues.

Ingvord:
@bourtemb ,Hi I have fixed the case with *netalias* -> *netalias.domain*. I have not got a chance to look very carefully into all of the cases though. But maybe this is enough for you to deploy now. The rest can be fixed later.

Ingvord:
@bourtemb , I have tried to fix AppVeyor. No luck, unfortunately, there is no more compilation errors, but for some reason it fails anyway. Maybe @NexeyaSGara could help?

cppapi/client/attr_proxy.cpp
```diff
@@ -294,12 +294,12 @@ AttributeProxy::AttributeProxy(const AttributeProxy &prev):ext(Tango_nullptr)
 #ifdef HAS_UNIQUE_PTR
     if (prev.ext.get() != NULL)
     {
-        ext.reset(new AttributeProxyExt);
+        ext.reset(new AttributeProxyExt{prev.get_user_defined_name()});
```
bourtemb: Curly brackets should be replaced with parentheses here

cppapi/client/attr_proxy.cpp
```diff
@@ -370,13 +370,13 @@ AttributeProxy &AttributeProxy::operator=(const AttributeProxy &rval)
 
 #ifdef HAS_UNIQUE_PTR
         if (rval.ext.get() != NULL)
-            ext.reset(new AttributeProxyExt);
+            ext.reset(new AttributeProxyExt{rval.get_user_defined_name()});
```
bourtemb: Curly brackets should be replaced with parentheses here

bourtemb:
The problem with appveyor comes from the fact the library version changed from 9.2.6 to 9.3.0.
The generated files are named <something>9.3.0<something> whereas appveyor is expecting <something>9.2.6<something>.
@NexeyaSGara, it would be good to update appveyor.yml file in order to have the library version automatically updated in this file and we should remove all the hard coded library versions in this file, or in the worst case, have it only once in the file.

Ingvord:
@bourtemb , good catch!!!

NexeyaSGara:
A quick fix will be to add :
- cmd: set TANG_VER=9.3.0

and then use %TANGO_VER% in the "after_build" section when I have to copy the debug generated files to the directory that will be scanned for artifacts:
ex:
libtango_%TANGO_VER%_Windows7.1SDK_x64_debug.msi libtango_%TANGO_VER%_v100_x64_debug.msi

So just need to edit it when a new build is comming

bourtemb:
I confirm the heartbeat problem in the case of new client subscribing to events on attributes from an old server.
This is actually a blocking issue because we will definitely encounter this use case in our control system.
A simple example are the HDB++ subscribers which will subscribe to archive events on attributes from some Tango 8 devices.

Ingvord:
@bourtemb looks like this is fixed in 5d61a7c

bourtemb:
Codacy is right, intr_change is no longer initialized in void ZmqEventSupplier::push_event(DeviceImpl *device_impl,string event_type,...)

Ingvord:
Won't it be **false** by default?

bourtemb:
> Won't it be false by default?
No, reading an uninitialized variable is undefined behaviour.

bourtemb:
My first tests "new client -> old server" seem to confirm the heartbeat problem in this case is fixed.
There are still a few cases which do not work perfectly as I described before (mainly issues with event name passed in EventData to the client, when _TANGO_HOST=\<network_alias_name_without_domain_name\>:\<port\>_ for instance ) but still, the current situation in this PR seems to be a nice improvement compared to what we had before.
We could fix the other cases in other PRs.
Could you please start to clean up the code of this PR (remove debug cout traces, fix indentation, typos...)?
We will then proceed to a last code review and we will hopefully merge it after that.
Thank you very much for your work!

cppapi/client/attr_proxy.cpp
```diff
@@ -1358,7 +1358,10 @@ int AttributeProxy::subscribe_event (EventType event, CallBack *callback,
 	int ret;
 	try
 	{
-        ret = api_ptr->get_zmq_event_consumer()->subscribe_event(dev_proxy,attr_name, event, callback, filters, stateless);
+        vector<string> non_const_filters = const_cast<vector<string> &>(filters);
+        non_const_filters.clear();
+        non_const_filters.push_back(get_user_defined_name());
+        ret = api_ptr->get_zmq_event_consumer()->subscribe_event(dev_proxy,attr_name, event, callback, non_const_filters, stateless);
```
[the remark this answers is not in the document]
bourtemb: This trick does not seem to be implemented when subscribing using a DeviceProxy...
The problem seems to be that the db_host member from DeviceProxy is set to the real Tango database hostname.
We should change this behaviour if we want to get a better behaviour with the network aliases.
The question is: what could be the potential side effects of such a change?

RELEASE_NOTES
```diff
@@ -1,3 +1,11 @@
+9.3.0
+=====
+
+1. Fix FQDN events.
+   Limitations:
+     - this fix uses a dirty trick which breaks backward compatibility with notifd;
```
bourtemb: Actually this is not really true. Events are still working with notifd, even with this version.
The only feature which might not work is the event filtering feature from the notifd events, but it seems like this feature is no longer supported since Tango 8 (http://tango-controls.readthedocs.io/en/latest/development/client-api/cpp-client-programmers-guide.html?highlight=Event%20filtering#event-filtering-removed-in-tango-release-8-and-above). So we are good here it seems ;-).

cppapi/client/event.cpp
```diff
@@ -1822,6 +1822,54 @@ int EventConsumer::connect_event(DeviceProxy *device,
...
+    if(filters.size() == 1)
+        return filters[0];// here filters[0] is expected to be user defined attribute name
+
+
+	size_t pos = local_callback_key.rfind('.');//remove event_type e.g. .idl5_change
+	return local_callback_key.substr(0, pos);
+}
+
+void Tango::EventConsumer::initialize_received_from_admin(const Tango::DevVarLongStringArray *dvlsa,
+														  const string &local_callback_key,
```
bourtemb: I would put spaces here instead of tabs

cppapi/client/event.cpp
```diff
@@ -1822,6 +1822,54 @@ int EventConsumer::connect_event(DeviceProxy *device,
...
+    if (server_tango_lib_ver >= 930)
+    {
+        received_from_admin.event_name = (dvlsa->svalue[dvlsa->svalue.length() - 2]);
+    }
+    else
+    {
+        received_from_admin.event_name = local_callback_key;
+    }
+
+    if (server_tango_lib_ver >= 930)
```
bourtemb: I think this can be grouped with the previous if which is on line 1842

Ingvord: I intentionally split them, because these are two distinct actions. Keep them separate will help in the future refactoing

RELEASE_NOTES
Ingvord: OK, good to know :) Then we have to _really_ remove them at some point...

Ingvord:
Squeeze?! O_o I doubt anyone will ever try to compile anything on that... but if you insist

PR #429, opened 2018-01-18
Tango 9 can again be built with older compilers

bourtemb:
(Tested with gcc 3.4.6 and gcc 4.4.5)

### DIFF
--- a/configure/CMakeLists.txt
+++ b/configure/CMakeLists.txt
@@ -159,7 +159,16 @@ else(WIN32)
 endif(WIN32)
 
 if(NOT WIN32)
-    add_compile_options(-std=c++11)
+    include(CheckCXXCompilerFlag)
+    CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
+    CHECK_CXX_COMPILER_FLAG("-std=c++0x" COMPILER_SUPPORTS_CXX0X)
+    if(COMPILER_SUPPORTS_CXX11)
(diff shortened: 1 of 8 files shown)


bourtemb:
I agree with you @t-b. We should try to set up Travis to compile with older compilers in order to avoid to introduce again some C++ 11 code which is not supported by these old compilers.

bourtemb:
This does not mean we cannot put any C++ 11 code in Tango 9 LTS. If we put C++ 11 code, it should be surrounded by some #ifdef directives and an alternative code for older compilers should be there too.

andygotz:
Thanks for reactivating this option again. I have been thinking about supporting old compilers and after checking I found that gcc 3.4 was released in 2006 while gcc 4.4 was released in 2010. I wonder how long we need to support these compilers? The first version of gcc which supported C++11 was gcc 4.7 which was released in 2012. We plan to support LTS-9 for another 5 years at least. Would it not be reasonable to say we need C++11 support at some point? It will simplify the code and the development.

bourtemb:
I just wanted to check how difficult it would be to support these compilers again...
It appeared it was not too difficult so I did the job ;-) ...
Of course, supporting them will generate more work on our side, especially when including changes coming from Tango 10 development branch (master branch) into Tango 9 LTS.

t-b:
@andygotz 

> The first version of gcc which supported C++11 was gcc 4.7 which was released in 2012. 

Using gcc 4.7 with C++11 is no fun. According to https://gcc.gnu.org/projects/cxx-status.html#cxx11 one should at least use 4.8. From my experience it starts to get usable with 4.9 (mainly due to STL issues in earlier versions).

Issue #424, opened 2017-12-22
CPU load when device has large number of attributes


andygotz:
An issue has been reported by the SKA consortium when running a Tango device server  with hundreds of attributes on a small CPU - the CPU load goes up to 100% for their CPU (a Taskit Stamp9G20, SPFC) when one or more external Tango clients connect to this server. This issue could be related to the way attributes are accessed in the Tango library. This issue will be used to try to reproduce the CPU load problem and to fix it if it is in the library.

Attached are some plots (produced with nmon) provided by Isak Theron (SKA):
![image001](https://user-images.githubusercontent.com/1167099/34307535-e8c56dac-e748-11e7-914d-a329cb45a3a7.png)
and the source code used for the device servers used to demonstrate the issue.
[spfcPerformance_200.tar.gz](https://github.com/tango-controls/cppTango/files/1583151/spfcPerformance_200.tar.gz)
[pyPerformance_200.tar.gz](https://github.com/tango-controls/cppTango/files/1583152/pyPerformance_200.tar.gz)

andygotz:
I have run some profiling on the Spf C++ server example (see attachment above) and the first results show that a lot of calls are made and time spent in iterating over attributes. The main culprit seems to be the function Tango::MultiAttribute::get_attr_ind_by_name(char const*). There are also many calls to tolower() (3rd most called function!). Most of the time is spent in libzmq which I do not understand as this example does not trigger events as far as I can see and atkpanel is reading via polling. Tests were run on a laptop with Ubuntu 17.10 and 1 client (atkpanel) connected to the device. Polling was not configured. 

Output from callgrind_annotate attached.
[Spf_callgrind.log](https://github.com/tango-controls/cppTango/files/1598117/Spf_callgrind.log)

andygotz:
After playing around a bit with valgrind's callgrind I think I found the proof that the cpu bottleneck is in get_attr_by_name() as seen in this output from kcachegrind:
![kcachegrind](https://user-images.githubusercontent.com/1167099/34564435-003f9454-f157-11e7-8ba5-e88a1356cee0.png)
To get this output you need to run the device server as follows:
`valgrind --tool=callgrind --callgrind-out-file=callgrind-cpu.out bin/Spf test`
and look out the output using kcachegrind:
`kcachegrind callgrind-cpu.out`
I tried google's perftools as well but they were not so useful because of the multi threading of the device server.

bourtemb:
[...]

PR #430, opened 2018-01-31
Add a map in MultiAttribute object to improve performances (#424)

bourtemb:
This solution is using an std::map.
Using an unordered_map if the compiler supports it could slightly improve the performances even more.
The gain is already big using an std::map.

### DIFF
--- a/cppapi/server/multiattribute.cpp
+++ b/cppapi/server/multiattribute.cpp
@@ -811,10 +811,7 @@ void MultiAttribute::add_attribute(string &dev_name,DeviceClass *dev_class_ptr,l
 			Attribute * new_attr = new WAttribute(prop_list,attr,dev_name,index);
 			attr_list.insert(ite,new_attr);
 			index = attr_list.size() - 3;
-			MultiAttributeExt::AttributePtrAndIndex mapElement;
-			mapElement.att_ptr = new_attr;
-			mapElement.att_index_in_vector = index;
-			ext->attr_map[new_attr->get_name_lower()] = mapElement;
+			ext->put_attribute_in_map(new_attr,index);
 		}
 	}
 	else
@@ -830,10 +827,7 @@ void MultiAttribute::add_attribute(string &dev_name,DeviceClass *dev_class_ptr,l
 			Attribute * new_attr = new Attribute(prop_list,attr,dev_name,index);
 			attr_list.insert(ite,new_attr);
 			index = attr_list.size() - 3;
-			MultiAttributeExt::AttributePtrAndIndex mapElement;
-			mapElement.att_ptr = new_attr;
-			mapElement.att_index_in_vector = index;
-			ext->attr_map[new_attr->get_name_lower()] = mapElement;
+			ext->put_attribute_in_map(new_attr,index);
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


cppapi/server/multiattribute.cpp
```diff
@@ -1756,7 +1840,28 @@ void MultiAttribute::add_alarmed_quality_factor(string &status)
...
+//
+// description :
+//		Add given attribute to the attribute list vector and map
+//
+// argument:
+//		in :
+//			- att : The newly configured attribute
+//
+//-------------------------------------------------------------------------------------------------------------------
+void MultiAttribute::add_attr(Attribute *att)
```
Ingvord: I think this can be overloaded with `add_attr(Attribute *att, size_t index)` and used to prevent code duplications above.

cppapi/server/multiattribute.cpp
```diff
@@ -936,8 +964,13 @@ void MultiAttribute::add_fwd_attribute(string &dev_name,DeviceClass *dev_class_p
 	vector<Attribute *>::iterator ite;
 	ite = attr_list.end() - 2;
 
-	attr_list.insert(ite,new FwdAttribute(prop_list,*new_attr,dev_name,index));
+	Attribute * new_fwd_attr = new FwdAttribute(prop_list,*new_attr,dev_name,index);
```
Ingvord: Seems to me as a code duplication. Must be extracted into a dedicated method overloaded from `add_attr`, see below

cppapi/server/multiattribute.cpp
```diff
@@ -786,26 +802,38 @@ void MultiAttribute::add_attribute(string &dev_name,DeviceClass *dev_class_ptr,l
...
 		{
-			attr_list.push_back(new Attribute(prop_list,attr,dev_name,index));
+			Attribute * new_attr = new Attribute(prop_list,attr,dev_name,index);
+			add_attr(new_attr);
 			index = attr_list.size() - 1;
 		}
 		else
 		{
-			attr_list.insert(ite,new Attribute(prop_list,attr,dev_name,index));
+			Attribute * new_attr = new Attribute(prop_list,attr,dev_name,index);
```
Ingvord: Seems to me as a code duplication. Must be extracted into a dedicated method overloaded from add_attr, see below

cppapi/server/multiattribute.cpp
```diff
@@ -786,26 +802,38 @@ void MultiAttribute::add_attribute(string &dev_name,DeviceClass *dev_class_ptr,l
...
 		{
-			attr_list.push_back(new WAttribute(prop_list,attr,dev_name,index));
+			Attribute * new_attr = new WAttribute(prop_list,attr,dev_name,index);
+			add_attr(new_attr);
 			index = attr_list.size() - 1;
 		}
 		else
 		{
-			attr_list.insert(ite,new WAttribute(prop_list,attr,dev_name,index));
+			Attribute * new_attr = new WAttribute(prop_list,attr,dev_name,index);
```
Ingvord: Seems to me as a code duplication. Must be extracted into a dedicated method overloaded from add_attr, see below

cppapi/server/multiattribute.cpp
Ingvord: After a little bit more thinking I think it would be better to split this method into two:

`size_t add_attr(Attribute)`, where size_t is an index of the newly added attribute, may be passed in overloaded method

`void put_attr_into_map(Attribute, index)`

And finally aggregate these two in a third one:

```
size_t ndx = add_attr(attr);
put_attr_into_map(attr, ndx)
```

bourtemb: I added put_attribute_in_map() method to reduce code duplication... 
add_attr() method is used in some other parts of the code and already existed before this PR so I think we should stick with the same interface for this method.
I could rename put_attribute_in_map in put_attr_into_map if you prefer...

Ingvord: @bourtemb that is just fine keep it as it is now!

PR #431, opened 2018-02-01
Get exception message from event callback

t-b:


t-b:
@bourtemb Good point. Done.

PR #432, opened 2018-02-02
Doxygen: Add searchbar

t-b:
If you think this is a good change then please also update https://tango-controls.github.io/cppTango-docs/.


t-b:
@bourtemb It is very responsive here. But I also do have a powerful machine.

PR #435, opened 2018-02-14
Add noreturn

t-b:


cppapi/server/tango_config.h
```diff
@@ -81,11 +81,15 @@
...
-			#define WIN32_VC12
-        #endif   // VC8+/VC9/VC10/VC11/VC12
+        #elif ((_MSC_VER >= 1700) && (_MSC_VER < 1800))
+            #define WIN32_VC11
+        #elif ((_MSC_VER >= 1800) && (_MSC_VER < 1900))
+            #define WIN32_VC12
+        #elif ((_MSC_VER >= 1900) && (_MSC_VER < 1910))
+            #define WIN32_VC14
+        #elif (_MSC_VER >= 1910)
+            #define WIN32_VC141
```
bourtemb: It looks like there is no added value in adding a special case for VC141 for the moment so I would keep it simple for the moment and define WIN32_VC14 if (_MSC_VER >= 1900).

t-b:
@bourtemb Sure. Done.

cppapi/server/tango_config.h
```diff
@@ -153,6 +153,7 @@
                 #define HAS_THREAD
                 #define HAS_TYPE_TRAITS
                 #define HAS_VARIADIC_TEMPLATE
+                #define HAS_ATTRIBUTE_SPECIFIERS
```
bourtemb: According to the comment on line 142, shouldn't HAS_ATTRIBUTE_SPECIFIERS be defined for GCC >= 4.8 instead of >4.3?

bourtemb: @t-b , this is probably the reason why Travis CI tests failed on Debian 7, where the compiler is gcc 4.7.2.

t-b: @bourtemb Yes you are totally right. Looks like my conflict resolution messed that up again. Fixed.

PR #441, opened 2018-03-03
Compile flags

t-b:
I'll revise the warning flags supression for compiling on windows the next time.


configure/CMakeLists.txt
```diff
@@ -178,7 +178,7 @@ if(NOT WIN32)
             add_definitions(-D_FORTIFY_SOURCE=3)
         else()
             message("setup for debug build ...")
-            add_compile_options(-O0 -Wall -Wextra)# -pedantic
+            add_compile_options(-O0 -Wall -Wextra -pedantic)
```
bourtemb: Adding this -pedantic flag will generate errors when compiling Tango library Debug version on very old compilers like gcc 3.4.6 for instance (default version on Redhat Enterprise 4, yes we still have some VME running this version!).
We decided during the last Tango kernel teleconf that we should not provide such a high level of support for these very old compilers. If someone needs to compile the debug version with such an old version of gcc, it should simply remove this flag manually so I think we can accept your PR.

t-b: @bourtemb One someone wants to use an ancient compiler I would say adding compiler version logic as described in https://cmake.org/cmake/help/v3.11/variable/CMAKE_LANG_COMPILER_VERSION.html or https://stackoverflow.com/questions/435708/any-way-in-cmake-to-require-gcc-version-4 is the way to go.

Thanks for merging!

PR #442, opened 2018-03-10
tango_config.h: Add clang detection

t-b:
According to [1] clang 3.3 has full C++11 support and 3.4 full C++14
support. So by requiring version 4 or higher we are on the safe side.

[1]: https://clang.llvm.org/cxx_status.html


t-b:
I found that bug due to clang 6.0 complaining about deprecated auto_ptr which it should have never seen.

t-b:
@bourtemb Thanks! So we are good to go here?

bourtemb:
@t-b  I think so... Let's merge this one! :-)

Issue #444, opened 2018-04-03
push_change_event() - events stop being issued after some time


mkomel:
We've noticed a problem where the events from some of our device servers stop coming to the clients some time after startup.
By troubleshooting and testing we were able to create a minimalistic setup with which we can reliably reproduce the problem: [simpledevice.zip](https://github.com/tango-controls/cppTango/files/1871905/simpledevice.zip)

The zip file contains a simple device server and two types of clients (one is PyTango-based, the other written in c++ by using tango libraries..  we experience the same issue with both).
The device server contains a single `DevULong `attribute `simplepushedvalue`, which is actively being pushed from code via `push_change_event()`. The attribute is configured with `set_event_abs_change("1")` and `set_change_event(true, true);` (note that the result is the same if we use `set_change_event(true, false);` ).
The sample should be set up with 2 device server instances, each one with a single device:
- device server instance `simple1 `=> `simple/test/1`
- device server instance `simple2 `=> `simple/test/2`

After starting both device server instances, the client (or both clients) should be started, which will show the events as they come in. 
Approximately after 8-9minutes (interestingly it is reliably after cca.8-9minutes), the events from one of the device servers will stop coming to the client (at least this is how we experience the issue).
It's interesting to note that if the affected device server is then stopped, the client will correctly detect `API_EventTimeout`. And if that device server is then restarted, the events will again correctly start to arrive to the client (but then again, after cca.8-9 minutes the problem reappears, although not necessarily on the same device server).

We are using Tango 9.2.5, omniorb 4.1.6, zmq 4.2.1  (all Ubuntu apt packages) on Ubuntu 17.10.

To summarize, we can reproduce the issue with broken events if:
- client subscribes to an attribute which is not polled, but rather manually pushed from code
- client subscribes on at least two devices from two separate device server instances (if only one device server instance hosts both devices, we do not observe this problem)

mkomel:
Additional piece of information:

by observing the traffic (with wireshark) between the device servers and the client, we can still see packets going from source to destination for both device servers, even after the client stops registering events from one of them.
I guess this is then an indication that something must be wrong on the client side.

bourtemb:
Thank you very much for the detailed bug report, Matej,
I can reproduce the problem on an Ubuntu 17.10 Virtual machine. This will help us to understand the origin of the bug.

mkomel:
After some more testing we have found out that this is actually not specific to attributes which are manually pushed.
[...]

PR #445, opened 2018-04-10
Set the ZMQ Receive Buffer High Water Mark only if it changes (#444)

bourtemb:
This is to reduce the impact of a bug present in ZMQ 4.2.0 and ZMQ 4.2.1 which led to a bad lwm and hwm calculation when ZMQ_RCVHWM was set after the bind of the socket. See #444 for more details


t-b:
@bourtemb Do you know the upstream commit ID where the bug is fixed?

tango-tickets-migrator:
Matej (@mkomel) was talking about the following commit: https://github.com/zeromq/libzmq/commit/edc770d680b2be5a633526bd2c724d66406c8360 in #444

t-b:
@tango-tickets-migrator Thx.

PR #454, opened 2018-05-17
Fix #453: replace assert with exception

### LINKED ISSUE #453: cppTango 9.3.0 : Crash of C++ client when trying to subscribe to a Tango 7 device
We experienced a crash of one of our HDB++ Event Subscriber when it was trying to subscribe to archive events on an attribute from a Tango 7 device (which is still using notifd).
The crash was due to one of the assert calls which are at the end of Tango::EventConsumer::initialize_received_from_admin() method.
We could discuss whether we still want to support the compatibility with notifd events but I think attempting to subscribe to some events on Tango 7 device attributes should in any case not stop the clients.
If we decide to stop the support for these old devices (to be discussed), we should at least throw an exception notifying that this is no longer supported without crashing the client.

Ingvord:
subj

### DIFF
--- a/cppapi/client/event.cpp
+++ b/cppapi/client/event.cpp
@@ -1837,14 +1837,14 @@ void Tango::EventConsumer::initialize_received_from_admin(const Tango::DevVarLon
                                                           const string &adm_name,
                                                           bool device_from_env_var)
 {
-	long server_tango_lib_ver = dvlsa->lvalue[0];
+ 	long server_tango_lib_ver = (dvlsa->lvalue.length() > 0) ? dvlsa->lvalue[0] : 0;
 
     //event name is used for zmq topics filtering
     //channel name is used for heartbeat events
 	if (server_tango_lib_ver >= 930)
 	{
 		received_from_admin.event_name = (dvlsa->svalue[dvlsa->svalue.length() - 2]);
-        received_from_admin.channel_name = (dvlsa->svalue[dvlsa->svalue.length() - 1]);
+                received_from_admin.channel_name = (dvlsa->svalue[dvlsa->svalue.length() - 1]);
 	}
 	else
 	{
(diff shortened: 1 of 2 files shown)


Ingvord:
Should I update version number? It is already 9.3.1

cppapi/client/event.cpp
```diff
@@ -1860,9 +1860,22 @@ void Tango::EventConsumer::initialize_received_from_admin(const Tango::DevVarLon
 		received_from_admin.channel_name = adm_name_lower;
 	}
 
-	assert(!(received_from_admin.event_name.empty()));
+    if (received_from_admin.event_name.empty())
+    {
+        EventSystemExcept::throw_exception(API_NotSupported,
+                                           "Server does not send event_name. Possibly to old server. Event system is not be initialized!",
+                                           "EventConsumer::initialize_received_from_admin()");
+        throw;
```
bourtemb: I think this "throw;" is not necessary since EventSystemExcept::throw_exception is already throwing an exception.

cppapi/client/event.cpp
```diff
@@ -1860,9 +1860,22 @@ void Tango::EventConsumer::initialize_received_from_admin(const Tango::DevVarLon
...
+    }
+
 	cout4 << "received_from_admin.event_name = " << received_from_admin.event_name << endl;
-	assert(!(received_from_admin.channel_name.empty()));
+    if (received_from_admin.channel_name.empty())
+    {
+        EventSystemExcept::throw_exception(API_NotSupported,
+                                           "Server does not send channel_name. Possibly to old server. Event system is not be initialized!",
```
bourtemb: => Server did not send the channel name. The server is possibly too old. The event system is not initialized"

Ingvord:
> I replaced the first lines of the method with:
long server_tango_lib_ver = 0;
if(dvlsa->lvalue.length() > 0)
{
	server_tango_lib_ver = dvlsa->lvalue[0];
}

Well, I wonder if we should actually throw exception in this case. I guess version number was added in Tango 8, am I right?

Ingvord:
@bourtemb , I have update this PR so it throws an exception in case tango lib version is not provided by the server. I think now your client should not be able to subscribe and hence avoid seg faults. Could you please test it? Thanks

cppapi/client/event.cpp
```diff
@@ -1837,7 +1837,14 @@ void Tango::EventConsumer::initialize_received_from_admin(const Tango::DevVarLon
                                                           const string &adm_name,
                                                           bool device_from_env_var)
 {
- 	long server_tango_lib_ver = (dvlsa->lvalue.length() > 0) ? dvlsa->lvalue[0] : 0;
+ 	if(dvlsa->lvalue.length() == 0)
```
bourtemb: After deeper investigations and tests, it appears that this modification makes it impossible for Tango 9.3.1 device servers to subscribe to Tango <= 8 device servers. This is definitely a too strong limitation.
We need to change that.

Ingvord: @bourtemb , we will need a docker with Tango 7 server for that - it is very hard to fix this blindly, I will be back in office next week and can start to prepare one... Hope this does not delay you too much!

Issue #458, opened 2018-05-25
cppTango 9.3.0 and 9.3.1: Issue with dynamic attributes


bourtemb:
A problem has been observed with cppTango 9.3.0 and 9.3.1 in device servers using dynamic attributes.
This could prevent some device servers invoking add_dynamic_attributes() in the init_device() for instance to start up correctly (invoking add_dynamic_attributes() in the init_device() is usually a bad idea since the device_factory method will erase them immediately after and invoke again add_dynamic_attributes(), but this was an easy way to show the problem. The problem can occur as well when restarting a device on a typically well written device server using dynamic attributes).
A new map to improve performances has been added in #430 but the old vector was kept to maintain the compatibility with the older versions. The index in the vector is now stored as well in the map to improve the performances of get_attr_ind_by_name() method but these indexes are not correctly updated in some cases (typically when an attribute is removed). Moreover, there is a trick in the way the attribute list vector is created and maintained, _state_ and _status_ attributes must always be at the end of the vector. This was forgotten in #430.

bourtemb:
Fixed with PR #459

bourtemb:
This issue is actually still present when a device is restarted on a device server exporting **several** devices which are creating dynamic attributes.
d43fbefd85617a2690f7648e7471c8191fd27a41 fixed only the case where there was only one device with dynamic attributes exported by the device server.
A fix seems to have been found and a PR will be suggested soon.

bourtemb:
Second problem reported on 8 Nov 2018 should be fixed in #508

PR #459, opened 2018-05-25
State and Status attributes are always at the end of attr list vector

bourtemb:
Indexes are now correctly updated taking this into account (#458).
This should fix a bug introduced in #430 which was impacting servers using dynamic attributes.

### DIFF
--- a/cppapi/server/multiattribute.cpp
+++ b/cppapi/server/multiattribute.cpp
@@ -1050,7 +1050,8 @@ void MultiAttribute::remove_attribute(string &attr_name,bool update_idx)
 		for (;pos != attr_list.end();++pos)
 		{
 			(*pos)->set_attr_idx((*pos)->get_attr_idx() - 1);
-			ext->attr_map[(*pos)->get_name_lower()].att_index_in_vector--;
+			string & attr_name_lower = (*pos)->get_name_lower();
+			ext->attr_map[attr_name_lower].att_index_in_vector--;
 		}
 
 		Tango::Util *tg = Tango::Util::instance();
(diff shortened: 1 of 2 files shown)


cppapi/server/multiattribute.cpp
```diff
@@ -1061,7 +1067,10 @@ void MultiAttribute::remove_attribute(string &attr_name,bool update_idx)
 			{
 				int idx = dev_att_list[loop]->get_attr_idx();
 				if (idx > old_idx)
+				{
 					dev_att_list[loop]->set_attr_idx(idx - 1);
+					(*dev_ite)->get_device_attr()->ext->attr_map[dev_att_list[loop]->get_name_lower()].att_index_in_vector--;
```
Ingvord: It is better to avoid such long lines. It is a hell to debug it and maintain. Just split it declaring some local variables:

```cpp
string attr_name_lower = dev_att_list[loop]->get_name_lower();
DeviceAttributeExt* ext = (*dev_ite)->get_device_attr()->ext;
ext->attr_map[attr_name_lower].att_index_in_vector--;

```

cppapi/server/multiattribute.cpp
```diff
@@ -1045,7 +1048,10 @@ void MultiAttribute::remove_attribute(string &attr_name,bool update_idx)
 	if (update_idx == true)
 	{
 		for (;pos != attr_list.end();++pos)
+		{
 			(*pos)->set_attr_idx((*pos)->get_attr_idx() - 1);
+			ext->attr_map[(*pos)->get_name_lower()].att_index_in_vector--;
```
Ingvord: It is better to split this line declaring `string attr_name_lower` variable - better for debugging and maintaining

Issue #457, opened 2018-05-25
cppTango 9.3.0-9.3.1 : Memory leak in DServer::zmq_event_subscription_change(Tango::DevVarStringArray const*)


bourtemb:
A memory leak has been introduced in cppTango 9.3.0 and cppTango 9.3.1 (introduced in #423).

The leak has been introduced here:
Tango::DServer::zmq_event_subscription_change(Tango::DevVarStringArray const*) (eventcmds.cpp:1083)

Here is the detail report by valgrind
blocks are definitely lost in loss record 254 of 258
==9156==    at 0x4C298A0: operator new[](unsigned long) (vg_replace_malloc.c:389)
==9156==    by 0x623DF18: CORBA::string_dup(char const*) (in /mntdirect/_segfs/tango/ORB/omniORB4.2.1_p1/debian8/lib/libomniORB4.so.2.1)
==9156==    by 0x55326C2: Tango::DeviceImpl::adm_name() (device.cpp:1663)
==9156==    by 0x55B8306: Tango::DServer::zmq_event_subscription_change(Tango::DevVarStringArray const*) (eventcmds.cpp:1083)
==9156==    by 0x559D6B0: Tango::ZmqEventSubscriptionChangeCmd::execute(Tango::DeviceImpl*, CORBA::Any const&) (dserverclass.cpp:1371)
==9156==    by 0x5579FD3: Tango::DeviceClass::command_handler(Tango::DeviceImpl*, std::string&, CORBA::Any const&) (deviceclass.cpp:1197)
==9156==    by 0x5531ED6: Tango::DeviceImpl::command_inout(char const*, CORBA::Any const&) (device.cpp:1537)
==9156==    by 0x5550665: Tango::Device_2Impl::command_inout_2(char const*, CORBA::Any const&, Tango::DevSource) (device_2.cpp:438)
==9156==    by 0x556B025: Tango::Device_4Impl::command_inout_4(char const*, CORBA::Any const&, Tango::DevSource, Tango::ClntIdent const&) (device_4.cpp:473)
==9156==    by 0x549EDC1: _0RL_lcfn_6fe2f94a21a10053_a3000000(omniCallDescriptor*, omniServant*) (tangoSK.cpp:5383)
==9156==    by 0x6288A50: omniCallHandle::upcall(omniServant*, omniCallDescriptor&) (in /mntdirect/_segfs/tango/ORB/omniORB4.2.1_p1/debian8/lib/libomniORB4.so.2.1)
[...]

PR #460, opened 2018-05-28
Add Tango::string_free and fix memory leak in zmq_event_subscription_change() (#457)

bourtemb:
Add Tango::string_free() to hide CORBA::string_free()
Fix memory leak in DServer::zmq_event_subscription_change()


cppapi/server/eventcmds.cpp
```diff
@@ -1080,7 +1080,9 @@ DevVarLongStringArray *DServer::zmq_event_subscription_change(const Tango::DevVa
         ret_data->svalue[size] = Tango::string_dup(event_topic.c_str());
 
         string channel_name = ev->get_fqdn_prefix();
-        channel_name += dev->adm_name();
+        char * adm_name = dev->adm_name();
```
Ingvord: Good to know!

PR #469, opened 2018-08-06
Forwarded dynamic attributes

jairomoldes:
This pull request is a continuation of #342 (https://github.com/tango-controls/cppTango/pull/342). Due to the numerous problems described in that solution I propose another one.
If you all agree that this is a better solution I will update tango documentation in another commit, so please do not merge until then.

### DIFF
--- a/cppapi/server/device.cpp
+++ b/cppapi/server/device.cpp
@@ -3400,12 +3400,13 @@ void DeviceImpl::add_attribute(Tango::Attr *new_attr)
 
 //
 // An attribute with the same name is already defined within the class. Check if the data type, data format and
-// write type are the same
+// write type are the same (allow this only if attribute is forwarded)
 //
 
-        if ((attr_list[i]->get_type() != new_attr->get_type()) ||
-            (attr_list[i]->get_format() != new_attr->get_format()) ||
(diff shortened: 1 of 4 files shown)


bourtemb:
Thank you very much @jairomoldes for proposing this PR to improve the previous one.
It looks like the following case is not well handled:

Let's take the TestFwd device server you provided in the PR #342
The user wants to add a dynamic forwarded attribute using the following code:
```
// Add a forwarded dynamic attribute without specifying the root attr (DB value should be used)
Tango::FwdAttr *fwd_attr2 = new Tango::FwdAttr("TestFwdDyn2");
Tango::UserDefaultFwdAttrProp attr_prop2;
attr_prop2.set_label("Test dyn fwd attr2");
fwd_attr2->set_default_properties(attr_prop2);
this->add_attribute(fwd_attr2);
```
But the user did not configure the __root_att TestFwdDyn2 attribute property.
In this case, if you use tg_devtest (test device from jive) on the device, TestFwdDyn2 is listed in the attribute list but with an unknown data format, so atkpanel considers this attribute as invalid and does not list it (except in the diagnostics window to tall us it is invalid).
The state of the device stays on and the status says "The device is in ON state." in this case.
The status should inform the user TestFwdDyn2 attribute property is not correctly defined in this case.

PR #475, opened 2018-08-21
Backport 37abbe7: Doxygen files

Ingvord:
Backport 37abbe7


Ingvord:
@NexeyaSGara , Windows built has failed. Could you please have a look? Thanks.

bourtemb:
The windows builds are failing because of appveyor retention policy...
We are still trying to download the artifacts which were built many months ago.
These artifacts have been deleted by appveyor in the mean time because they were older than 6 months.
@NexeyaSGara has done some work to work around this issue but this work is not yet finished.

bourtemb:
Actually it should not be too difficult to fix.
@NexeyaSGara has done some work to transfer all the dependencies to tango-controls github organization and build some releases we should be able to download instead of the previous artifacts.
So, in theory it is just a matter of updating the links.
Here are links to the release pages of the projects he transferred:
https://github.com/tango-controls/boost-ci/releases --> 1.64
https://github.com/tango-controls/omniorb-windows-ci/releases --> 4.2.1-2
https://github.com/tango-controls/zmq-windows-ci/releases --> 4.0.5
https://github.com/tango-controls/Pthread_WIN32/releases --> 2.91

Issue #428, opened 2018-01-18
Regular test failure in new_tests/cxx_fwd_att.cpp:749


bourtemb:
There is quite often a failure of the Travis CI tests due to an assertion failure (Assertion failed: cb.cb_err == 1) in new_tests/cxx_fwd_att.cpp file at line 749

Here is an example of logs seen on Travis:

> 1: Testing Fwd Att :
> 1: 
> 1: 	error case in forwarded attribute configuration --> OK
> 1: 	reading forwarded attribute --> OK
> 1: 	writing forwarded attribute --> OK
> 1: 	data propagation --> OK
> 1: 	tango://172.17.0.3:10000/test/debian8/10/short_attr_rw==tango://172.17.0.3:10000/test/debian8/10/short_attr_rw
> 1: 	attribute configuration --> OK
> 1: 	attribute configuration propagation --> OK
> 1: 	label is local --> OK
> 1: 	write read on forwarded attribute --> OK
> 1: 	polling on a forwarded attribute --> OK
> 1: Tango exception
> 1: Severity = ERROR 
> 1: Error reason = API_AttributePollingNotStarted
> 1: Desc : The polling (necessary to send events) for the attribute short_attr_rw is not started
> 1: Origin : DServer::event_subscription
> 1: 
> 1: Tango exception
> 1: Severity = ERROR 
> 1: Error reason = API_CommandFailed
> 1: Desc : Failed to execute command_inout on device dserver/devtest/test, command ZmqEventSubscriptionChange
> 1: Origin : Connection::command_inout()
> 1: 
> 1: Tango exception
> 1: Severity = ERROR 
> 1: Error reason = API_DSFailedRegisteringEvent
> 1: Desc : Device server send exception while trying to register event
> 1: Origin : EventConsumer::connect_event()
> 1: 
> 1: Tango exception
> 1: Severity = ERROR 
> 1: Error reason = API_CommandFailed
> 1: Desc : Failed to execute command_inout on device dserver/fwdtest/test, command ZmqEventSubscriptionChange
> 1: Origin : Connection::command_inout()
> 1: 
> 1: Tango exception
> 1: Severity = ERROR 
> 1: Error reason = API_DSFailedRegisteringEvent
> 1: Desc : Device server send exception while trying to register event
> 1: Origin : EventConsumer::connect_event()
> 1: 
[...]

Issue #384, opened 2017-07-25
CXX::cxx_fwd_att (Failed)


Ingvord:
This test fails randomly in Release build.

Ingvord:
See also #428

bourtemb:
This should be fixed with #493

PR #493, opened 2018-10-23
Fix cxx_fwd_att.cpp:749 occasional test failure (#384,#428)

bourtemb:
Added a sleep time after sending AddObjPolling command to the root device and before subscribing event on the forwarded device to give some time for the polling to start up correctly and changed the assertion cb.cb_err == 1 into cb.cb_err == 0.

The test failures we were experiencing before were due to the fact that sometimes (most of the times) an error event was sent with the following exception:
```
Severity = ERROR 
Error reason = API_NoDataYet
Desc : No data available in cache for attribute short_attr_rw
Origin : Device_3Impl::read_attributes_from_cache

Severity = ERROR 
Error reason = API_AttributeFailed
Desc : Failed to read_attribute on device test/debian8/10, attribute short_attr_rw
Origin : DeviceProxy::read_attribute()
```

And sometimes, no error event was sent because the polling had enough time to start up correctly on the root device. So there was already a value available in the cache.
So the assertion (cb.cb_err == 1) was wrong in this last case because cb.cb_err was equal to 0.

### DIFF
--- a/cpp_test_suite/new_tests/cxx_fwd_att.cpp
+++ b/cpp_test_suite/new_tests/cxx_fwd_att.cpp
@@ -714,7 +714,7 @@ public:
 		EventCallBack cb(this);
 		cb.cb_executed = 0;
 		cb.cb_err = 0;
-
+		fwd_device->set_source(CACHE_DEV);
 		try
 		{
 			fwd_device->subscribe_event("fwd_short_rw",Tango::PERIODIC_EVENT,&cb);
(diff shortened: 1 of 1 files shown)


bourtemb:
The error event _API_NoDataYet_ was actually sent because fwd_device DeviceProxy source was set to _CACHE_ by the previous test (test_polling_on_a_forwarded_attribute). So the synchronous call occurring during the event subscription is trying to read from the cache only and gets an exception because there is nothing yet in the cache.
If we add:
`fwd_device->set_source(CACHE_DEV);`
in test_event_on_a_forwarded_attribute() test, we could even remove
`TangoSleep(1);`
after _AddObjPolling_ command.
The last test (test_locking_device_with_forwarded_attribute) should not be affected by that.
This change was implemented in fa27e33d54d5d7ce831cc773a155223fefd61523 commit

PR #503, opened 2018-12-04
PR to fix #359

### LINKED ISSUE #359: Change event subscription blind to change events right after device server restart

Ingvord:
I think we can take a chance to refactor event supplier a little bit here as well...

### DIFF
--- a/cpp_test_suite/event/archive_event.cpp
+++ b/cpp_test_suite/event/archive_event.cpp
@@ -235,7 +235,7 @@ int main(int argc, char **argv)
 #endif
 
 		coutv << "cb excuted = " << cb.cb_executed << endl;
-		assert (cb.cb_executed == 2);
+		assert (cb.cb_executed == 3);
 		assert (cb.val == 31);
 		assert (cb.val_size == 4);
 				
@@ -256,7 +256,7 @@ int main(int argc, char **argv)
 #endif
 			
 		coutv << "cb excuted = " << cb.cb_executed << endl;
-		assert (cb.cb_executed == 3);
+		assert (cb.cb_executed == 4);
 		assert (cb.val == 30);
 		assert (cb.val_size == 4);
 				
@@ -271,7 +271,7 @@ int main(int argc, char **argv)
 		Tango_sleep(2);
 			
 		coutv << "cb excuted = " << cb.cb_executed << endl;
-		assert (cb.cb_executed == 4);
+		assert (cb.cb_executed == 5);
 		assert (cb.val == 30);
 		assert (cb.val_size == 5);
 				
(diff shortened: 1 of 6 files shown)


bourtemb:
> I think we can take a chance to refactor event supplier a little bit here as well...

I think this fix (if it really fixes the pb) is quite simple so we should keep it simple.
If we decide to do some refactoring, I would do that in another PR. 
So we can revert the refactoring PR or this one easily if needed in case of troubles.

This change introduces a change in the behaviour of the events (an additional event is sent the first time the polling is executed), so this has an impact on the tests a we can see (it was expected).
I would focus on the tests for the moment and update them carefully.

Ingvord:
@bourtemb , ok!

Issue #509, opened 2019-01-07
Tango9.3.2 cpp read_attribute no more DevFail on ATTR_INVALID


alexElettra:
Same server e same client compiled for Tango9,
a client reading its attribute being ATTR_INVALID:

1) when using tango-9.2.5a/lib/libtango.so.9 the client is notified with a DevFail:
       Tango exception
        Severity = ERROR 
        Error reason = API_EmptyDeviceAttribute
        Desc : cannot extract, no data in DeviceAttribute object 
        Origin : DeviceAttribute::is_empty

**2) when using tango-9.3.2/lib/libtango.so.9 the client IS NO MORE notified with DevFail:
     --------------------------------
     No data in DeviceAttribute object
     --------------------------------**

Is this the new behavior? or a bug? or I missed the relevant line in the docs?

Thanks,
Alessandro

bourtemb:
Hi @alexElettra 

I think this is a side effect of https://github.com/tango-controls/cppTango/pull/393 which was done to solve https://github.com/tango-controls/cppTango/issues/392.
Were you getting this exception when invoking DeviceAttribute::get_type() method?

alexElettra:
Hi @bourtemb 

a very/very short snippet of my code:

try {
    Tango::DeviceAttribute attr = deviceproxy->read_attribute(...);
    attr_type=attr.get_type();
    ....
    switch (attr_type){
      .... since attribute may be of different type, do proper way....
    }
} catch (Tango::DevFailed &e) {
 .... here I fall if using Tango 9.2.5a ....
} catch(...) {
}

..... while in Tango 9.3.2 read_attribute of a ATTR_INVALID do not send a DevFail.

So I see now that wasn't the read_attribute() but attr.get_type() leading me to DevFail!
In the docs: read_attribute(): throws ConnectionFailed, CommunicationFailed... never DevFail! so I was driven into a mistake by the change in get_type()!

Perhaps this change may be underlined in RELEASE_NOTES doc.

bourtemb:
> Perhaps this change may be underlined in RELEASE_NOTES doc.

I agree.

bourtemb:
This DeviceAttribute::get_type() behaviour change is now mentioned in the changelog and release notes and the potential consequences are now listed there.
Solved in #519

PR #519, opened 2019-01-24
Add CHANGELOG.md

bourtemb:
based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
Add more details about changes occurring since Tango 9.2.5 and potential impact on user experience as suggested by #509


bourtemb:
RELEASE_NOTES file is kept for the moment but should probably be replaced with something less technical and giving a better summary of the main changes, asking users to refer to the CHANGELOG.md file if they want more technical details.
I think release notes should also focus on the changes between stable releases.
The user does not care if a bug was introduced in 9.2.9 and fixed in 9.3.3 if this bug was not present in 9.2.5.
We could actually create a wiki page for the Release Notes when there will be a new stable release to have them in a more friendly format where we could eventually insert snapshots and images.

bourtemb:
I propose to go ahead and to do a cppTango 9.3.3 release, flagged as pre-release on Github.
Before to tag this new version, I'd like to merge this Pull Request listing all the changes.
The exact date of the release will have to be updated on CHANGELOG.md and RELEASE_NOTES.md when we will create the release.

Ingvord:
> We could actually create a wiki page for the Release Notes when there will be a new stable release to have them in a more friendly format where we could eventually insert snapshots and images.

IMHO we should use GitHub releases for RELEASE_NOTES. Seems more logical for me than a dedicated wiki page. When you go to releases it lists all the releases with some details as in [Waltz](https://github.com/tango-controls/waltz/releases) for instance.

bourtemb:
> IMHO we should use GitHub releases for RELEASE_NOTES. Seems more logical for me than a dedicated wiki page. When you go to releases it lists all the releases with some details as in [Waltz](https://github.com/tango-controls/waltz/releases) for instance.

The release notes should be on the Github release page I agree with you. And your example with Waltz shows that there is no problem to include images if needed on this Github release page.

bourtemb:
@ingvord, thanks for your review.
I added 5 small commits since you last reviewed this PR.
Could you please review again this PR and if you agree, could you please merge it?
I'd like to tag the release tomorrow (Friday 29).
As I already wrote about the RELEASE_NOTES file, IMHO it is good to keep it as a file provided with the source code too. And it's easier to proceed like that when preparing the release so we can prepare and discuss the contents of the release notes in a Pull Request.
I agree with you when you write "it should be on the Github release page".
I simply propose to copy the content of this file in the Github release page when the release is created.

PR #520, opened 2019-01-26
Ignore ENOENT during zmq_disconnect call (fixes #514)

### LINKED ISSUE #514: Uncaught DevFailed exception in DeviceProxy destructor in cxx_stateless_subscription testcase
Hello,
I got following error every time I run stateless_subscription testcase:
```
38: terminate called after throwing an instance of 'Tango::DevFailed'
1/1 Test #38: CXX::cxx_stateless_subscription ...Child aborted***Exception:  55.67 sec
```

I'm using:
* cppTango tango-9-lts, 07eea09441aba365e17b077347797e1f8ab04c91
* gcc 8.2.1 20181127
* libzmq built from tag v4.3.0

`DevFailed` is being thrown in `device2` destructor:
https://github.com/tango-controls/cppTango/blob/07eea09441aba365e17b077347797e1f8ab04c91/cpp_test_suite/new_tests/cxx_stateless_subscription.cpp#L88-L95

The root cause of the exception is `disconnect` failure in heartbeat socket:

https://github.com/tango-controls/cppTango/blob/07eea09441aba365e17b077347797e1f8ab04c91/cppapi/client/zmqeventconsumer.cpp#L836

errno is being set to ENOENT. This means that given endpoint was already disconnected. Indeed SUB socket is disconnected when PUB endpoint is terminated (`kill_server` call).

The workaround is to delete `device2` proxy before `kill_server` is called.

A proper fix would be to handle disconnection failure (at least the one due to ENOENT) in zmqeventconsumer.cpp. I can submit a patch if you're ok with such a solution.

Probably there are other scenarios where exception can be thrown in a destructor. Such exception will always result in a call to `terminate` (as destructor is implicitly `noexcept`).

mliszcz:
A change in ZMQ 4.2.3 breaks one of Tango tests (`cxx_stateless_subscription`, see #514 for details). 

Since zeromq/libzmq@edb4ca1023ff9001b545e827b227402b76a2b55d, peer disconnection triggers `zmq_disconnect` on a zmq socket. Further `zmq_disconnect` calls for such socket and endpoint will fail with errno set to `ENOENT`. This patch silently ignores such failure.

I'm not adding any new testcases. I don't think it is possible to have a reliable reproduction with ZMQ older than v4.2.3.

### DIFF
--- a/cppapi/client/eventconsumer.h
+++ b/cppapi/client/eventconsumer.h
@@ -666,6 +666,7 @@ private :
 	void set_ctrl_sock_bound() {sock_bound_mutex.lock();ctrl_socket_bound=true;sock_bound_mutex.unlock();}
 	bool is_ctrl_sock_bound() {bool _b;sock_bound_mutex.lock();_b=ctrl_socket_bound;sock_bound_mutex.unlock();return _b;}
 	void set_socket_hwm(int hwm);
+	static void disconnect_socket(zmq::socket_t&, const char*);
 
     bool check_zmq_endpoint(const string &);
 
--- a/cppapi/client/zmqeventconsumer.cpp
+++ b/cppapi/client/zmqeventconsumer.cpp
@@ -833,7 +833,7 @@ bool ZmqEventConsumer::process_ctrl(zmq::message_t &received_ctrl,zmq::pollitem_
 			if (pos != connected_heartbeat.end())
 				connected_heartbeat.erase(pos);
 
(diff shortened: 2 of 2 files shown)


bourtemb:
> I'm not adding any new testcases. I don't think it is possible to have a reliable reproduction with ZMQ older than v4.2.3.

cxx_stateless_subscription seems to be a good test to highlight this problem with ZMQ >= 4.2.3
For older ZMQ versions, the problem was not there so no test is needed.

PR #521, opened 2019-01-26
Prevent exceptions from unsubscribe_event leaving DeviceProxy destructor

mliszcz:
Per @bourtemb suggestion from #514 let's handle `unsubscribe_event` failures in `~DeviceProxy` to prevent exceptions being thrown out of a destructor.

### DIFF
--- a/cppapi/client/devapi_base.cpp
+++ b/cppapi/client/devapi_base.cpp
@@ -2639,26 +2639,7 @@ DeviceProxy::~DeviceProxy()
 //
 // If the device has some subscribed event, unsubscribe them
 //
-
-    if (ApiUtil::_is_instance_null() == false)
-    {
-        ApiUtil *api_ptr = ApiUtil::instance();
-        ZmqEventConsumer *zmq = api_ptr->get_zmq_event_consumer();
-        if (zmq != Tango_nullptr)
-        {
-            vector<int> ids;
-            zmq->get_subscribed_event_ids(this, ids);
-            if (ids.empty() == false)
-            {
-                vector<int>::iterator ite;
-                for (ite = ids.begin(); ite != ids.end(); ++ite)
-                {
-                    unsubscribe_event(*ite);
-                }
-
-            }
-        }
-    }
+    unsubscribe_all_events();
 
 //
 // If the device is locked, unlock it whatever the lock counter is
@@ -2688,6 +2669,38 @@ DeviceProxy::~DeviceProxy()
 #endif
 }
 
+void DeviceProxy::unsubscribe_all_events()
+{
+    if (ApiUtil* api = ApiUtil::instance())
+    {
+        if (ZmqEventConsumer* zmq_event_consumer = api->get_zmq_event_consumer())
+        {
+            vector<int> event_ids;
+            zmq_event_consumer->get_subscribed_event_ids(this, event_ids);
+
+            for (vector<int>::iterator event_id = event_ids.begin();
+                event_id != event_ids.end();
+                ++event_id)
+            {
+                try
+                {
+                    unsubscribe_event(*event_id);
+                }
+                catch (CORBA::Exception& e)
+                {
+                    Tango::Except::print_exception(e);
+                }
+                catch (...)
+                {
+                    std::cerr <<
+                        "DeviceProxy::unsubscribe_all_events(): "
+                        "Unknown exception thrown from unsubscribe_event() for "
+                        "device \"" << name() << "\" and event_id=" << *event_id << std::endl;
+                }
+            }
+        }
+    }
+}
 
 //-----------------------------------------------------------------------------
 //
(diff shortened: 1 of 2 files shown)


cppapi/client/devapi_base.cpp
```diff
@@ -2688,6 +2669,38 @@ DeviceProxy::~DeviceProxy()
...
+                    }
+                    catch (CORBA::Exception& e)
+                    {
+                        Tango::Except::print_exception(e);
+                    }
+                    catch (...)
+                    {
+                    }
+                }
+
```
bourtemb: The only improvement I could think of is to take the opportunity of this Pull Request to remove this empty line.
@mliszcz, could you do that, please?

t-b: @mliszcz @bourtemb How about getting also rid of the useless `if (ids.empty() == false)`? And I would also include a message output in `catch(...)` as reaching that branch seems fishy to me.

mliszcz: Thanks for reviewing the PR!

@bourtemb I've removed the redundant line.

@t-b I've removed the `empty()` check. I've also did a few cosmetic changes and renames. I don't know what is the preferred coding style, so please check those as well.

As for the error message in `catch(...)`, I've added a printout. Is there a logger for client API that can print at least the file and line number?
BTW In my opinion a library should not print anything to std out/err, but in case of d-tor failure there is no other way to notify the user.

t-b: @mliszcz Looks better! Regarding the printout I would print to `cerr` as that is what `Except::print_exception` does as well, IIRC this is a log stream and not the same as std::cerr.

mliszcz: @t-b thanks for checking! I've already looked into `print_exception` and my assumption was that `cerr` is `std::cerr`. `cerr` would be very unfortunate name for a custom logger stream, especially that `tango.h` includes `iostream` and brings `std` into default scope.

Following grep does not yield any meaningful result: `grep -r cerr | grep -v 'cerr <<'`
Seems that `cerr` actually IS `std::cerr`. Would it be ok for you if I'll leave the namespace in the name as it is now?

bourtemb: @t-b, thanks for the review and the good catch for the empty() if.

**cout** (not cerr) is redirected to the log stream, except if the library is compiled in Debug mode on Linux (_TANGO_LIB compilation flag is not set in this case). In Debug mode on Linux, cout is the standard std::cout.

There is the following code in cppapi/server/logging.h:

```
#ifdef _TANGO_LIB //== compiling TANGO lib ====================

//-- Overwrite std::cout  -------------------------------------
#define cout \
    if (API_LOGGER) \
      API_LOGGER->get_stream(log4tango::Level::INFO, false) \
        << log4tango::LogInitiator::_begin_log

#endif //== compiling TANGO lib ===============================
```

It looks like _TANGO_LIB is always defined when compiling the Tango library on Windows... So it seems cout is always redirected to the log stream on Windows, even when the library is compiled in Debug mode.

cerr is actually always std:cerr with the current source code.

> As for the error message in `catch(...)`, I've added a printout. Is there a logger for client API that can print at least the file and line number?

Not yet.

> Would it be ok for you if I'll leave the namespace in the name as it is now?

With the current code, I don't see any problem with that since cerr is std::cerr.
The question is: do we want to use a trick at some point to overload cerr with something else in a similar way as it was done for cout?
I would say if it's the case in the future, we can always modify this part of the code if needed...
We could also use cout3 to log a message to the logging system.

cppapi/client/devapi_base.cpp
```diff
@@ -2688,6 +2669,37 @@ DeviceProxy::~DeviceProxy()
...
+                }
+                catch (CORBA::Exception& e)
+                {
+                    Tango::Except::print_exception(e);
+                }
+                catch (...)
+                {
+                    std::cerr
+                        << "Unknown exception thrown from unsubscribe_event() event_id="
+                        << *event_id << std::endl;
```
bourtemb: I think we should also add the name of the device in the error message as well as the current method name.
So I would suggest to change the error message with something like:
```
std::cerr << " DeviceProxy::unsubscribe_all_events(): Unknown exception thrown from unsubscribe_event() for device \""
             << name() << "\" and event_id=" << *event_id << " std::endl;
```

mliszcz: Thanks! I've updated the code according to your suggestion.

cppapi/client/devapi_base.cpp
```diff
@@ -2688,6 +2669,38 @@ DeviceProxy::~DeviceProxy()
...
+                }
+                catch (CORBA::Exception& e)
+                {
+                    Tango::Except::print_exception(e);
+                }
+                catch (...)
+                {
+                    std::cerr <<
+                        "DeviceProxy::unsubscribe_all_events(): "
+                        "Unknown exception thrown from unsubscribe_event() for"
```
bourtemb: a space is missing at the end between "for" and "device"

mliszcz: Thanks! Fixed.

PR #523, opened 2019-01-28
Use python2 (if available) to run cxxtestgen

mliszcz:
As python 2.x is reaching EOL soon (PEP 373) and some distros already switched to use python 3.x as a default python interpreter, I propose to follow PEP 394 and use `python2` command/link instead of `python` when invoking cxxtestgen.

If `python2` is missing, CMake will fall-back to using `python` (if it's on `$PATH`).

### DIFF
--- a/cpp_test_suite/cxxtest/CMakeLists.txt
+++ b/cpp_test_suite/cxxtest/CMakeLists.txt
@@ -1,8 +1,10 @@
 find_package (Threads REQUIRED)
+find_program(PYTHON_INTERPRETER NAMES python2 python)
 
 macro(CXX_GENERATE_TEST name)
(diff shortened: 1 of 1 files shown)


bourtemb:
Did you try to replace cxxtestgen.py with a more recent version from https://github.com/CxxTest/cxxtest github repository?
Is there a version which would be compatible with python2 & python3?
Note that there might be more than 1 file to update if we want to use a more recent version of cxxtest... I didn't investigate.
Your solution has the advantage to be simple and can buy us some time before upgrading cxxtest.

t-b:
I think the PR is first step in the right direction i.e. a pure enhancement.

PR #526, opened 2019-02-02
Add support for precompiling server/tango.h (build time reduction)

mliszcz:
`server/tango.h` includes quite a lot of other headers and is used in most of the files under `cppapi` and `cpp_test_suite` directories. Compiling this header again and again in every translation unit consumes more than 50% of build time.

This PR allows to precompile `server/tango.h` and include precompiled binary instead of plain header.

Below are some numbers from the `time` command (zsh built-in). Before each `make` invocation I've deleted the `build` directory and ran `cmake` again.

```
cmake .. -DCMAKE_INSTALL_PREFIX=${PREFIX}/cppTango -DIDL_BASE=${PREFIX}/tangoidl -DOMNI_BASE=${PREFIX}/omniORB -DZMQ_BASE=${PREFIX}/libzmq -DCMAKE_BUILD_TYPE=Debug -DUSE_PCH=ON
make -j 4  787,72s user 85,49s system 367% cpu 3:57,45 total
make -j 4  779,58s user 85,73s system 364% cpu 3:57,53 total
make -j 4  780,02s user 86,20s system 366% cpu 3:56,32 total

cmake .. -DCMAKE_INSTALL_PREFIX=${PREFIX}/cppTango -DIDL_BASE=${PREFIX}/tangoidl -DOMNI_BASE=${PREFIX}/omniORB -DZMQ_BASE=${PREFIX}/libzmq -DCMAKE_BUILD_TYPE=Debug -DUSE_PCH=OFF
make -j 4  1859,02s user 113,27s system 372% cpu 8:49,15 total
make -j 4  1874,94s user 114,89s system 383% cpu 8:38,51 total
make -j 4  1881,58s user 114,88s system 382% cpu 8:42,14 total
```

Some remarks:
* `tango.h` must be precompiled twice (once for libtango and once for tests). This is due to `-fPIC`/`-pie` mismatch,
* due to the above constraint, PCHs are located in `build/cppapi` and `build/cpp_test_suite`,
* PCH must be passed to gcc via `-include` option
* absolute path is used with `-include` to prevent adding new directory via `-I(...)` - this means that `tango.h` needs to be copied to the directory containing PCH, as gcc needs to see both,
* I've conditionally added some defines (`VALGRIND` `_PTHREADS`, `_REENTRANT`) to `conf_devtest` , `compare_test_object`, `DevTest` and `FwdTest`. They should do no harm and will allow gcc to use PCH for these targets as well (this reduced the build time by 1min).


mliszcz:
I've enabled the PCH in one job on Travis but it did not work.
GCC is rejecting the PCH because object files have some extra defines:
`-D__x86_64__ -D__linux__ -D__OSVERSION__=2`
I don't know yet where these defines come from (gcc, cmake or travis). Once I have this sorted out, I'll update the PR.

```
cd /home/tango/src/build/cpp_test_suite/event && /usr/bin/c++   -DVALGRIND -D_PTHREADS -D_REENTRANT -I/home/tango/include -I/home/tango/src/cppapi/client -I/home/tango/src/cppapi/client/helpers -I/home/tango/src/cppapi/server -I/home/tango/src/cppapi/server/idl -I/home/tango/src/log4tango/include -I/home/tango/src/build/cppapi/server -I/home/tango/src/build/log4tango/include  -std=c++11 -g   -O0 -Wall -Wextra -pedantic -Winvalid-pch -include /home/tango/src/build/cpp_test_suite/tango.h -D__x86_64__ -D__linux__ -D__OSVERSION__=2 -o CMakeFiles/scan.dir/scan.cpp.o -c /home/tango/src/cpp_test_suite/event/scan.cpp
cc1plus: warning: /home/tango/src/build/cpp_test_suite/tango.h.gch: not used because `__OSVERSION__' is defined [-Winvalid-pch]
```

mliszcz:
CI is fixed now. The mismatch was due to some extra flags set by omniORB and added to libtango's PUBLIC flags.

Some results:
Job 1383.1: 26 min 50 sec - PCH
Job 1381.1: 27 min 20 sec - PCH
Job 1380.1: 30 min 41 sec - PCH only for libtango
Job 1379.1: 34 min 8 sec - no PCH
Job 1378.1: 33 min 56 sec - no PCH

t-b:
This is definitly a very nice feature! Improving the build time is desparately needed!

I would suggest, some minor code changes though:
- Check if USE_PCH is ON on non-supported platforms (Windows, etc.) and then warn the user and set it OFF.
- Move the `if(USE_PCH)` check out of the pch functions and do the check at the callsite. This makes the code easier to understand.
- Add constants for ` -DVALGRIND -D_PTHREADS -D_REENTRANT` to avoid having to duplicate it.

bourtemb:
I love the idea of this PR!
It would be great to use these precompiled headers on Windows too! This would reduce the compilation time on appveyor.
Once we are confident that it works fine, I think we should even activate this option by default.

I agree with @t-b's code change suggestions.
@mliszcz , could you please implement the proposed changes in use-pch branch from your fork?

mliszcz:
Hi,
@t-b thanks a lot for your comments!
@bourtemb sure, I'll try to do the rework over the weekend.

bourtemb:
It seems to build on Windows now! :-)
appveyor build succeeded and was much faster than usual builds!
Thank you very much for this work!

mliszcz:
Yes, seems to work. Today I'll do some cleanup, squash the commits and write a short summary of windows-specific changes.

mliszcz:
I did the changes. @bourtemb , @t-b could you please review.

Here are promised comments regarding MSVC and PCHs:
* GCC solution (dedicated target, that builds the pch) cannot be used - CL refuses to use such header due to .pdb mismatch between pch target and client/server_objects target. CL fails with:
`error C2859: xxx.pdb is not the pdb file that was used when this precompiled header was created, recreate the precompiled header.`
There seems to be a [workaround](https://devblogs.microsoft.com/cppblog/shared-pch-usage-sample-in-visual-studio/) that involves copying the pdb but I have not tried this.
* Unlike make, MSVC generator provides supports for precompiled headers. If one of the project (target) sources is compiled with `/Yc` and others are compiled with `/Yu`, the `/Yc` one is compiled first and is expected to produce the pch. `/Yu` sources are compiled later and can use the pch produced by `/Yu`
* Such pch cannot be shared between projects (different .pdb), so we have separate pchs for e.g.  `server_objects_sta` and `client_objects_sta`
* In each project I'm adding a dummy source (`pch_${target}.cpp`) that has no code and is only used for pch creation.
* Object file produced by a source compiled with `/Yc` exports a symbol `__@_PchSym_@00@<compiler-specific>@<generated-name>`. If we would try to link `server_objects_dyn` and `client_objects_dyn` into `tangod` we would get:
  ```
  warning LNK4022: cannot find unique match for symbol '__'
  warning LNK4002: __@@_PchSym_@00@UkilqvxghUxkkgzmtlUxkkzkrUhvieviUhvieviPlyqvxghPwbmOwriUwvyftUgzmtlPhvieviPlyqvxghPwbmOlyq@4B2008FD98C1DD4 defined in ...
  warning LNK4002: __@@_PchSym_@00@UkilqvxghUxkkgzmtlUxkkzkrUxorvmgUxorvmgPlyqvxghPwbmOwriUwvyftUgzmtlPxorvmgPlyqvxghPwbmOlyq@4B2008FD98C1DD4 defined in ...
  error LNK2001: unresolved external symbol __
  ```
* One possible workaround would be to exclude the dummy object (used to produce the pch) from linking (remove one of the objects from `$<TARGET_OBJECTS:client_objects_dyn>`), but cmake does not support filtering in generator expressions
* Another workaround is to pass `/Yl-` switch to CL.exe. According to the documentation, it should prevent CL.exe from generating `__@@_PchSym_` symbol (which is only for debug purposes, I assume), however `/Yl-` option is ignored if the header contains inline code marked with `__declspec(dllexport)` (`TANGO_IMP_EXP` in our case) like this one:
  ```
      TANGO_IMP_EXP static inline bool _is_instance_null()
      {return _instance == NULL;} 
  ```
   ```
  warning C4728: /Yl- option ignored because PCH refe rence is required
  ```
  I don't know if this is expected behavior or compiler issue. I've opened a [ticket](https://developercommunity.visualstudio.com/content/problem/497354/yl-option-is-ignored-c4728-if-declspecdllexport-is.html) to get clear on this topic, but there's no response so far.
* To sum up:
  * there is one precompiled header for `server_objects_sta`
  * there is one precompiled header for `client_objects_sta`
  * there is one precompiled header for `server_objects_dyn`
  * there is NO precompiled header for `client_objects_dyn` (due to `_PchSym_` issue)
  * there are NO precompiled headers for tests (as we cannot easily share pch between targets)
* **As a result, total build time in appveyor went down from 4h30m-6h30m to 2h30m-3h30m** (the numbers are varying a lot, they probably depend on available appveyor resources / overall load)

mliszcz:
As discussed with @bourtemb, we would like to have this PR in next release.

@bourtemb, @Ingvord, @t-b, @NexeyaSGara I have added you as reviewers for this PR. Please help to do the review if you have time.

@t-b, @NexeyaSGara I was told that you are using Windows for kernel development. Could you check if everything works on that platform?

t-b:
Functionality:
Using `git clean -fdx . && cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_PCH=ON|OFF .. && time make` I got

without PCH:
```
real    2m16,868s
user    24m32,824s
sys     1m15,672s
```

with PCH:
```
real    1m8,975s
user    9m55,259s
sys     0m54,738s
```

which is indeed impressive.

I've tested both and it works here on debian jessie.
I've also redone the tests with cmake 2.8.12.2 (compiled from source) to check that there are no incompatibilties introduced.

Codewise:
I've only coarsely glanced over the cmake logic. Looks good.

Commitwise:
I would prefer (this should be discussed for the project I guess) to have separte commits for whitespace only changes, as that makes review much easier.

Documentationwise:
I've added some documentation to INSTALL.md:
```
diff --git a/INSTALL.md b/INSTALL.md
index 72f7e92f..ad551ad8 100644
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -27,6 +27,7 @@
 - `-DCMAKE_BUILD_TYPE=<Release|Debug>`
 - `-DCMAKE_VERBOSE_MAKEFILE=true`
 - `-DTANGO_USE_USING_NAMESPACE=<ON|OFF>` choose `OFF` for modern builds
+- `-DUSE_PCH=<ON|OFF>`
```

and force pushed.

As appveyor passes I've not tested that on windows myself.

bourtemb:
I really like this PR and I am actually in favor of setting USE_PCH to ON as default value in the CMakeLists.txt files and to enable it for all the travis builds too.

t-b:
@bourtemb We could do that in a separate post 9.3.3 release PR as well?

bourtemb:
> @bourtemb We could do that in a separate post 9.3.3 release PR as well?

Sure!

PR #527, opened 2019-02-04
Support newer visual studio version

t-b:


t-b:
@NexeyaSGara Any opinions on that PR?

t-b:
@bourtemb This is ready as well I guess.

bourtemb:
Hi @t-b, thanks for the Pull Request and reminding us about this one.
Maybe I am wrong, but it looks like, this Pull Request does not bring much for the moment, besides the definition of WIN32_VC16 when using MSVC++>=14.2.
It looks like we can compile with Visual Studio 2017 and Visual Studio 2019 without this Pull Request?
Appveyor is already doing it for Visual Studio 2017.
Is this definition really necessary? It might be, sorry for my ignorance, I don't use tango on Windows usually.

And if we do that, why not defining WIN32_VC15 for Visual Studio 2017 (toolset v141), then?

What is not clear to me is what should we put as xx in WIN32_VCxx.
Is it the MS VC++ compiler version, or the Visual studio version?
On this [CMake documentation page](https://cmake.org/cmake/help/v3.14/variable/MSVC_VERSION.html), they are using something named VSxx, which corresponds to the Visual Studio version and it looks like this is what you used to come to WIN32-VC16.
But in our case, we are using something named WIN32_V**C**xx, not WIN32_V**S**xx, so do we refer to the version of the compiler, or the visual studio version? The name suggests that we refer more to the compiler version... But this WIN32_VCxx definition was created years ago when there was a direct match between the Visual studio version and the VC++ compiler version (this changed with Visual Studio 2017) as we can see on [this link](https://dev.to/yumetodo/list-of-mscver-and-mscfullver-8nd)...

t-b:
We are using the compiler versions to determine which C++11 features are enabled or not.

I just followed the existing approach and added support for VS 2019.

Stepping back a second reveals that this kind of compiler feature determination is totally the wrong approach. One should check if a feature works and then enable it. But I think fixing that is out of the scope of this PR.

The cmake link you sent is correct.

The implemented logic still thinks that  `compiler version` == `visual studio version`.

bourtemb:
> We are using the compiler versions to determine which C++11 features are enabled or not.

Indeed, and it looks like it is the only location where we are using WIN32_VCxx definition... Unless our Windows users are relying on that... ?
The C++11 supported features we are using did not change since _MSC_VER >= 1900 (Visual Studio 15) so this PR will not change the behaviour of the code when _MSC_VER >= 1920, except the definition of WIN32_VC16 instead of WIN32_VC14. The question is: Do we need this definition? It looks like we don't need it.
 
> I just followed the existing approach and added support for VS 2019.

I fully understand your approach, which was quite logical, based on the current source code.
 
> Stepping back a second reveals that this kind of compiler feature determination is totally the wrong approach. One should check if a feature works and then enable it. But I think fixing that is out of the scope of this PR.

I agree with you, this would be a better approach.

> The cmake link you sent is correct.
> 
> The implemented logic still thinks that `compiler version` == `visual studio version`.

Which is a bit confusing with the recent Visual Studio versions...

I would suggest to close this PR without merging it and to work on the other approach you suggested on another PR.

t-b:
@bourtemb Okay I'm convinced.

PR #528, opened 2019-02-04
Make `using namespace std;` optional

### LINKED ISSUE #206: Don't import namespaces in header files
I'm using tango 8.1.2 with patches.

In the main include file tango.h, and some others, at the very bottom I'm seeing

`````` ~
# using namespace std;
```~

and this is really ugly. Including namespace in header files is bad practice and forces the using declaration on the the includer of this file.

Therefore I propose to remove these using declarations. If you rely on its presence for backward compatibility at least allow users to not use it by something like

```~
#ifndef TANGO_DONT_USE_STD_NAMESPACE
    using namespace std;
#endif
```~

Reported by: thomas_braun

Original Ticket: [tango-cs/bugs/706](https://sourceforge.net/p/tango-cs/bugs/706)
``````

### PR DESCRIPTION
**Pull request by t-b**
Closes #206.

### CODE REVIEW

**Comment by t-b** [comment]
The addition of the std:: namespace specifier was done via the following
procedure:

- Add cmake config to build without `using namespace std;`
- Replace sure matches with

```
sed -i 's/endl;$/std::endl;/g' *.cpp *.h
sed -i 's/ends;$/std::ends;/g' *.cpp *.h
```

- Start compiling using `make -j1 2>&1 | head -n 100`
-  Open the first failing file in vim and search and replace all occurences with the following ex command. Requires to ACK/NACK each change.

```ex
.,$s/\(\([, \t(<{]\+\)\|^\)\(string\|vector\|transform\|ostringstream\|ios\|ends\|tolower\|toupper\|bad_alloc\|map\|type_info\|unique_ptr\|cerr\|endl\|list\|bitset\|is_same\|is_enum\|underlying_type\|binary_function\|multimap\|istringstream\|boolalpha\|thread\|stringstream\|ostream\|setfill\|setw\|move\|bind2nd\|ptr_fun\|pair\|setprecision\|dec\|hex\|out_of_range\)\>/\1std::\3/gcI
```
- Commit each cpp file separetely
- Repeat

t-b:
@bourtemb Ready to go as travis passes and the new config passes as well.

andygotz:
@grm84 I am sure you are happy to see this PR by @t-b ! Can you review the PR?

bourtemb:
@t-b, appveyor builds are failing for this PR with errors related to std::tolower.
Could you please have a look?

t-b:
@bourtemb Thx, will do.

t-b:
@bourtemb @andygotz Compiles now on windows as well :)

t-b:
@mliszcz @Ingvord @bourtemb Assuming apveyor passes, any objections of merging that soonish? It is a bit of a pain to keep a 12k line diff PR conflict-free.

mliszcz:
I have not checked the whole diff (for obvious reason) but for me it's also ok to merge ASAP.

bourtemb:
> I have not checked the whole diff (for obvious reason) but for me it's also ok to merge ASAP.

Fine with me too, once appveyor passes! 
I didn't check the whole diff neither ;-)
It was actually already foreseen that this PR would be one of the first to be merged because it would be a nightmare to keep it conflict free.

Issue #498, opened 2018-11-15
Add server_init_hook() method


bourtemb:
Following discussion from https://github.com/tango-controls/TangoTickets/issues/7 and offline discussions, we came to the conclusion it would be useful to add a new server_init_hook() method (in the same spirit as always_executed_hook() method) at device level which would be called once the device server admin device is exported.
This would allow for instance for the different devices to subscribe to events at server startup on attributes from other devices of the same device server with stateless parameter set to false. The subscription should work in this case since the admin device would be exported when the hook would be called.

t-b:
Closed via #551.

PR #551, opened 2019-05-13
 Add server_init_hook() method (#498)

Ingvord:
#498

### DIFF
--- a/cpp_test_suite/cpp_test_ds/DevTest.cpp
+++ b/cpp_test_suite/cpp_test_ds/DevTest.cpp
@@ -217,6 +217,11 @@ void DevTest::always_executed_hook()
 
 }
 
+void DevTest::server_init_hook()
+{
+    cout2 << "In server_init_hook method" << std::endl;
+}
+
 //+----------------------------------------------------------------------------
 //
 // method : 		DevTest::state_cmd()
--- a/cppapi/server/device.h
+++ b/cppapi/server/device.h
@@ -452,6 +452,18 @@ public:
 	virtual void always_executed_hook(void) {};
 
 /**
+ * Hook method.
+ *
+ * Default method to implement an action necessary on a device after server has been initialized.
+ *
+ * @exception DevFailed This method does not throw exception but a
+ * redefined method can.
+ * Click <a href="https://tango-controls.readthedocs.io/en/latest/development/advanced/IDL.html#exceptions">here</a> to read
+ * <b>DevFailed</b> exception specification
+ */
+	virtual void server_init_hook() {};
+
+/**
  * Read the hardware to return attribute value(s).
  *
  * Default method to implement an action necessary on a device to read the
--- a/cppapi/server/dserver.cpp
+++ b/cppapi/server/dserver.cpp
@@ -100,10 +100,6 @@ DServer::DServer(DeviceClass *cl_ptr,const char *n,const char *d,Tango::DevState
 
 	polling_th_pool_size = DEFAULT_POLLING_THREADS_POOL_SIZE;
 	optimize_pool_usage = true;
-
-	from_constructor = true;
-	init_device();
-	from_constructor = false;
 }
 
 bool less_than (Command *a,Command *b)
--- a/cppapi/server/dserver.cpp
+++ b/cppapi/server/dserver.cpp
@@ -480,6 +480,61 @@ void DServer::init_device()
 	}
 }
 
+void DServer::server_init_hook()
+{
+#ifdef HAS_RANGE_BASE_FOR
+    for (DeviceClass *dclass : this->get_class_list())
+    {
+        for (DeviceImpl *device : dclass->get_device_list())
+        {
+            cout4 << "Device " << device->get_name_lower() << " executes init_server_hook" << std::endl;
+            try
+            {
+                device->server_init_hook();
+            }
+            catch (const DevFailed &devFailed)
+            {
+                device->set_state(FAULT);
+
+                std::ostringstream ss;
+                ss << "Device[" << device->get_name_lower() << "] server_init_hook has failed due to DevFailed:"
+                   << std::endl;
+                ss << devFailed;
+
+                device->set_status(ss.str());
+            }
+        }
+    }
+#else
+    std::vector<DeviceClass*> &dclass_vector = this->get_class_list();
+    std::vector<DeviceClass*>::iterator dclass_vector_pos, dclass_vector_end;
+    for (dclass_vector_pos = dclass_vector.begin(), dclass_vector_end = dclass_vector.end(); dclass_vector_pos != dclass_vector_end; ++dclass_vector_pos)
+    {
+        std::vector<DeviceImpl*> &device_vector = (*dclass_vector_pos)->get_device_list();
+        std::vector<DeviceImpl*>::iterator device_vector_pos, device_vector_end;
+        for (device_vector_pos = device_vector.begin(), device_vector_end = device_vector.end(); device_vector_pos != device_vector_end; ++device_vector_pos)
+        {
+            DeviceImpl *device = *device_vector_pos;
+            cout4 << "Device " << device->get_name_lower() << " executes init_server_hook" << std::endl;
+            try
+            {
+                device->server_init_hook();
+            }
+            catch(const DevFailed& devFailed){
+                device->set_state(FAULT);
+
+                std::ostringstream ss;
+                ss << "Device[" << device->get_name_lower() << "] server_init_hook has failed due to DevFailed:"
+                   << std::endl;
+                ss << devFailed;
+
+                device->set_status(ss.str());
+            }
+        }
+    }
+#endif
+}
+
 //+----------------------------------------------------------------------------------------------------------------
 //
 // method :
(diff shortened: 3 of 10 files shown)


bourtemb:
@ingvord, FYI, I just updated .travis/debian7/Dockerfile in tango-9-lts branch to fix the problems with Debian 7 Travis CI build.

Ingvord:
@bourtemb, thanks! That would probably fix Travis in this branch too

Ingvord:
@bourtemb , I think this is more or less done.

A few notes:
- I pass DServer const reference into server_init_hook - maybe it will be useful
- init_device call has been removed from DServer ctr - I prefer when there is no side effects
- I am not sure about DevFailed handling in DServer::server_init_hook - maybe we have some utility function for such case, i.e. set device status using DevFailed
- I use **auto** and since Travis does not complain even in Debian 7, I guess it is OK. But final call is up to you of course

@t-b, @mliszcz , could you please also review this PR? I do not see you in the reviwers list, so can not add you as reviewers. Thanks!

cppapi/server/device.h
```diff
@@ -454,6 +454,18 @@ class DeviceImpl : public virtual POA_Tango::Device
...
+ *
+ * Default method to implement an action necessary on a device after server has been initialized.
+ *
+ * @exception DevFailed This method does not throw exception but a
+ * redefined method can.
+ * Click <a href="https://tango-controls.readthedocs.io/en/latest/development/advanced/IDL.html#exceptions">here</a> to read
+ * <b>DevFailed</b> exception specification
+ */
+    virtual void server_init_hook(const DServer &)
+    {};
```
bourtemb: This Pull Request will result in a new major version of Tango because we lose the binary compatibility by adding this virtual function. This is not a problem but we should keep this in mind when updating the version number.
By the way, I created [this wiki page](https://github.com/tango-controls/cppTango/wiki/Policies-Binary-Compatibility-Issues) which refers to a web page explaining the Do's and Don'ts to preserve binary compatibility.

cppapi/server/dserver.h
```diff
@@ -171,7 +173,6 @@ protected :
 	bool is_event_name(string &);
 	bool is_ip_address(string &);
 
-	bool			from_constructor;
```
bourtemb: from_constructor does not seem to be used any longer indeed.

cppapi/server/dserverclass.cpp
```diff
@@ -1845,12 +1845,15 @@ void DServerClass::device_factory(const Tango::DevVarStringArray *devlist_ptr)
 //
 // Create device and add it into the device list
 //
+        auto dserver = new DServer(this,
```
bourtemb: We should not use auto if we want to be able to compile on old windows compilers.

cpp_test_suite/cpp_test_ds/DevTest.cpp
```diff
@@ -217,6 +217,11 @@ void DevTest::always_executed_hook()
 
 }
 
+void DevTest::server_init_hook()
+{
+    cout2 << "In server_init_hook method" << endl;
```
mliszcz: Consider using `std::endl`, otherwise it may break with #528 (general comment, applies to other places as well)

cppapi/server/except.cpp
```diff
@@ -45,6 +45,19 @@ namespace Tango
 char Except::mess[256];
 omni_mutex Except::the_mutex;
 
+std::ostream &operator<<(std::ostream &ss, const Tango::DevFailed &devFailed)
+{
+    for (int i = 0, size = devFailed.errors.length(); i < size; ++i)
+    {
+        DevError error = devFailed.errors[i];
```
mliszcz: Consider using reference instead of a copy. `DevError` has CORBA string members which allocate and copy memory from their copy-ctors.

cppapi/server/except.cpp
```diff
@@ -45,6 +45,19 @@ namespace Tango
...
+        ss << "\tOrigin:" << error.origin << "\n";
+        ss << "\tSeverity:" << error.severity << "\n";
+        ss << "\tReason:" << error.reason << "\n";
+        ss << "\tDescription:" << error.desc << "\n";
```
mliszcz: Do you intend to use this operator in future as a generic utility for printing DevError or is it just for exception handler in dserver's init_hook? In former case, would concurrent access be a problem? In this example we will get mixed lines:
```
thread 1: std::cout << DevFailed{ ... } << std::cout;
thread 2: std::cout << DevFailed{ ... } << std::cout;
```
Buffering parts of the output in `ostringstream` before writing may help (probably depends on whether underlying `write` call is synchronized at OS level).

cpp_test_suite/cpp_test_ds/DevTest.cpp
Ingvord: This is a DevTest - a part of cpp_test_suite, so I think it is not relevant here to use `std::` as anyway in other parts of this class it is missing. I would wait till #528 is merged and then fix the issues if any.

cppapi/server/except.cpp
Ingvord: It is intended to use this operator on a stack trace local stringstream, as it is just a convenience function. I will mark it as **inline**.

cppapi/server/except.cpp
Ingvord: ok.

Ingvord:
Looks like even without **auto** Windows fails to compile simple *for*-loop:

```
3>.\dserver.cpp(486) : error C2143: syntax error : missing ',' before ':'
15843>.\dserver.cpp(487) : error C2143: syntax error : missing ';' before '{'
15853>.\dserver.cpp(488) : error C2143: syntax error : missing ',' before ':'
15863>.\dserver.cpp(489) : error C2143: syntax error : missing ';' before '{'
1587
```

where dserver.cpp#486

```cpp
for (DeviceClass *dclass : this->get_class_list())
```

Any help from Windows expert is very well appreciated! Thanks.

mliszcz:
@Ingvord, range-based for is c++11 feature (see: https://en.cppreference.com/w/cpp/language/range-for)

bourtemb:
@mliszcz is correct. You are using a C++ 11 feature. It's not Windows' fault, this time ;-) ...
I think after the release of cppTango 9.3.3, it is time to apply what was decided during the last Tango kernel meeting at Krakow.
If my notes are correct, it was decided to remove the support for MSVC9. MSVC10 and MSVC12.
So now that cppTango 9.3.3 is out, we could remove the appveyor builds for these compilers.
If some users still need support for these old compilers, they will need to maintain the compatibility in their own fork.

Ingvord:
@mlizszc, @bourtemb thanks a lot for the clarification.  I guess we need some macro here,  i.e. if old windows then do it the old way with iterators, otherwise use range for. I will try to find existing pattern for this...

bourtemb:
If you want to maintain the compatibility, you can indeed use something like:
```
#ifdef HAS_RANGE_BASE_FOR
// C++ 11 code
#else
// code for old compilers
#endif
```

You can search for `#ifdef HAS_RANGE_BASE_FOR` in the code to get some examples.
For instance in cppapi/client/devapi_base.cpp

Ingvord:
@bourtemb , done in e52e2a8. A bit of a useless job as probably as this PR will be released for sure after 9.3.3 due to new virtual function. Let's see if AppVeyor is now happy...

bourtemb:
@reszelaz, could you please review/test this PR and tell us whether it fits your needs?

reszelaz:
I'm quite busy this week to test it, but at the same time don't want to block you.
If I understand correctly this will work in the following way

```
for class in server_classes:
    for device in class:
        device.init_device()

export_device_server_admin_device()

for class in server_classes:
    for device in class:
        device.server_init_hook()
```

This should work for us!

Ingvord:
@t-b, thanks for the review!

> The output operator overload needs some more love:

It seems you have already performed required changes, hence diff. If it is so, could you please commit those directly into the branch (Note it is not in my fork, so you should be able to do so)? 

> Does not tell me what has changed and why.

I usually squash and merge so these commits will vanish anyway... and sometimes there is no logic in my commits, that's true :)

t-b:
@Ingvord Will do!

t-b:
@Ingvord Done. I've used your email as author on the important commit. Feel free to suggest/rewrite otherwise.

cppapi/server/dserver.cpp
```diff
@@ -484,6 +480,61 @@ void DServer::init_device()
...
 }
 
+void DServer::server_init_hook()
+{
+#ifdef HAS_RANGE_BASE_FOR
+    for (DeviceClass *dclass : this->get_class_list())
+    {
+        for (DeviceImpl *device : dclass->get_device_list())
+        {
+            cout4 << "Device " << device->get_name_lower() << " executes init_server_hook" << std::endl;
```
bourtemb: " executes init_server_hook"  => " executes server_init_hook()" 
Same at line 518

Ingvord:
@t-b , thanks!

cppapi/server/dserver.cpp
t-b: @bourtemb Thx. The fix was done by @Ingvord already. I've folded into the correct commit.

Issue #550, opened 2019-05-09
Forwarding State


lorenzopivetta:
Hi,
is there any issue with forwarding the State attribute? Consider the following scenario:
- device "lorenzo/uno/uno" provides attribute "uno" (DevDouble,rw) and "anotherState" (DevState,rw).
- device "lorenzo/due/due" is configured with the following forwarded attirbutes:
    1) unoFwd with __root_attr lorenzo/uno/uno/uno 
    2) anotherStateFwd with __root_attr lorenzo/uno/uno/anotherState
    3) unoStateFwd with __root_attr lorenzo/uno/uno/State

Reading the forwarded attribute from device "lorenzo/due/due" works for case 1 and 2; events work too.
Reading forwarded attribute from device "lorenzo/due/due" for case 3 results in a segmentation:

...
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0xb0dfab70 (LWP 16669)]
0xb7d549ae in void Tango::FwdAttribute::set_local_attribute<Tango::DevVarStateArray>(Tango::DeviceAttribute&, Tango::DevVarStateArray*&) ()
   from /usr/local/tango-9.3.3-debug/lib/libtango.so.9

Tests with Tango 9.3.3, OmniORB 4.2.3, ZeroMQ 4.0.8

bourtemb:
Hi @lorenzopivetta 

Thank you very much for the bug report. I can reproduce the problem.
This bug does not occur when forwarding a DevState Read Only or RW scalar attribute which is different than the attribute named "State".

Issue #512, opened 2019-01-10
DevString Forwarded Attribute: is it possible?


alexElettra:
Hello,
I have created a very simple ds with pogo 9.4.5: it has only one FwdAttr, compile: ok,
db config of __root_att with the string_scalar attribute of TangoTest.
At the very first read of the FwdAttr I get a seg fault:

```
 (gdb) bt
#0  0xb76eb424 in __kernel_vsyscall ()
#1  0xb6757681 in *__GI_raise (sig=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:64
#2  0xb675aab2 in *__GI_abort () at abort.c:92
#3  0xb679124d in __libc_message (do_abort=2, fmt=0xb6868e58 "*** glibc detected *** %s: %s: 0x%s ***\n") at ../sysdeps/unix/sysv/linux/libc_fatal.c:189
#4  0xb679b341 in malloc_printerr (action=<value optimized out>, str=0x6 <Address 0x6 out of bounds>, ptr=0x975f8b0) at malloc.c:6266
#5  0xb679c5be in munmap_chunk (p=<value optimized out>) at malloc.c:3540
#6  0xb698c741 in operator delete(void*) () from /usr/lib/libstdc++.so.6
#7  0xb698c79d in operator delete[](void*) () from /usr/lib/libstdc++.so.6
#8  0xb7348b34 in Tango::Attribute::set_value (this=0x978f2c0, p_data=0x975f8b0, x=1, y=0, release=true) at ../../../../lib/cpp/server/attrsetval.cpp:746
#9  0xb7348d0b in Tango::Attribute::set_value_date_quality (this=0x978f2c0, p_data=0x975f8b0, t=..., qual=Tango::ATTR_VALID, x=1, y=0, release=<value optimized out>)
    at ../../../../lib/cpp/server/attrsetval.cpp:1947
#10 0xb741f770 in void Tango::FwdAttribute::set_local_attribute<Tango::DevVarStringArray>(Tango::DeviceAttribute&, Tango::DevVarStringArray*&) () from /usr/local/tango-9.2.5a/lib/libtango.so.9
#11 0xb741bdbc in Tango::FwdAttr::read (this=0x9771998, dev=0x9771e70, attr=...) at ../../../../lib/cpp/server/fwdattrdesc.cpp:380
[...]

PR #552, opened 2019-05-14
Fix crash when reading a forwarded State attribute #550

bourtemb:
- [x] Add a test to read a State forwarded attribute 
- [x] Fix issue

### DIFF
--- a/cpp_test_suite/new_tests/cxx_fwd_att.cpp
+++ b/cpp_test_suite/new_tests/cxx_fwd_att.cpp
@@ -796,6 +796,28 @@ public:
 		lo = device2->is_locked();
 		TS_ASSERT(lo == false);
 	}
+
+	void test_reading_state_forwarded_attribute(void)
+	{
+		DeviceAttribute state_attr;
+
+		// read attributes
+		TS_ASSERT_THROWS_NOTHING(state_attr = fwd_device->read_attribute("fwd_state"));
+		DevState the_state;
+		if(state_attr >> the_state)
+		{
+			TS_ASSERT(state_attr.name == "fwd_state");
+			TS_ASSERT(state_attr.quality == Tango::ATTR_VALID);
+			TS_ASSERT(state_attr.dim_x == 1);
+			TS_ASSERT(state_attr.dim_y == 0);
+			cout << "the_state = " << the_state << endl;
+			TS_ASSERT(the_state == Tango::ON);
+		}
+		else
+		{
+			TS_FAIL("Could not extract attribute value to DevState");
+		}
+	}
 };
 
 void FwdAttTestSuite::EventCallBack::push_event(Tango::EventData* event_data)
--- a/cppapi/client/devapi_attr.cpp
+++ b/cppapi/client/devapi_attr.cpp
@@ -3181,6 +3181,14 @@ bool DeviceAttribute::operator >> (std::vector<DevState>& datum)
 	if ( ret == false)
 		return false;
 
+	if(d_state_filled)
+	{
+		datum.resize(1);
+		datum[0] = d_state;
+		d_state_filled = false;
+		return true;
+	}
+
 	if (StateSeq.operator->() != NULL)
 	{
 		if (StateSeq->length() != 0)
@@ -3193,7 +3201,7 @@ bool DeviceAttribute::operator >> (std::vector<DevState>& datum)
        	 	}
 		}
 		else
-			ret = false;
+			return false;
 	}
 	else
 	{
@@ -3573,6 +3581,15 @@ bool DeviceAttribute::operator >> (DevVarStateArray* &datum)
 	if ( ret == false)
 		return false;
 
+	if(d_state_filled)
+	{
+		Tango::DevState *tmp_ptr = new Tango::DevState[1];
+		*tmp_ptr = d_state;
+		datum = new DevVarStateArray(1,1,tmp_ptr,true);
+		d_state_filled = false;
+		return true;
+	}
+
 	if (StateSeq.operator->() != NULL)
 	{
 		if (StateSeq->length() != 0)
@@ -3580,7 +3597,7 @@ bool DeviceAttribute::operator >> (DevVarStateArray* &datum)
 			datum = StateSeq._retn();
 		}
 		else
-			ret = false;
+			return false;
 	}
 	else
 	{
--- a/cppapi/server/attrsetval.cpp
+++ b/cppapi/server/attrsetval.cpp
@@ -1456,7 +1456,7 @@ void Attribute::set_value(Tango::DevState *p_data,long x,long y,bool release)
 				*tmp_ptr = *p_data;
 				value.state_seq = new Tango::DevVarStateArray(data_size,data_size,tmp_ptr,release);
 				if (is_fwd_att() == true)
-					delete p_data;
+					delete [] p_data;
 				else
 					delete p_data;
 			}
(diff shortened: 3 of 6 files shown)


bourtemb:
Travis CI should have failed for the first commit (Add a test to read State forwarded attribute).
It looks like it didn't fail in this PR because of the following reasons:
When the PR was created with only the first commit, Travis could not build it immediately because some other jobs were already running and the job was put in the waiting queue.
I pushed the commit with the fix only a few minutes after creating the Pull Request.
It looks like Travis is always getting the HEAD of the branch when testing/building a pull request (quite confusing! See https://github.com/travis-ci/travis-ci/issues/8577) so it looks like when it finally could build for the first commit, the fix commit was already there. So it passed...

So I think we should automatically cancel older not already built Travis jobs when a new commit is added to a PR because the old ones become meaningless and bring confusion.

If you look at my fork, you will see Travis failed for the first commit and succeeded for the second one:
https://github.com/bourtemb/cppTango/commits/fix-550

cpp_test_suite/new_tests/cxx_fwd_att.cpp
```diff
@@ -803,6 +803,28 @@ class FwdAttTestSuite: public CxxTest::TestSuite
...
 	}
+
+	void test_reading_state_forwarded_attribute(void)
+	{
+		DeviceAttribute state_attr;
+
+		// read attributes
+		TS_ASSERT_THROWS_NOTHING(state_attr = fwd_device->read_attribute("fwd_state"));
+		DevState the_state;
+		if(state_attr >> the_state)
```
mliszcz: Shouldn't you extract to a  `std::vector<DevState>`?
I think this one is called now: `bool DeviceAttribute::operator >> (DevState &datum)`
And there were no (direct) changes in above operator.

cppapi/server/attrsetval.cpp
```diff
@@ -1457,7 +1457,7 @@ void Attribute::set_value(Tango::DevState *p_data,long x,long y,bool release)
 				*tmp_ptr = *p_data;
 				value.state_seq = new Tango::DevVarStateArray(data_size,data_size,tmp_ptr,release);
 				if (is_fwd_att() == true)
-					delete p_data;
+					delete [] p_data;
 				else
 					delete p_data;
```
mliszcz: consider using `SAFE_DELETE(p_data)` from `attribute.h`. There will be redundant check for `release == true` but you will not have duplicate code.

bourtemb: Macros are evil and not easy to debug and there was a redundant check. This is indeed why I didn't use the macro in this case.
SAFE_DELETE macro is convenient but it is a nightmare to debug issues in this macro, like https://github.com/tango-controls/cppTango/issues/512 for instance.

cpp_test_suite/new_tests/cxx_fwd_att.cpp
bourtemb: This is what we really want to test: what happens when we read a forwarded DevState attribute.
The change was indeed in DeviceAttribute::operator >> (DevState &datum) which is used internally when the user tries to read a DevState forwarded attribute.
It is actually used internally to read the orginal DevState attribute.

mliszcz: My point was that the change was done in:
```
bool DeviceAttribute::operator >> (vector<DevState>& datum)
```
and not in:
```
bool DeviceAttribute::operator >> (DevState &datum)
```

But maybe the control goes internally through the vector overload anyway.

cppapi/server/attrsetval.cpp
mliszcz: Maybe we should remove the macro and add something like
* `void safe_delete(T* pointer)` and/or
* `void safe_delete_if_needed(T* pointer, bool needed)`

Also, shall we define a new item for removing all macros? We may start from `server/attribute.h` file.

bourtemb: > Maybe we should remove the macro and add something like
> 
>     * `void safe_delete(T* pointer)` and/or
> 
>     * `void safe_delete_if_needed(T* pointer, bool needed)`

I would be in favour of such a change.

> Also, shall we define a new item for removing all macros? We may start from `server/attribute.h` file.

A new issue could be created indeed in order to replace existing macros when possible.

mliszcz: I've opened #553 to track this effort.

cpp_test_suite/new_tests/cxx_fwd_att.cpp
bourtemb: > But maybe the control goes internally through the vector overload anyway.

I understand why you had some doubts.
There was a change in bool DeviceAttribute::operator >> (vector<DevState>& datum) and no test was added to test that.
The control goes internally to bool DeviceAttribute::operator >> (DevVarStateArray* &datum) (which was changed in this PR too) when reading a DevState forwarded attribute.
It is called from template<typename T>void FwdAttribute::set_local_attribute(DeviceAttribute &da, T *&seq_ptr), which is called by void FwdAttr::read(DeviceImpl *dev,Attribute &attr) when reading a forwarded attribute.

bool DeviceAttribute::operator >> (vector<DevState>& datum) was also changed in this PR for consistency reasons.
A test could be added to test it.

mliszcz: Thanks! It's clear now.

cppapi/client/devapi_attr.cpp
```diff
@@ -3194,10 +3194,16 @@ bool DeviceAttribute::operator >> (vector<DevState>& datum)
...
+			return false;
 	}
 	else
 	{
+		if(d_state_filled)
+		{
+			datum.resize(1);
+			datum[0] = d_state;
+			return true;
+		}
```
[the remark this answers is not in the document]
bourtemb: It is not clear to me why d_state_filled is reset in the example you provided.
It is probably done to simulate the consumption of the data.
If you try to extract twice from the same DeviceData object, at the second invocation check_for_data() will return false because is_empty() will return true, because of this d_state_filled reset to false.
To be consistent, it would probably good to add `d_state_filled = false` in the methods I modified indeed.

t-b:
@bourtemb Could you rebase so that the review is about the code which will be merged?

bourtemb:
> The standard namespace is still imported in the tests, see
> 
> https://github.com/tango-controls/cppTango/blob/4d2ad6555fd6bc6e6603406ea83b6ed0ed797eb3/cpp_test_suite/new_tests/cxx_common.h#L9
> so [8da03eb](https://github.com/tango-controls/cppTango/commit/8da03eb1a270abbb2edf283083fc5c1de674888c) should not be needed.

@t-b, I think this one is needed because this commit concerns the FwdTest class which is part of the device server used to test the forwarded attributes. cxx_common.h is not included in this case.

I will take into account your 2 other comments. Thanks for the review.

bourtemb:
> The standard namespace is still imported in the tests, see
> https://github.com/tango-controls/cppTango/blob/4d2ad6555fd6bc6e6603406ea83b6ed0ed797eb3/cpp_test_suite/new_tests/cxx_common.h#L9
> 
> so [8da03eb](https://github.com/tango-controls/cppTango/commit/8da03eb1a270abbb2edf283083fc5c1de674888c) should not be needed.

I actually added it to fix Travis failure reported for the previous commit: https://travis-ci.org/tango-controls/cppTango/builds/617855756

I pushed commit 5457a4adcc1f8c3391b49579906877961cffdafe for d_state_filled consistency issue.

I will then rebase/reorder the commits so at the end, there will be only 2 commits: One for the new test and one for the fix.

t-b:
> @t-b, I think this one is needed because this commit concerns the FwdTest class which is part of the device server used to test the forwarded attributes. cxx_common.h is not included in this case.

Aha, got it.

bourtemb:
@mliszcz , good for you?

Issue #330, opened 2016-12-12
Segmentation fault after execution state command/attribute, at check_level_alarm


aleolivo:
Using Tango 9.2.2 (with omniORB-4.2.1 and with zermq-4.0.7):
A segmentation fault occurred after the **state command/attribute** execution, by client (tested with Jive or AtkPanel), if the **alarm threshold** on one or more attributes is configured and if an **exception is thrown during the "read"** method of an attribute.

I did this simple testing device server [testThreshold.zip](https://github.com/tango-controls/cppTango/files/646288/testThreshold.zip) (threshold-srv) to point out the segmentation fault. It has two commands: "On" and "Off" in order to change the device state and has two attributes: AttrUShort (DevUShort) and ActivateException (DevBoolean). I configured an alarm threshold on the first attribute and when the second attribute is set to "true" it makes the read_AttrUShort() throw an exception.
To test the crash launch the AtkPanel, excecute "On" command and check the ActivateException checkbox.

_Tests done_ :

- It crashes even if I configure the **threshold at runtime** time in tango db;
- It crashes if compiled with Tango-9.2.2 **without p922_1 patch** and **with Tango-9.1.0** (with omniORB-4.2.0 and with zermq-4.0.5);
- It crashes with **every type** of numeric attributes;
- It **doesn't crash with tango 8.1.2** (with omniORB-4.1.6 and with zermq-4.), when the exception is thrown by the read_<attribute> the client show the exception;
- activating **Polling** works around the problem

_Debugger_ :

- The error:

> Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0xb2c87b70 (LWP 5392)]
0xb7b374b7 in Tango::Attribute::check_level_alarm (this=0x8099688) at ../../../../lib/cpp/server/attribute.cpp:1898
1898						if ((*value.ush_seq)[i] <= min_alarm.ush)
- Back trace:
 > #0  0xb7b374b7 in Tango::Attribute::check_level_alarm (this=0x8099688) at ../../../../lib/cpp/server/attribute.cpp:1898
#1  0xb7b36efb in Tango::Attribute::check_alarm (this=0x8099688) at ../../../../lib/cpp/server/attribute.cpp:1703
#2  0xb7c33e4a in Tango::MultiAttribute::check_alarm (this=0x8096fd8, ind=0) at ../../../../lib/cpp/server/multiattribute.h:221
[...]

PR #555, opened 2019-05-26
Fix crash during alarm state evaluation if attribute value is not set (#330)

mliszcz:
Fix for following scenario:
* there is an alarm configured for an attribute
* Device::dev_state is evaluated
* attribute's read callback throws DevFailed before calling set_value
* attempt to evaluate alarm fails (segfault during access to value which is not set)

Proposed correction is to silently skip such attribute during alarm evaluation (it will not contribute to device's ALARM state). @bourtemb, is this acceptable solution?

I have two other points I'd like to discuss:
* To simplify the reproduction, I had to reset attribute's value pointer after it was deleted. Otherwise something else was allocated in that memory and I was getting garbage read instead of a segfault. I propose to define an action to reset all the pointers after corresponding memory is deleted (applies only to pointers that can be reused, e.g. global or member variables). We may find some new issues that way.
* There is a lot of duplicated code in `attribute.cpp` and in `attrsetval.cpp` for different data types. It could be easily removed and replaced with template code. Maybe we should define an action to remove duplications (for now I've re-set only `value.lg_seq` - see previous point).

BTW all commits are failed on appveyor because MSVC don't like `not` operator. I've fixed this just now.

### DIFF
--- a/cpp_test_suite/cpp_test_ds/DevTest.cpp
+++ b/cpp_test_suite/cpp_test_ds/DevTest.cpp
@@ -117,6 +117,7 @@ void DevTest::init_device()
 	Short_attr_except = false;
 	if (tg->is_svr_starting() == true || tg->is_device_restarting(device_name) == true)
 		Short_attr_w_except = false;
+	Long_attr_except = false;
 	event_change_attr_except = false;
 	event_quality_attr_except = false;
 	event_throw_out_of_sync = false;
@@ -1281,8 +1282,15 @@ void DevTest::read_Short_attr(Tango::Attribute &att)
 
 void DevTest::read_Long_attr(Tango::Attribute &att)
 {
-      	cout << "[DevTest::read_attr] attribute name Long_attr" << std::endl;
-      	att.set_value(&attr_long);
+    cout << "[DevTest::read_attr] attribute name Long_attr" << std::endl;
+    if (Long_attr_except)
+    {
+        Tango::Except::throw_exception(
+            "Long_attr_except",
+            "Test exception is enabled for this attribute",
+            "DevTest::read_Long_attr");
+    }
+    att.set_value(&attr_long);
 }
 
 void DevTest::read_Long64_attr(Tango::Attribute &att)
--- a/cpp_test_suite/cpp_test_ds/DevTest.h
+++ b/cpp_test_suite/cpp_test_ds/DevTest.h
@@ -289,6 +289,7 @@ protected :
 
 	bool 				Short_attr_except;
 	bool 				Short_attr_w_except;
+	bool 				Long_attr_except;
 	bool 				event_change_attr_except;
 	bool 				event_quality_attr_except;
 	bool 				event_throw_out_of_sync;
--- a/cpp_test_suite/cpp_test_ds/IOMisc.cpp
+++ b/cpp_test_suite/cpp_test_ds/IOMisc.cpp
@@ -665,41 +665,23 @@ CORBA::Any *IOAttrThrowEx::execute(Tango::DeviceImpl *device, const CORBA::Any &
 	const Tango::DevVarShortArray *in;
 	extract(in_any, in);
 
-	if ((*in)[0] == 0)
-	{
-		if ((*in)[1] == 0)
-			(static_cast<DevTest *>(device))->Short_attr_except = false;
-		else
-			(static_cast<DevTest *>(device))->Short_attr_except = true;
-	}
-	else if ((*in)[0] == 1)
-	{
-		if ((*in)[1] == 0)
-			(static_cast<DevTest *>(device))->event_change_attr_except = false;
-		else
-			(static_cast<DevTest *>(device))->event_change_attr_except = true;
-	}
-	else if ((*in)[0] == 2)
-	{
-		if ((*in)[1] == 0)
-			(static_cast<DevTest *>(device))->event_quality_attr_except = false;
-		else
-			(static_cast<DevTest *>(device))->event_quality_attr_except = true;
-	}
-	else if ((*in)[0] == 3)
-	{
-		if ((*in)[1] == 0)
-			(static_cast<DevTest *>(device))->event_throw_out_of_sync = false;
-		else
-			(static_cast<DevTest *>(device))->event_throw_out_of_sync = true;
-	}
-	else if ((*in)[0] == 4)
-	{
-		if ((*in)[1] == 0)
-			(static_cast<DevTest *>(device))->Short_attr_w_except = false;
-		else
-			(static_cast<DevTest *>(device))->Short_attr_w_except = true;
-	}
+	DevTest& dev = static_cast<DevTest&>(*device);
+
+	const int flag_disc = (*in)[0];
+	const bool flag_value = (*in)[1];
+	bool default_flag = false;
+
+	bool& flag =
+		(flag_disc == 0) ? dev.Short_attr_except :
+		(flag_disc == 1) ? dev.event_change_attr_except :
+		(flag_disc == 2) ? dev.event_quality_attr_except :
+		(flag_disc == 3) ? dev.event_throw_out_of_sync :
+		(flag_disc == 4) ? dev.Short_attr_w_except :
+		(flag_disc == 5) ? dev.Long_attr_except :
+		default_flag;
+
+	flag = flag_value;
+
 	return insert();
 }
 
--- a/cpp_test_suite/cpp_test_ds/TypeCmds.cpp
+++ b/cpp_test_suite/cpp_test_ds/TypeCmds.cpp
@@ -1,9 +1,5 @@
 #include "TypeCmds.h"
 
-#define DEV_DEBUG_STREAM(dev) \
-  if (dev->get_logger()->is_debug_enabled())\
-    dev->get_logger()->debug_stream()
-
 //+----------------------------------------------------------------------------
 //
 // method : 		IOVoid::IOVoid()
--- a/cpp_test_suite/new_tests/cxx_attr_misc.cpp
+++ b/cpp_test_suite/new_tests/cxx_attr_misc.cpp
@@ -1021,6 +1021,54 @@ cout << "status = " << status << endl;
 		dout >> status;
 		TS_ASSERT(strcmp(status,"The device is in ON state.") == 0);
 	}
+
+	void set_attribute_exception_flag(short attribute_disc, bool enabled)
+	{
+		std::vector<short> flags(2);
+		flags[0] = attribute_disc;
+		flags[1] = enabled ? 1 : 0;
+		DeviceData data;
+		data << flags;
+		TS_ASSERT_THROWS_NOTHING(device1->command_inout("IOAttrThrowEx", data));
+	}
+
+	void __assert_dev_state(DevState expected, const char* file, const char* line)
+	{
+		std::string message = std::string("Called from ") + file + ":" + line;
+
+		DevState state;
+		DeviceData data;
+		TSM_ASSERT_THROWS_NOTHING(message, data = device1->command_inout("State"));
+		data >> state;
+		TSM_ASSERT_EQUALS(message, expected, state);
+	}
+
+#define __QUOTE(x) #x
+#define QUOTE(x) __QUOTE(x)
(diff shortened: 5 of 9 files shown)


mliszcz:
@t-b thanks for checking this PR.

> Setting the pointer to nullptr after deleting sounds like something we want to do for the other types as well. What is the reason why that is only done for this type?

Zeroing the pointer is not necessary for the correction to work. It is just to simplify the reproduction.

I agree it is a good thing to reset the pointers, but I'd like to have a dedicated issue to fix this kind of problem in the whole codebase. Some examples in other places:

https://github.com/tango-controls/cppTango/blob/ab05abc289d187130614966e4c79e5390ca41b16/cppapi/server/device.cpp#L2944-L2950

https://github.com/tango-controls/cppTango/blob/ab05abc289d187130614966e4c79e5390ca41b16/cppapi/server/device.h#L3683-L3684

https://github.com/tango-controls/cppTango/blob/ab05abc289d187130614966e4c79e5390ca41b16/cppapi/server/device.cpp#L5333-L5336

If the kernel team agrees that we need this kind of change, and if it is in [Tango 9 LTS project](https://github.com/tango-controls/cppTango/projects/5), I can work on this task and priotitize it over other S2Innovation activities.

Note that some refactoring will be needed because sometimes pointer to attribute data is passed by value and delete is done using copied pointer, not the original one.

CC @bourtemb

t-b:
@mliszcz 
> I agree it is a good thing to reset the pointers, but I'd like to have a dedicated issue to fix this kind of problem in the whole codebase. 

I think this would be a reasonable thing to do. We should then also define some templates which will do delete and set to null in one go. Having two templates would solve the issue of `delete` vs `delete []`.

bourtemb:
> > I agree it is a good thing to reset the pointers, but I'd like to have a dedicated issue to fix this kind of problem in the whole codebase.
> 
> I think this would be a reasonable thing to do. We should then also define some templates which will do delete and set to null in one go. Having two templates would solve the issue of `delete` vs `delete []`.

I agree this would be a reasonable thing to do. This will probably help to discover other issues.

andygotz:
Fixing all places in the library to reset pointers sounds like a big job potentially. How much work do you estimate this to be and how can we test all the changes? Do we have tests for all cases?

bourtemb:
> I agree it is a good thing to reset the pointers, but I'd like to have a dedicated issue to fix this kind of problem in the whole codebase. Some examples in other places:
> 
> https://github.com/tango-controls/cppTango/blob/ab05abc289d187130614966e4c79e5390ca41b16/cppapi/server/device.cpp#L2944-L2950

Well, in this specific case, ptr gets out of scope immediately after. So, in this specific case, it looks like there is no interest in resetting it.

I think we should do this only for pointers which can be reused somewhere else.

mliszcz:
> Fixing all places in the library to reset pointers sounds like a big job potentially. How much work do you estimate this to be and how can we test all the changes? Do we have tests for all cases?

From what I have checked, I think the problem is only with pointer(s) to attribute data. These pointers are deleted (at least) in: attribute.cpp, device.cpp, attrsetval.cpp in multiple places. 

Zeroing the pointer is quite trivial job, but the problem is sometimes there is no access to the original pointer, but to a copy. Thus, I estimate the effort to 16h (two days, for potential refactoring).

It is not possible to test such change.

> Well, in this specific case, ptr gets out of scope immediately after. So, in this specific case, it looks like there is no interest in resetting it.

> I think we should do this only for pointers which can be reused somewhere else.

This specific case also needs to be fixed.The *copy* of a pointer goes out of scope. See what is returned from `get_short_value`:
https://github.com/tango-controls/cppTango/blob/ab05abc289d187130614966e4c79e5390ca41b16/cppapi/server/attribute.h#L2186

```cpp
 case Tango::DEV_ENUM : 
 { 
     Tango::DevVarShortArray *ptr = att.get_short_value(); 
     a <<= *ptr; 
     delete ptr; 
     // somehow we need to reset: att.value.sh_seq = nullptr
     break; 
}
```

bourtemb:
> // somehow we need to reset: att.value.sh_seq = nullptr
@mliszcz, OK, I see your point now. Thanks for the clarification.

mliszcz:
@t-b thanks for the review! I've applied changes per your comments:
> fbcbc8a (Reset Attribute::value.lg_seq pointer after delete, 2019-05-26)

> Setting the pointer to nullptr after deleting sounds like something we want to do for the other types as well. What is the reason why that is only done for this type?

Changed to: Reset Attribute::value pointers in Attribute::delete_seq bf8ce6c
Pointers of other types are reset as well, in `del_seq` function. For global reset action there is #565.

> 49ad354 (Fix appveyor builds, 2019-05-27)
That should probably be folded into 0ef70cc as that commit introduced the incompatibility.

Also, this one was squashed into previous commit:
> 65e8e92 (Initialize exception flag for Long_attr in DevTest, 2019-05-26)

@bourtemb, @andygotz I've opened #565 to reset the pointers. Lets use it for further discussions on that topic and leave this PR for crash correction review.

bourtemb:
> Proposed correction is to silently skip such attribute during alarm evaluation (it will not contribute to device's ALARM state). @bourtemb, is this acceptable solution?

Very good question.
I talked a bit with some colleagues and we think it would be more consistent to switch the global state to UNKNOWN (if State evaluation was triggered (State was ON before evaluation)) in this specific case and to add an explanation in the status, with the received exception description...
It would be good to get some feedback from the users on this point actually.

>     * There is a lot of duplicated code in `attribute.cpp` and in `attrsetval.cpp` for different data types. It could be easily removed and replaced with template code. Maybe we should define an action to remove duplications (for now I've re-set only `value.lg_seq` - see previous point).

Yes, we can create an issue for that.

mliszcz:
Thanks @bourtemb!

> Very good question.
> I talked a bit with some colleagues and we think it would be more consistent to switch the global state to UNKNOWN (if State evaluation was triggered (State was ON before evaluation)) in this specific case and to add an explanation in the status, with the received exception description...
> It would be good to get some feedback from the users on this point actually.

I can change it the way you propose. Would it be possible for you to ask some users (I guess there is a lot of them at ESRF) for feedback?

> > ```
> > * There is a lot of duplicated code [...]
> > ```
> 
> Yes, we can create an issue for that.

Done: https://github.com/tango-controls/cppTango/issues/569

mliszcz:
Hi @bourtemb,
Thanks for triggering the discussion on the kernel list.
So far we have a few contradictory opinions. Do you think we can make any conclusion now?

Also, up until now we were discussing "exception" scenario (as in original issue), but the same problem will be visible if the user will simply forget to call `Attribute::set_value(...)`. i.e. all these shall crash equally:
```cpp
void DevTest::read_Long_attr(Tango::Attribute &att)
{
    Tango::Except::throw_exception(/* ... */);
    att.set_value(&attr_long);
}
void DevTest::read_Long_attr(Tango::Attribute &att)
{
    // do nothing
}
void DevTest::read_Long_attr(Tango::Attribute &att)
{
    att.set_quality(ATTR_CHANGING);
}
```

bourtemb:
> Hi @bourtemb,
> Thanks for triggering the discussion on the kernel list.
> So far we have a few contradictory opinions. Do you think we can make any conclusion now?

Indeed with have contradictory opinions.
I triggered the discussion with my colleagues at the ESRF and their opinion was that in the case of an exception thrown by an attribute with an alarm threshold defined, the device state should switch to ALARM and the status should report that this attribute has thrown an exception.

> 
> Also, up until now we were discussing "exception" scenario (as in original issue), but the same problem will be visible if the user will simply forget to call `Attribute::set_value(...)`. i.e. all these shall crash equally:
> 
> ```c++
> void DevTest::read_Long_attr(Tango::Attribute &att)
> {
>     Tango::Except::throw_exception(/* ... */);
>     att.set_value(&attr_long);
> }
> void DevTest::read_Long_attr(Tango::Attribute &att)
> {
>     // do nothing
> }
> void DevTest::read_Long_attr(Tango::Attribute &att)
> {
>     att.set_quality(ATTR_CHANGING);
> }
> ```
> 
> Do you think we need to differentiate between all these scenarios?

The use case where the user does not set the attribute value is actually currently considered as a bug on the device server programmer side. I think it is a mistake of the programmer...
Maybe we can prevent the device server to crash in this case...?

mliszcz:
Thanks!

> The use case where the user does not set the attribute value is actually currently considered as a bug on the device server programmer side. I think it is a mistake of the programmer...
Maybe we can prevent the device server to crash in this case...?

Crash will not happen once this PR is applied even in it's current state.

> the device state should switch to ALARM and the status should report that this attribute has thrown an exception.

I'll then update the PR accordingly. I propose the following:

* If exception is thrown during read:
  * State will be set to ALARM
    * `Attribute::check_alarm()` will return `true` if `get_value_flag()` is false
  * Status will be updated with exception's message
    * **I need to store this message/exception in `Attribute` class**
    * `MultiAttribute::read_alarm` will update status with the message if attribute `is_alarmed()` and `get_value_flag()` is false
* If set_value was never called:
  * State will be set to ALARM
    * `Attribute::check_alarm()` will return `true` if `get_value_flag()` is false
  * Status will be updated with sth like.: "No value set for attribute xxx"
    * `MultiAttribute::read_alarm` will update status if attribute `is_alarmed()` and `get_value_flag()` is false

Is this ok?

bourtemb:
I think we'll need to discuss a bit more and agree on this during the next Tango kernel teleconf meeting because there were some very different opinions on the kernel mailing list, and even among my colleagues, some of them had different opinions...
Be careful with the INVALID quality factor case where the value is not set and where I think the device state should not be switched to ALARM...

mliszcz:
Ok, I'll wait with any changes until we all agree on a way forward.

bourtemb:
> Proposed correction is to silently skip such attribute during alarm evaluation (it will not contribute to device's ALARM state). @bourtemb, is this acceptable solution?

It was agreed during today's Tango Kernel teleconf meeting to add a warning log when this use case occurs. 
State and Status will not be modified automatically by the Tango library in this use case.

mliszcz:
I've added a log:
```cpp
DEV_WARN_STREAM(dev) << "Attribute has no value, skipping alarm evaluation for: " << get_name() << std::endl;
```

mliszcz:
Thanks @bourtemb! I confirm there is an issue in current solution.
If alarm was already reported for a given attribute (which later starts to throw), we don't even run the alarm check again, but instead just check if quality has been changed and assume it is still in alarm state.

https://github.com/tango-controls/cppTango/blob/6a2857c32baba5454d936815279314e60b838b9b/cppapi/server/multiattribute.cpp#L1330-L1336

I'll correct this and update the PR.

mliszcz:
@bourtemb 
I've updated the test case to cover this scenario and moved the check to earlier step, immediately after exception is thrown from read callback: ca4bac405b72b34575d61e7a08733df635535bf3. I've set quality to INVALID, because (judging from the check few lines above), it is the only reasonable quality if attribute has no value. This was sufficient to skip alarm evaluation in `Multiattribute::check_alarm`.

```diff
diff --git a/cppapi/server/device.cpp b/cppapi/server/device.cpp
index ef21033b..3d28cb1f 100644
--- a/cppapi/server/device.cpp
+++ b/cppapi/server/device.cpp
@@ -1256,40 +1256,48 @@ Tango::DevState DeviceImpl::dev_state()
                                 continue;
                             }
                             attr_vect[att.get_attr_idx()]->read(this, att);
                             Tango::AttrQuality qua = att.get_quality();
                             if ((qua != Tango::ATTR_INVALID) && (att.get_value_flag() == false))
                             {
                                 TangoSys_OMemStream o;

                                 o << "Read value for attribute ";
                                 o << att.get_name();
                                 o << " has not been updated";
                                 o << "Hint: Did the server follow Tango V5 attribute reading framework ?" << std::ends;

                                 Except::throw_exception((const char *) API_AttrValueNotSet, o.str(),
                                                         (const char *) "DeviceImpl::dev_state");
                             }
                         }
                     }
                     catch (Tango::DevFailed &)
                     {
+                        if (!att.get_value_flag())
+                        {
+                            WARN_STREAM
+                                << "Attribute has no value, forcing INVALID quality for: "
+                                << att.get_name() << std::endl;
+                            att.set_quality(Tango::ATTR_INVALID);
+                        }
+
```

PR #557, opened 2019-05-31
Do not publish port in docker container in tests

mliszcz:
I propose to remove `-p` option when starting `tango_cs` container for tests.

It is not necessary to publish the port on host. We source generated `cpp_test_suite/tango_host` file, which specifies `TANGO_HOST` and uses container IP address directly.

With `-p` option, if you have tango already running on your system on port 10000, docker will refuse to start:
```
docker: Error response from daemon: driver failed programming external connectivity on endpoint pedantic_morse (b9bb28fc27ff178de0472035e0840bfda6c1291577e360013f4730dd8c3a9f5e): Error starting userland proxy: listen tcp 0.0.0.0:10000: bind: address already in use.
```


Ingvord:
I use this script not only for tests but to run Tango for instance to test REST. It is more convenient than just raw Tango docker stack as this one has more devices (DevTest + FwdTest etc). So I would leave `-p` as it is. Is it that critical to run these scripts on machines with Tango installed, I mean it is intended to use it on a raw machine, like Travis. For developer it is also better to start fresh Tango to reduce effect of the persistent properties etc which may affect test results i.e. less environment depended the whole infrastructure - it is easier to reproduce and transfer it. Just my IMHO

mliszcz:
Thanks! I think we can support REST testing usecase with something like:

```bash
PUBLISH_PORT="-p 10000:10000"
if lsof -tnPi TCP:10000 >/dev/null; then
    echo "Some process is already listening on port 10000, it will not be published"
    PUBLISH_PORT=""
fi
docker run $PUBLISH_PORT ... tango_cs
```

There is no technical reason to keep `-p` as tests are working without it. Let's give users freedom to configure (mess up) their system as they see fit.

I can put a note in `INSTALL.md` in test section that "it is recomended to shut down existing Tango instances before running the tests". (recommended != mandatory).

Ingvord:
@mliszcz , you are right in the first place. After carefully rethinking I would say it is better to remove this port, as this will also allow to run multiple Tango docker stacks (if we name them properly) in the future. This will allow us to test multiple Tango hosts devices communication etc.

I have also checked if REST can access Tango via IP and this works fine (I was not sure about dots).

**BTW** do you know if we can set Starter name properly e.g. to IP or define a host name for docker? Now it is some rubbish, like `1aef7207caf2` in the picture belo:.

rest

mliszcz:
`1aef7207caf2` is not rubbish but random (?) container hostname. You can provide your own hostname with `-h` (or `--hostname`). I can add it to this PR if you want, just tell me what value shall we use there.
https://docs.docker.com/config/containers/container-networking/#dns-services

If we are to touch that file, there is one more thing we can improve:
https://github.com/tango-controls/cppTango/blob/ab05abc289d187130614966e4c79e5390ca41b16/cpp_test_suite/environment/setup.sh#L33-L35
We could probably use:
```
tango_admin --ping-device sys/database/2 2
```
or
```
tango_admin --ping-database 2
```

The timeout (2 seconds in above examples) works in some unexpected way if the system is down:
```
$ time tango_admin --ping-device sys/database/2 1

real    0m12,642s
user    0m0,009s
sys     0m0,009s
$ time tango_admin --ping-device sys/database/2 2

real    0m25,807s
user    0m0,000s
sys     0m0,019s
$ time tango_admin --ping-device sys/database/2 3

real    0m38,948s
user    0m0,000s
sys     0m0,024s
$ time tango_admin --ping-device sys/database/2

real    0m6,101s
user    0m0,008s
sys     0m0,008s
```

Ingvord:
> 1aef7207caf2 is not rubbish but random (?) container hostname. You can provide your own hostname with -h (or --hostname). I can add it to this PR if you want, just tell me what value shall we use there.
https://docs.docker.com/config/containers/container-networking/#dns-services

I have played a little bit with `-h|--hostname` and it does not work for me :( I have tried with `--hostname tango` and REST does not see this host:

rest-resolve-tango-host-failure
 
Starter is named correctly though if Tango host is accessed via IP:

rest-resolve-tango-host-failure1

And another problem is that devices are no longer accessible, as they point to `localhost:10000`

rest-resolve-tango-host-failure2

So my conclusion is that we need more careful setup for Tango docker stack, to address all the issues above... 

Let's fallback for `if-fi` solution for the moment.

`tango-admin ping` seems like a good improvement to me!

mliszcz:
> I have played a little bit with -h|--hostname and it does not work for me :( I have tried with --hostname tango and REST does not see this host:

Because REST server is running on localhost (outside docker) and hostname resolution will not work in this case. `-h` will only set hostname inside docker container (and will add entry to docker's internal DNS so that different containers could talk to each other by hostname). To access container by its hostname from your machine you need to modify your system configuration with either:
* additional DNS proxy server shared by host and containers or
* hardcode container's IP address in `/etc/hosts` file

> Starter is named correctly though if Tango host is accessed via IP:

This is expected as hostname is changed in starter's container, and starter server is started with:
```
/usr/bin/Starter `hostname --short`
```

> And another problem is that devices are no longer accessible, as they point to localhost:10000 (Why?!)

This is probably configuration issue. Is correct `TANGO_HOST` set for the devices? Maybe they are reading `/etc/tangorc` or are using some default value.

> So my conclusion is that we need more careful setup for Tango docker stack, to address all the issues above...

There were some efforts in latest TangoBox release to dockerize as much as possible. If there is a demand (and proper priotity is assiged), we can implement DNS resolution and move DataBaseds to a container. I did some tests few weeks ago and it is possible to expose docker's internal DNS server to the host, but some effort is still required to make it usable.

mliszcz:
I've did some more changes in latest commit:
* Remove -p, so that another instance can use port 10000
* Speedup DB setup by disabling InnoDB
* Use tango_admin to wait until the DB goes online

Results:

Before the PR - we sleep for 30s:
```
$ time make start-tango
...
real    0m32,212s
user    0m0,217s
sys     0m0,123s
```

With --ping-database instead of sleep - 15s to start the db:
```
real    0m16,085s
user    0m0,364s
sys     0m0,206s
```

With InnoDB and timezone detection disabled - just 5 seconds to start the db:
```
real    0m5,367s
user    0m0,216s
sys     0m0,155s
```

I think we need this kind of improvement anyway if we want to re-start the DB and servers before each test.

mliszcz:
Also changed pre_test.sh to use tango_admin:
* Use tango_admin to wait until device servers are online

mliszcz:
Obsoleted by #568

mliszcz:
Closing in favor of #568

Issue #524, opened 2019-01-29
Unstable event::pipe_event testcase (assertion failed)


mliszcz:
`event::pipe_event` testcase failed in travis job (Debian 8, Release, zmq 4.2.2):
https://travis-ci.org/tango-controls/cppTango/jobs/485832345

I've restarted the job without touching the code and the issue disappeared.
I haven't investigated this yet. I don't know if this is a Tango bug or some race condition in the testcase.

```
      Start 93: event::pipe_event
93/94 Test #93: event::pipe_event .................***Exception: Other  6.19 sec
   subscribe_event --> OK
   received event --> OK
   received event (with specified date) --> OK
   received event (with error) --> OK
   received event (no copy sending) --> OK
   unsubscribe_event --> OK
pipe_event: /home/tango/src/cpp_test_suite/event/pipe_event.cpp:238: int main(int, char**): Assertion `cb.cb_executed == 2' failed.
```

bourtemb:
Hi @mliszcz,

I've spent some time trying to understand the origin of these recurrent test failures in pipe_event test.
I could reproduce the error from time to time.
The tests were failing at 2 different locations:
https://github.com/tango-controls/cppTango/blob/tango-9-lts/cpp_test_suite/event/pipe_event.cpp#L132
and
https://github.com/tango-controls/cppTango/blob/tango-9-lts/cpp_test_suite/event/pipe_event.cpp#L238

What I observed, using Wireshark was that, even though DeviceProxy::subscribe_event() previous call succeeded, the ZMQ subscription message was received later by the publisher, and sometimes. even after the execution of the PushPipeEvent command (case of the [Line132](https://github.com/tango-controls/cppTango/blob/tango-9-lts/cpp_test_suite/event/pipe_event.cpp#L132) failure). So the subscriber could not receive the events because the ZMQ subscription phase was not completed.

This reminded me https://github.com/tango-controls/cppTango/issues/399 and https://github.com/tango-controls/cppTango/pull/415 where we reduced a sleep time during the subscription command.
When we first tested it, the pipe_event test was also failing.
@sergirubio managed to find a sleep time value where the tests would not fail but it looks like the tests have been failing more often recently.
This might be due to the new Travis CI infrastructure.

Here is what the comments above those sleep() calls are saying:
> // Sleep for some mS in order to give to ZMQ some times to propagate the subscription to the publisher

and

> // A small mS sleep here! This is required in case there is a push_event in the read_attribute (or pipe)
[...]

PR #558, opened 2019-05-31
Reduce sleeping in tests

mliszcz:
Two weeks ago we discussed with @bourtemb the problem of sleeps/waits in test suites.

I'd like to trigger a discussion (in PR, just to show some code) - whether we shall focus (or not) on the problem and how can we improve our tests.

I've done a simple PoC of an event callback that maintains event queue and allows to perform assertions on received events. When applied to `dev_intr_event` testcase, execution time went down from ~11s to ~2s.

Pros:
* test suite execution time reduction -> faster CI feedback
* simpler test code (calback's state is reduced), no need to track `cb_executed`, etc.
* we can safely increase timeout value for cases where current sleep period is sometimes too low (unstable testcases) -> CI stability improvement

Cons:
* may hide some bugs when blindly applied to all sleep statements (see discussion in #524)

It's C++11 only, but I think with enough patience it can be rewritten to pthreads (lambdas will be a problem). Maybe we can start introducing C++11 first in test code (as there is little end-user impact).

Long running testcases:

- [ ] 100.09 CXX::cxx_reconnection_zmq
- [ ] 55.07 CXX::cxx_stateless_subscription
- [ ] 49.28 CXX::cxx_old_poll
- [ ] 44.21 CXX::cxx_poll_admin
- [ ] 40.12 event::multi_dev_event
- [ ] 38.66 old_tests::lock
- [ ] 38.18 CXX::cxx_fwd_att
- [ ] 32.08 asyn::asyn_write_cb
- [ ] 32.07 asyn::asyn_attr_cb
- [ ] 31.21 event::change_event
- [ ] 30.10 asyn::asyn_attr
- [ ] 30.07 asyn::asyn_write_attr
- [ ] 29.17 event::per_event
- [ ] 29.07 asyn::asyn_write_attr_multi
- [ ] 29.07 asyn::asyn_cb_cmd
- [ ] 29.07 asyn::asyn_attr_multi
- [ ] 28.26 event::att_type_event
- [ ] 27.07 asyn::asyn_cmd
- [ ] 26.12 event::archive_event
- [ ] 24.03 asyn::auto_asyn_cmd
- [ ] 23.13 event::change_event_buffer
- [ ] 23.12 event::multi_event
- [ ] 21.09 CXX::cxx_server_event
- [ ] 20.05 CXX::cxx_asyn_reconnection
- [ ] 13.06 asyn::asyn_cb2
- [ ] 13.03 asyn::asyn_cb
- [ ] 12.29 event::att_conf_event_buffer
- [ ] 12.16 old_tests::att_conf
- [ ] 11.31 old_tests::ring_depth
- [ ] 11.11 event::state_event
- [ ] 11.11 event::data_ready_event_buffer
- [ ] 11.10 event::dev_intr_event
- [ ] 11.09 event::change_event64
- [ ] 10.20 CXX::cxx_attr_misc
- [ ] 9.79 old_tests::state_attr
- [ ] 9.07 CXX::cxx_dserver_misc
- [ ] 8.07 CXX::cxx_enum_att
- [ ] 8.07 CXX::cxx_blackbox
- [ ] 8.04 CXX::cxx_signal
- [ ] 7.07 event::pipe_event
- [ ] 6.07 event::event_lock
- [ ] 6.04 CXX::cxx_attr_write
- [ ] 5.20 old_tests::rds
- [ ] 5.07 old_tests::attr_proxy
- [ ] 4.34 old_tests::attr_conf_test
- [ ] 4.05 CXX::cxx_class_dev_signal
- [ ] 4.04 CXX::cxx_class_signal
- [ ] 4.03 old_tests::sub_dev
- [ ] 3.55 old_tests::mem_att
- [ ] 3.08 CXX::cxx_misc
- [ ] 3.05 CXX::cxx_poll
- [ ] 2.07 event::user_event
- [ ] 2.06 old_tests::ConfEventBugClient
- [ ] 2.06 event::data_ready_event
- [ ] 2.03 event::att_conf_event
- [ ] 1.07 CXX::cxx_pipe_conf

### DIFF
--- a/cpp_test_suite/CMakeLists.txt
+++ b/cpp_test_suite/CMakeLists.txt
@@ -4,6 +4,8 @@ set(CMAKE_CXX_FLAGS_RELEASE "-O0 -g")
 
 configure_file(CTestCustom.cmake CTestCustom.cmake)
 
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/test_utils)
+
 #TODO different OS
 set(SERV_NAME "DevTest")
 set(INST_NAME "test")
--- a/cpp_test_suite/event/dev_intr_event.cpp
+++ b/cpp_test_suite/event/dev_intr_event.cpp
@@ -13,50 +13,13 @@
 #include <unistd.h>
 #endif
 
+#include <TangoCallbackMock.h>
+
 #define	coutv	if (verbose == true) cout
 
 using namespace Tango;
 
 bool verbose = false;
-
-class EventCallBack : public Tango::CallBack
-{
-	void push_event(Tango::DevIntrChangeEventData*);
-
-public:
-	int cb_executed;
-	int cb_err;
-	size_t nb_cmd;
-	size_t nb_att;
-	bool dev_start;
-};
-
-void EventCallBack::push_event(Tango::DevIntrChangeEventData* event_data)
-{
-	cb_executed++;
-
-	try
-	{
-		coutv << "EventCallBack::push_event(): called device " << event_data->device_name << " event " << event_data->event << "\n";
-		if (!event_data->err)
-		{
-			coutv << "Received device interface change event for device " << event_data->device_name << std::endl;
-			nb_cmd = event_data->cmd_list.size();
-			nb_att = event_data->att_list.size();
-			dev_start = event_data->dev_started;
-		}
-		else
-		{
-			coutv << "Error send to callback" << std::endl;
-			Tango::Except::print_error_stack(event_data->errors);
-		}
-	}
-	catch (...)
-	{
-		coutv << "EventCallBack::push_event(): could not extract data !\n";
-	}
-
-}
 
 int main(int argc, char **argv)
 {
@@ -90,9 +53,7 @@ int main(int argc, char **argv)
 
 	try
 	{
-		EventCallBack cb;
-		cb.cb_executed = 0;
-		cb.cb_err = 0;
+		Tango::CallbackMock<Tango::DevIntrChangeEventData> cb{};
 
 //
 // subscribe to a data ready event
@@ -104,12 +65,15 @@ int main(int argc, char **argv)
 // The callback should have been executed once
 //
 
-		assert (cb.cb_executed == 1);
-		assert (cb.dev_start == true);
+		size_t old_cmd_nb = 0;
+		size_t old_att_nb = 0;
 
-		size_t old_cmd_nb = cb.nb_cmd;
-		size_t old_att_nb = cb.nb_att;
-		cb.dev_start = false;
+		cb.assert_called(0, [&](const decltype(cb)::Event& event) {
+			assert(not event.err);
+			assert(event.dev_started);
+			old_cmd_nb = event.cmd_list.size();
+			old_att_nb = event.att_list.size();
+		});
 
 		cout << "   subscribe_event --> OK" << std::endl;
 
@@ -122,11 +86,10 @@ int main(int argc, char **argv)
 		d_in << in;
 		device->command_inout("IOAddCommand",d_in);
 
-		Tango_sleep(1);
-
-		assert (cb.cb_executed == 2);
-		assert (cb.nb_cmd == old_cmd_nb + 1);
-		assert (cb.nb_att == old_att_nb);
+		cb.assert_called(1000, [&](const decltype(cb)::Event& event) {
+			assert(event.cmd_list.size() == old_cmd_nb + 1);
+			assert(event.att_list.size() == old_att_nb);
+		});
 
 //
 // Remove the dynamic command -> Should generate a device interface change event
@@ -134,11 +97,10 @@ int main(int argc, char **argv)
 
 		device->command_inout("IORemoveCommand",d_in);
 
-		Tango_sleep(1);
-
-		assert (cb.cb_executed == 3);
-		assert (cb.nb_cmd == old_cmd_nb);
-		assert (cb.nb_att == old_att_nb);
+		cb.assert_called(1000, [&](const decltype(cb)::Event& event) {
+			assert(event.cmd_list.size() == old_cmd_nb);
+			assert(event.att_list.size() == old_att_nb);
+		});
 
 //
 // Add dynamic commands in loop -> 1 event
@@ -148,11 +110,10 @@ int main(int argc, char **argv)
 		d_in << in;
 		device->command_inout("IOAddCommand",d_in);
 
-		Tango_sleep(1);
-
-		assert (cb.cb_executed == 4);
-		assert (cb.nb_cmd == old_cmd_nb + 3);
-		assert (cb.nb_att == old_att_nb);
+		cb.assert_called(1000, [&](const decltype(cb)::Event& event) {
+			assert(event.cmd_list.size() == old_cmd_nb + 3);
+			assert(event.att_list.size() == old_att_nb);
+		});
 
 		cout << "   Dev. interface change event after add/remove dynamic command --> OK" << std::endl;
 
@@ -161,20 +122,21 @@ int main(int argc, char **argv)
 //
 
 		device->command_inout("Init");
-		Tango_sleep(1);
 
-		assert (cb.cb_executed == 4);
+		Tango_sleep(1);
+		cb.assert_not_called();
+
 		std::string adm_name = device->adm_name();
 		DeviceProxy adm(adm_name);
 
 		Tango::DeviceData d_in_restart;
 		d_in_restart << device_name;
 		adm.command_inout("DevRestart",d_in_restart);
-		Tango_sleep(1);
 
-		assert (cb.cb_executed == 5);
-		assert (cb.nb_cmd == old_cmd_nb);
-		assert (cb.nb_att == old_att_nb);
+		cb.assert_called(1000, [&](const decltype(cb)::Event& event) {
+			assert(event.cmd_list.size() == old_cmd_nb);
+			assert(event.att_list.size() == old_att_nb);
+		});
 
 		cout << "   Dev. interface change event after Init or DevRestart command --> OK" << std::endl;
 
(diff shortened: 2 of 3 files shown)


Ingvord:
It will be also good to re-write non-cxx test (like this one) to cxx if we touch it.

mliszcz:
Thanks @Ingvord. I suggest to have separate issue for refactoring the tests (and a dedicated discussion if we want to go with CxxTest or something else), as in this PR I'd like to only touch `Tango_sleep(x)` calls.

Below is a list of testcases that need more than 1s to complete (I assume they have some `sleep`s inside). List is sorted by execution time:

- [ ] 100.09 CXX::cxx_reconnection_zmq
- [ ] 55.07 CXX::cxx_stateless_subscription
- [ ] 49.28 CXX::cxx_old_poll
- [ ] 44.21 CXX::cxx_poll_admin
- [ ] 40.12 event::multi_dev_event
- [ ] 38.66 old_tests::lock
- [ ] 38.18 CXX::cxx_fwd_att
- [ ] 32.08 asyn::asyn_write_cb
- [ ] 32.07 asyn::asyn_attr_cb
- [ ] 31.21 event::change_event
- [ ] 30.10 asyn::asyn_attr
- [ ] 30.07 asyn::asyn_write_attr
- [ ] 29.17 event::per_event
- [ ] 29.07 asyn::asyn_write_attr_multi
- [ ] 29.07 asyn::asyn_cb_cmd
- [ ] 29.07 asyn::asyn_attr_multi
- [ ] 28.26 event::att_type_event
- [ ] 27.07 asyn::asyn_cmd
- [ ] 26.12 event::archive_event
- [ ] 24.03 asyn::auto_asyn_cmd
- [ ] 23.13 event::change_event_buffer
- [ ] 23.12 event::multi_event
- [ ] 21.09 CXX::cxx_server_event
- [ ] 20.05 CXX::cxx_asyn_reconnection
- [ ] 13.06 asyn::asyn_cb2
- [ ] 13.03 asyn::asyn_cb
- [ ] 12.29 event::att_conf_event_buffer
- [ ] 12.16 old_tests::att_conf
- [ ] 11.31 old_tests::ring_depth
- [ ] 11.11 event::state_event
- [ ] 11.11 event::data_ready_event_buffer
- [ ] 11.10 event::dev_intr_event
- [ ] 11.09 event::change_event64
- [ ] 10.20 CXX::cxx_attr_misc
- [ ] 9.79 old_tests::state_attr
- [ ] 9.07 CXX::cxx_dserver_misc
- [ ] 8.07 CXX::cxx_enum_att
- [ ] 8.07 CXX::cxx_blackbox
- [ ] 8.04 CXX::cxx_signal
- [ ] 7.07 event::pipe_event
- [ ] 6.07 event::event_lock
- [ ] 6.04 CXX::cxx_attr_write
- [ ] 5.20 old_tests::rds
- [ ] 5.07 old_tests::attr_proxy
- [ ] 4.34 old_tests::attr_conf_test
- [ ] 4.05 CXX::cxx_class_dev_signal
- [ ] 4.04 CXX::cxx_class_signal
- [ ] 4.03 old_tests::sub_dev
- [ ] 3.55 old_tests::mem_att
- [ ] 3.08 CXX::cxx_misc
- [ ] 3.05 CXX::cxx_poll
- [ ] 2.07 event::user_event
- [ ] 2.06 old_tests::ConfEventBugClient
- [ ] 2.06 event::data_ready_event
- [ ] 2.03 event::att_conf_event
- [ ] 1.07 CXX::cxx_pipe_conf

bourtemb:
@mliszcz could you please solve the conflict?

bourtemb:
> @bourtemb I've resolved the conflicts as requested.

Thanks!

> In this PR I'd like to dicsuss if we see any benefit in this approach.

Well, one clear benefit is the test execution time!

> Applying this change to all test-cases seems rather simple

This is good news!
Then I would personally be in favour of applying this whenever possible.

mliszcz:
@Ingvord, @t-b what is your opinion on this PR? If you agree that it is useful, I propose following way forward:
1. We review and merge this PR
2. I'll open an issue with a list of long-running testcases
3. We start fixing tests one-by-one
4. We close the issue when no (or almost no) `sleep`s are left

Ingvord:
@mliszcz do you plan to disband PR #524 in favor of this one i.e. redo #524 using the pattern from this PR?

mliszcz:
@Ingvord I think this PR won't solve #524. All this PR allows for is:
* replace `sleep(X)` with `sleep-until-event-received-up-to(X)`
* make callback stateless (except the event queue) - no need to update and check `cb_executed`

If I recall, the problem with #524 is that we are waiting 1 second for an event which arrives later. However we can change the testcase so that it waits for this event e.g. up to 5s. This should reduce failure rate. This is very similar to what you did in #589.

Ingvord:
yeah, so we can drop #589 and put 5s here.

Sounds good enough to me.

bourtemb:
> If I recall, the problem with #524 is that we are waiting 1 second for an event which arrives later. However we can change the testcase so that it waits for this event e.g. up to 5s. This should reduce failure rate. This is very similar to what you did in #589.

No, the problem in #524 is that we missed the event. We will not get it, even if we wait for a longer time.
#524 has to be fixed in the library. We cannot do much on the tests side and we should rather focus on having subscribe_event() more reliable.

Ingvord:
> we should rather focus on having subscribe_event() more reliable.

Reliability really depends on if:

- 1) we want *hot* event subscription, i.e. when server continuously emits events and subscribers come and go. In this case subscriber MUST be prepared to miss events
- 2) we want *cold* event subscription i.e. server waits for a subscriber(s) before emitting any event. It MAY then cache last event and resend it to every new subscriber

These are totally different design approaches. The problem is that existing implementation tries to solve both in one shot.

This is of course out of this PR scope and even cppTango i.e. something for V10 as it requires new handshake mechanism

t-b:
I like the direction where this PR is going!

But I'm struggling a bit with the implementation details. We don't check the number of calls anymore. Maybe `assert(cb.get_call_count() == 1);` is just missing.

Also is CallbackMock designed to also handle multiple events? It looks like not, which is fine if it is on purpose.

But I think we can easily extend it to handle both issues. To something like the following maybe?

```cpp
#include <tango.h>

#include <chrono>
#include <condition_variable>
#include <deque>
#include <mutex>

using namespace std::literals::chrono_literals;

namespace Tango
{

template <typename TEvent>
class CallbackMock : public Tango::CallBack
{
public:

    using Event = TEvent;

    void push_event(TEvent* event)
    {
        {
            std::unique_lock<std::mutex> lk(m);
            events.emplace_back(std::shared_ptr<Event>(new Event(*event)));
        }

        cv.notify_one();
    }

    template <typename Duration>
    void assert_not_called(Duration timeout)
    {
        assert_called(timeout, 0u, [](const Event&){});
    }

    template <typename Duration, typename Function>
    void assert_called(Duration timeout, size_t num_expected_events, Function&& func)
    {
        std::unique_lock<std::mutex> lk(m);

        if (num_expected_events > 0u && num_expected_events == events.size())
        {
            apply(std::forward<Function>(func));
        }
        else
        {
            if (cv.wait_for(
                lk,
                timeout,
                [this, num_expected_events]{return events.size() == num_expected_events && !events.empty(); }))
            {
                apply(std::forward<Function>(func));
                assert(events.size() == num_expected_events);
            }
            else
            {
                assert(num_expected_events == 0);
                assert(events.empty());
            }
        }
        events.clear();
    }

private:

    template <typename Func>
    void apply(Func&& func)
    {
      for(auto event: events)
      {
        func(*event);
      }
    }

    std::deque<std::shared_ptr<TEvent>> events;
    std::mutex m;
    std::condition_variable cv{};
};

}
```

And that can now be used like

```
		cb.assert_called(0ms, 1, [&](const decltype(cb)::Event& event) {
			assert(not event.err);
			assert(event.dev_started);
			old_cmd_nb = event.cmd_list.size();
			old_att_nb = event.att_list.size();
		});
```

and if you don't expect events you can just do 

```
	cb.assert_not_called(1s);
```

and it will also do the right thing.jkk

mliszcz:
@t-b thanks for checking the PR and for your proposal

My initial assuumption was that checking for multiple events should be done at the call site, because events could be different (if the same payload is expected, just wrap it in a loop), e.g.:

```cpp
// assert that the callback is called exactly two times
cb.assert_called(0, verify_first_event);
cb.assert_called(0, verify_second_event);
cb.assert_not_called(0);
```

I have some comments for the changes you propose:
* `apply` member function hides the name from `std` namespace, I'd use a different name
* what is the reason for using `shared_ptr` to store data which is never shared and the ownership is is clear?
* it seems to me with your proposal it only possible to verify all events, i.e. following scenario is not supported:
  * 20 events are buffered in the callback
  * I expect 10 first events to match verify_a
  * and next 10 events to match verify_b
* the condition looks dangerous to me, do you know following scenario is possible?
  * ```cpp
    if (cv.wait_for(
    lk,
    timeout,
    [this, num_expected_events]{return events.size() == num_expected_events && !events.empty(); }))
    ```
  * testcase is waiting on the condition for one event to arrive
  * two events arrive in a short interval
  * first event takes the lock, queues event, notifies cv and releases the lock
  * testcase thread is notified (queue size == 1), it wakes and tries to re-acquire the lock
  * the CPU schedules another thread (second event) which takes the lock before testcase thread - is it possible? (probably, see https://stackoverflow.com/a/58139135)
  * condition will be never fulfilled, because now there are two events in the queue

t-b:
@mliszcz So I finally came back to
>    apply member function hides the name from std namespace, I'd use a different name

Sure!

>   what is the reason for using shared_ptr to store data which is never shared and the ownership is is clear?

The range for loop requires that currently. But I think using `const auto &` this should work with unique_ptr as well.

>    it seems to me with your proposal it only possible to verify all events, i.e. following scenario is not > supported:
>        20 events are buffered in the callback
>        I expect 10 first events to match verify_a
>       and next 10 events to match verify_b

Yes currently there is only support for one verification function. But if handling multiple events at 
the  call site works, we don't need my version.

>    the condition looks dangerous to me, do you know following scenario is possible?
 [...]

I don't really see why it sould fail like you described. If you are only expecting one event and you are getting two during the `timeout` this is treated correctly as an error right now.

But as I said if we can/should handle multiple events at the call site this makes the code much easier.

PR #562, opened 2019-06-05
Add debian buster to travis

t-b:


.travis/debian10/run.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
mliszcz: Can we re-use this script also for debian 7 and debian 9? Just move i to `.travis/run.sh`.

.travis/debian10/run.sh
```diff
@@ -0,0 +1,18 @@
...
+docker exec cpp_tango cmake            \
+  -H/home/tango/src                    \
+  -B/home/tango/src/build              \
+  -DCMAKE_VERBOSE_MAKEFILE=true        \
+  -DCPPZMQ_BASE=/home/tango            \
+  -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE
+
+if [ $? -ne "0" ]
+then
+    exit -1
```
mliszcz: Consider just adding `set -e` to the top of the file.

.travis/debian10/run.sh
Ingvord: @mliszcz , there are some diverses related to including coveralls and zmq_base. Apart from that we can think about re-use of course!

mliszcz: Yes, but only debian 8 setup is different. For debian 7, 9 and 10 we have exactly the same file (except formatting, newlines, etc.)

Ingvord: true

Ingvord: we can pass target system as a variable into the script or, in case debian 8, use dedicated script file. I have no strong opinion on the approach

mliszcz: I haven't noticed that we have a common call to the `run.sh` script:
```yaml
script:
  - .travis/${OS_TYPE}/run.sh
```

In this case, maybe we can just create soft links with relative paths like:
```bash
debian10/run.sh # actual file
debian9/run.sh -> ../debian10/run.sh
debian7/run.sh -> ../debian10/run.sh
```

Ingvord: Well, that is unnecessary complication to my taste due to misleading unclear call chain. It even reads strange: `debian 7 -> debian 10` 

I would rather stay with variable passed into script or test `OS_TYPE` in travis and make a call accordingly, something like (pseudo code):

```
script:
  - IF ${OS_TYPE} == debian8 THEN ./debian8/run.sh ELSE ./run.sh

```

Ingvord:
We should remove two other jobs then, as travis supports only 5 parallel jobs. 

debian7 and debian 8 RELEASE - fast approach

better would be debian 7 and debian 8 DEBUG. But in this case we need to migrate COVERALLS and Sonar to debian 9 DEBUG

Another option: debian 8 RELEASE and debian 10 RELEASE

bourtemb:
> 
> 
> We should remove two other jobs then, as travis supports only 5 parallel jobs.

I would keep at least 1 build for debian7 and debian8 for the moment because I guess some people are still using it in the community.
I would remove Debian 8 release build to keep the coveralls and sonar config as it is now in Debian 8 debug build.
I have no strong opinion on which other one to remove among Debian 9 and Debian 10 builds.
I guess Debian 9 is currently more used than Debian 10 in the Tango community.
On the other hand, Debian Buster (10) is the current Debian stable release.

.travis/debian10/run.sh
t-b: Actually we can reuse all of the code and handle the differences not OS specfic but feature flag specific. I've pushed something which goes along that line.

@mliszcz Thanks for the commit squashing suggestions. I've used these and also squashed  one more.

.travis/debian10/run.sh
t-b: Done.

t-b:
> We should remove two other jobs then, as travis supports only 5 parallel jobs.
@Ingvord According to https://docs.travis-ci.com/user/build-matrix/ it is nowadays 200 jobs, but I still reduced them.

@bourtemb I've removed all release builds from debian < 10, now only debian 10 has release and debug builds.

t-b:
Ready for review. The appveyor failure is bogus.

Ingvord:
> There are some strange messages at the end of Travis Debian 8 build with SONAR_SCANNER=ON :
.travis/sonar.sh: line 11: sonar-scanner: command not found

Looks like Travis supports SonarCloud addon only on `dist: trusty`. We need to change dist for Debian 8 or install manually [1](https://community.sonarsource.com/t/how-to-install-sonar-scanner-in-travis-ci/9976)

In fact, I would move SONER_CANNER=ON and COVERALLS to Debian 10 configuration, definitely in a new PR

PR #564, opened 2019-06-12
Use map for attribute lookup in MultiAttribute::check_associated

### LINKED ISSUE #556: Poor performance when removing dynamic attributes
I have a server which allows for adding and removing dynamic attributes:
* `CreateDynamicAttributes(500, 100)` - create 500x double spectrum with 100 elements
* `ClearDynamicAttributes` - remove all dynamic attributes

Removing 500 attributes takes more than a minute. Attribute creation is relatively fast.

I have had a look into the code. The problem is in `MultiAttribute::remove_attribute` which does `MultiAttribute::check_associated` for all attributes. The overall cost of a single `remove_attribute` call is thus proportional to N^2 string comparisons (N is the number of attributes). With 1000 attributes configured, removing a single attribute takes ~1 second.

Improvement idea: can we use `Attribute*` instead of `string` determine associated attribute?

Callgrind profile:
obraz

The helper code was generated by Pogo.
The server is located [here](https://github.com/tango-controls/sys-tango-benchmark/tree/feat-12-dynamic-attributes-vs-memory/ds/CppBenchmarkTarget)

mliszcz:
To slightly improve remove_attribute performance when multiple writable attributes are configured (fixes #556).

### DIFF
--- a/cppapi/server/multiattribute.cpp
+++ b/cppapi/server/multiattribute.cpp
@@ -580,7 +580,7 @@ void MultiAttribute::check_associated(long index, std::string &dev_name)
 		Except::throw_exception(
 			API_AttrOptProp,
 			o.str().c_str(),
-			"MultiAttribute::MultiAttribute");
+			"MultiAttribute::check_associated");
 	}
 
 	Attribute& assoc_attribute = *attr_list[assoc_index];
@@ -596,7 +596,7 @@ void MultiAttribute::check_associated(long index, std::string &dev_name)
 		Except::throw_exception(
 			API_AttrOptProp,
 			o.str().c_str(),
-			"MultiAttribute::MultiAttribute");
+			"MultiAttribute::check_associated");
 	}
 
 	if (attribute.get_data_type() != assoc_attribute.get_data_type())
@@ -609,7 +609,7 @@ void MultiAttribute::check_associated(long index, std::string &dev_name)
 		Except::throw_exception(
 			API_AttrOptProp,
 			o.str().c_str(),
-			"MultiAttribute::MultiAttribute");
+			"MultiAttribute::check_associated");
 	}
 
 	attribute.set_assoc_ind(assoc_index);
--- a/cppapi/server/multiattribute.cpp
+++ b/cppapi/server/multiattribute.cpp
@@ -553,52 +553,66 @@ void MultiAttribute::add_user_default(std::vector<AttrProperty> &prop_list,std::
 //
 //------------------------------------------------------------------------------------------------------------------
 
-void MultiAttribute::check_associated(long index,std::string &dev_name)
+void MultiAttribute::check_associated(long index, std::string &dev_name)
 {
-	if ((attr_list[index]->get_writable() == Tango::READ_WITH_WRITE) ||
-	    (attr_list[index]->get_writable() == Tango::READ_WRITE))
+	Attribute& attribute = *attr_list[index];
+	const AttrWriteType write_type = attribute.get_writable();
+
+	if (write_type != Tango::READ_WITH_WRITE && write_type != Tango::READ_WRITE)
 	{
-		unsigned long j;
-		std::string &assoc_name = attr_list[index]->get_assoc_name();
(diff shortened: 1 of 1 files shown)


t-b (dismissed):
Looks good.

Two minor comments:
- If we can rely on `get_attr_ind_by_name` doing the transform->lower we can ditch it here.
- I'm a big fan of reducing indentation and returning early. So I would write it as 

```cpp
if (write_type != Tango::READ_WITH_WRITE && write_type != Tango::READ_WRITE)
  {
    return;
  }

```
but that is just taste I guess.

mliszcz:
Thanks @t-b. I've changed the PR as you suggested. @bourtemb could you please check as well?

cppapi/server/multiattribute.cpp
```diff
@@ -553,52 +553,66 @@ void MultiAttribute::add_user_default(std::vector<AttrProperty> &prop_list,std::
...
+	{
+		TangoSys_OMemStream o;
+		o << "Device --> " << dev_name;
+		o << "\nProperty writable_attr_name for attribute " << attribute.get_name();
+		o << " is set to " << assoc_name;
+		o << ", but this attribute does not exist" << std::ends;
+		Except::throw_exception(
+			API_AttrOptProp,
+			o.str().c_str(),
+			"MultiAttribute::MultiAttribute");
```
bourtemb: I know it was already like that before, but I think the origin field of the exception should be set to MultiAttribute::check_associated instead of MultiAttribute::MultiAttribute.
Same comment also applies for lines 599 and 612

bourtemb (commented):
Thanks @mliszcz 
Just a comment on the origin field in the exceptions thrown, but this was like that before this PR.
I think we could change that.
Except that, this PR looks good to me!

cppapi/server/multiattribute.cpp
mliszcz: Thanks for this comment. I changed all three names. How about using some macro to figure out current function's name and applying it to all Except::throw_exception?

I think we can steal this one: https://www.boost.org/doc/libs/1_58_0/boost/current_function.hpp

Also, we modify WARN_STREAM, cout5 and others to automatically insert function name. We often need to spell it explicitly:
```cpp
cout1 << "Attribute::Attribute(): event_period_str " << event_period_str << " event_period = " << event_period << std::endl;
```

bourtemb: > Thanks for this comment. I changed all three names. How about using some macro to figure out current function's name and applying it to all Except::throw_exception?

This is something which would be very useful!
@ingvord created a PR for this a while ago: https://github.com/tango-controls/cppTango/pull/366
Please feel free to contribute on this PR.

> 
> I think we can steal this one: https://www.boost.org/doc/libs/1_58_0/boost/current_function.hpp
> 
> Also, we modify WARN_STREAM, cout5 and others to automatically insert function name. We often need to spell it explicitly:
> 
> ```c++
> cout1 << "Attribute::Attribute(): event_period_str " << event_period_str << " event_period = " << event_period << std::endl;
> ```

bourtemb (approved):

t-b (approved):

t-b:
@mliszcz Will you do a backports PR?

mliszcz:
@t-b sure, I can backport if you want. I don't know what is the policy for backports - do we backport everything that can be backported or just critical bugfixes? Can I assume that I should backport everything that is tagged with "Candidate for backport" maybe?

t-b:
@mliszcz Well I thought you added the "Candidate for backport" tag, but that was me. From my understanding "Candidate for backport" means that this should get backported. But maybe we need a different name then. @bourtemb Thoughts?

bourtemb:
> From my understanding "Candidate for backport" means that this should get backported.

I think we can take this convention indeed but we might need to discuss before to put this tag on a PR to decide whether it is worth being backported.

In the case of this PR, I would say if it is easy to backport, we can do it but it does not look like a critical bug fix to me, so I would personally not put a high priority on this one.

mliszcz:
I'll send a backport PR. It was waiting for a review for more than a year without conflicts. I think I just need to cherry-pick relevant commits and do not expect any conflicts in 9.3-backports.

PR #572, opened 2019-07-20
Fix #496 blind event clients after dev restart

### LINKED ISSUE #496: Client subscribing to events can be blind for up to 200 seconds if the target device is restarted (device restart)
A client subscribing to events on an attribute does not receive events during a period of time which can be as long as 200 seconds in case the target device is restarted  (device restart, not process restart).
No error is reported on the client side so a GUI will continue to display the old attribute value, which might be wrong.
The client will start receiving again events after the device restart when it will send an event resubscription command. These event resubscription commands are sent by every client every 200 seconds.
If there are several clients subscribing to the same event, as soon as one client has sent the resubscription command, the other clients will receive again the events.
This behaviour has been observed with Tango 9.2.5 (client and server side) too.

mliszcz:
To be updated

### DIFF
--- a/cppapi/server/attribute.cpp
+++ b/cppapi/server/attribute.cpp
@@ -3802,7 +3802,7 @@ void Attribute::fire_change_event(DevFailed *except)
 // Get client lib and if it's possible to send event (ZMQ socket created)
 //
 
-        vector<int> client_libs;
+        EventClientLibVersions client_libs;
         {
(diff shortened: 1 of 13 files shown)


mliszcz:
Closing in favour of #573

t-b:
Review was erroneously done for this PR and can be found at https://github.com/tango-controls/cppTango/pull/573#issuecomment-564184377.

PR #574, opened 2019-07-23
Fix 566 const reference parameters

### LINKED ISSUE #566: Use const parameters for operator<<(ostream &, ThisObjectShouldBeConst)
Hi,

We should add the const qualifier for the 2nd parameter of the ostream operator<<(ostream &, ...) functions.
For instance:
`ostream &operator<<(ostream &o_str, AttributeInfoEx &p)`
should be:
`ostream &operator<<(ostream &o_str, const AttributeInfoEx &p)`

There are many other cases to be fixed.
Thanks @chedburgh for the bug report.

mliszcz:
Following chages are included:
* pass object to print by `const&` in all `operator<<(std::ostream&, ...)`
* mark some query methods as `const`
* mark `DeviceAttribute::ext::ext_state` as `mutable` (it is updated from e.g. `Attribute::is_empty`

Fixes #566


t-b (changes requested):
Constifying is nice indeed.

One question though why do we need the mutable ext_state?

I've compiled it here with

```
diff --git a/cppapi/client/DeviceAttribute.h b/cppapi/client/DeviceAttribute.h
index 3c02d1d6..1b49abeb 100644
--- a/cppapi/client/DeviceAttribute.h
+++ b/cppapi/client/DeviceAttribute.h
@@ -1321,7 +1321,7 @@ protected :
         DeviceAttributeExt() {};
         DeviceAttributeExt & operator=(const DeviceAttributeExt &);
 
-        mutable bitset<numFlags> ext_state;
+        bitset<numFlags> ext_state;
 
         void deep_copy(const DeviceAttributeExt &);
     };
```

and that worked as well.

It would also be better to have a more verbose commit message. This is currently much shorter than the merge request description or the issue description. For a project like tango we can not just assume that the platform lives longer than the project ;) Nobody from us wants to crawl old threads on sourceforge nowadays I guess..

Ingvord:
@t-b 

> It would also be better to have a more verbose commit message. This is currently much shorter than the merge request description or the issue description. For a project like tango we can not just assume that the platform lives longer than the project ;) Nobody from us wants to crawl old threads on sourceforge nowadays I guess..

We used to squash and merge PRs. When you do it you can provide verbose commit message. I remember we already had that discussion (merge VS squash). I am still confident that master/tango-9-lts branches should not be polluted with intermediate commit messages, rather there must be small number of squashed commits linked to issues/PRs. I agree that we may pay more attention to squashed messages

mliszcz:
@t-b good catch! I must have put the `mutable` modifier accidentally or undo some chagne that required it. Indeed it compiles without mutable for both raw and unique_ptr variants of `ext` member. I'll remove the keyword.

@t-b, @Ingvord, I don't know what else I could put in the commit message. The PR is for constifying arguments to the `operator<<`. This is already stated in the commit message. What else would you like to see there?

t-b:
> @t-b, @Ingvord, I don't know what else I could put in the commit message. The PR is for constifying 
> arguments to the operator<<. This is already stated in the commit message. What else would you like > to see there?

I'm reviewing the PR with `git log --reverse -p master..` ( I use an alias which does some more stuff actually) and there I get

```
commit a6f86e21928cdf2b7980fe4d4c14b67f33581b8e (HEAD -> fix-566-const-reference-parameters, mliszcz/fix-566-const-reference-parameters)
Author: mliszcz <liszcz.michal@gmail.com>
Date:   Mon Jul 22 12:39:09 2019 +0200

    Use const& arguments in operator<<(ostream&, ...) (#566)
```

which tells me *what* you have done but not *why*. The PR message is a bit more explanative but also lacks the why. This is a bit bikeshedding in this case here but still we should strive for a code base where you can read the commit messages and understand the motiviations for the changes.

mliszcz:
> The PR message is a bit more explanative but also lacks the why. This is a bit bikeshedding in this case here

Exactly. I cannot find any explanation (the *why* part) that will not be trivial or obvious. It's not the kind of change that may (should?) be surprisig to the readers (like *change X ... because when Y also happens the Z will crash*).

Some possible explanations:
* because it is a good practice
* because printing should not modify the object
* so that we can print rvalues

But I don't think we should use commit messages to teach people the basic programming principles.
Let me know if you still want it to be changed.

t-b:
@mliszcz Okay let's leave the commit message as it is. If you remove the mutable I'll approve ;)

mliszcz:
@t-b thanks! Mutable is removed.

t-b (dismissed):

t-b:
Do we track somewhere that we have to raise the ABI version of the library? According to the link from https://github.com/tango-controls/cppTango/wiki/Policies-Binary-Compatibility-Issues we need to do that here due to

```
change its signature. This includes:

    changing any of the types of the arguments in the parameter list, including changing the const/volatile qualifiers of the existing parameters (instead, add a new method)
```

bourtemb:
> Do we track somewhere that we have to raise the ABI version of the library? 

Indeed, we need to raise the version number in the CMakeLists,txt at least.
The CHANGELOG will have to be updated as well at some point.

mliszcz:
@t-b, as for binary compatiblity, this was already pointed out by @bourtemb in #566. In general, next release will not be binary compatible with 9.3.3.

IMO a general comment that there is no binary compatibility should be added to the release description (https://github.com/tango-controls/cppTango/releases)

mliszcz:
@bourtemb could you check the PR?

t-b (approved):

t-b:
@bourtemb And this one.

bourtemb (approved):
Looks good to me!
Still a lot to be done to handle constness correctly everywhere in cppTango code but it's a good start.
Thanks @mliszcz for this work and @t-b for reviewing too.

PR #575, opened 2019-07-23
Respect TANGO_LOG_PATH set in rc files (#500)

### LINKED ISSUE #500: TANGO_LOG_PATH in rc file has no effect
One can change the TANGO_LOG_PATH using the environment but not using the rc file e.g. `/etc/tangorc`. More details in [this](http://www.tango-controls.org/community/forum/c/general/installation/tango-ds-default-logging-target/) forum thread.

mliszcz:
Use standard (environment, then user rc file, then system rc file) to determine TANGO_LOG_PATH.

Fixes #500.

### DIFF
--- a/cppapi/server/logging.cpp
+++ b/cppapi/server/logging.cpp
@@ -114,8 +114,8 @@ void Logging::init (const std::string& ds_name, // dserver name
     // set default <rollover> threshold for RollingFileAppender
     Logging::_rft = kDefaultRollingThreshold;
     // try to get logging path from kTangoLogPathVar env. var.
-    char *ftg_path = ::getenv(kTangoLogPathVar);
-    if (ftg_path == 0 || ::strlen(ftg_path) == 0) {
+    const int log_path_lookup_status = ApiUtil::get_env_var(kTangoLogPathVar, Logging::_log_path);
+    if (log_path_lookup_status != 0 || Logging::_log_path.empty()) {
       // log path not set or empty, use default path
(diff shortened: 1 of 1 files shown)


cppapi/server/logging.cpp
```diff
@@ -114,8 +114,8 @@ void Logging::init (const std::string& ds_name, // dserver name
     // set default <rollover> threshold for RollingFileAppender
     Logging::_rft = kDefaultRollingThreshold;
     // try to get logging path from kTangoLogPathVar env. var.
-    char *ftg_path = ::getenv(kTangoLogPathVar);
-    if (ftg_path == 0 || ::strlen(ftg_path) == 0) {
+    ApiUtil::instance()->get_env_var(kTangoLogPathVar, Logging::_log_path);
+    if (Logging::_log_path.empty()) {
```
t-b: get_env_var of ApiUtil is static, see https://github.com/tango-controls/cppTango/blob/4feffd7c8e24b51c9597a40b9ef9982dd6e99cdf/cppapi/client/ApiUtil.h#L84.

And other callers also check the return value of get_env_var so maybe something like

```
int ret = ApiUtil::get_env_var(kTangoLogPathVar, Logging::_log_path);
if(ret != 0 || Logging::_log_path.empty())
{
...
```

mliszcz: Thanks! Fixed as proposed.

mliszcz:
@bourtemb could you please check the PR?

PR #576, opened 2019-07-23
Warn about possible BAD_INV_ORDER_ORBHasShutdown (#449)

### LINKED ISSUE #449: BAD_INV_ORDER_ORBHasShutdown when accessing an external proxy from a device server
It can easily be reproduced using the following pytango server:

```python
import tango.server

proxy = tango.DeviceProxy('sys/database/2')


class Test(tango.server.Device):

    @tango.server.attribute(dtype=str)
    def database_status(self):
        return proxy.status()


if __name__ == '__main__':
    Test.run_server()
```

Exception:

```
DevFailed: DevFailed[
DevError[
    desc = BAD_INV_ORDER CORBA system exception: BAD_INV_ORDER_ORBHasShutdown
  origin = DeviceProxy::status()
  reason = API_CorbaException
severity = ERR]

DevError[
    desc = Failed to execute status() on device (CORBA exception)
  origin = DeviceProxy::status()
  reason = API_CommunicationFailed
severity = ERR]

DevError[
    desc = Failed to read_attribute on device test/nodb/test, attribute database_status
  origin = DeviceProxy::read_attribute()
  reason = API_AttributeFailed
severity = ERR]
]
```

mliszcz:
Fixes #449.
Sorry for the flag but logging is not initialized yet when the detection takes place.

### DIFF
--- a/cppapi/server/utils.cpp
+++ b/cppapi/server/utils.cpp
@@ -255,8 +255,10 @@ void Util::effective_job(int argc,char *argv[])
 
 		ApiUtil *au = Tango::ApiUtil::instance();
 		CORBA::ORB_ptr orb_clnt  = au->get_orb();
+		bool log_client_orb_deleted = false;
 		if (CORBA::is_nil(orb_clnt) == false)
 		{
+			log_client_orb_deleted = true;
 			orb_clnt->destroy();
 			CORBA::release(orb_clnt);
 			au->set_orb(CORBA::ORB::_nil());
@@ -417,6 +419,10 @@ void Util::effective_job(int argc,char *argv[])
 		Logging::init(ds_name, (int)_tracelevel,  ((!_FileDb) && _UseDb), *db, this);
 #endif
 
+		if (log_client_orb_deleted)
+		{
+			cout1 << "Client ORB was initialized in before the server ORB, all proxies must be invalidated" << std::endl;
+		}
 		cout4 << "Connected to database" << endl;
 		if (get_db_cache() == NULL)
 			cout4 << "DbServerCache unavailable, will call db..." << endl;


t-b:
@mliszcz Stupid question: Why can't you just output the log message when you set the newly introduced boolean variable to true?

mliszcz:
@t-b because logging is not initialized yet at that point. Any attempt to access `coutX` will crash due to segfault.

The whole point of this issue/PR is questionable to me, as it is developer error. But for some reason it was escalated to Tango 9 "To Do." Maybe we should just update tango-doc instead. @bourtemb can you comment?

t-b:
@mliszcz Got it. Thanks.

bourtemb:
> The whole point of this issue/PR is questionable to me, as it is developer error. But for some reason it was escalated to Tango 9 "To Do." Maybe we should just update tango-doc instead. @bourtemb can you comment?

@mliszcz, it is a developer error, you're right. But since this is not well documented, we cannot really consider this as a developer error.
It would be useful to document this pitfall in tango-doc indeed.
The warning message should not hurt in any case.

t-b (approved):
LGTM

bourtemb:
I tested this PR with a device server having the following lines in the main before the device server initialization:

```
Tango::Database db;
cout << db.get_info() << endl;

```

During my tests, I had to start the device server with -v5 option to be able to see this INFO log message on the console.
I guess many users will miss it.
Maybe we should display the warning on std::cerr? It would be visible only from the console but whatever the logging chosen logging level (and from astor - right-click server -> Standard Error).
I'm not sure what would be the most appropriate for this message...
I guess what you tried was to have it as an INFO logging messages similar to the log messages generated by the device server programmer code. I think this would be good if you can achieve this.

mliszcz:
@bourtemb 
The log was is not visible without -v5, because:
* `get_core_logger()->get_level()` is not affected by `-v` (see *note below)
* console appender is not created unless you specify `-v5` or add it manually
  https://github.com/tango-controls/cppTango/blob/617a5b27abdbdcf2825c025ac6f3d9fad906f03e/cppapi/server/logging.cpp#L222-L226

*In fact there is some special handling only for -v5 (or any larger value):
https://github.com/tango-controls/cppTango/blob/617a5b27abdbdcf2825c025ac6f3d9fad906f03e/cppapi/server/logging.cpp#L144-L154

So I see following options:
* change logging to always create console appender for core logger
  and log "orb shutdown" with e.g. warn or error severity
* print to stderr as you proposed
* print to stdout as `_VERBOSE` is doing (but skip level check)

What is your opinion?

BTW it would be nice if `-v` could impact `get_core_logger()->get_level()`, e.g. `-vN` could set level to

```cpp
get_core_logger->set_level((log4Tango::Level)(100*(N+1)))
```

andygotz:
@mliszcz can you also change the behaviour so that -v automatically goes to the highest level i.e. 5. I think this used to be the default behaviour but this is not the case anymore.

Thanks

mliszcz:
Thanks Andy! If I understand you correctly, `-v` could work just like `-v5`. But this will result in all debug messages being printed and it will be easy to miss "bad inv order" warning. Anyway, user will still have to pass the `-v` flag on the commandline.

PR #577, opened 2019-07-23
Move Group::enable/disable to doxygen public section (#440)

### LINKED ISSUE #440: Shall we document Group::enable(device_name,fwd) and Group::disable(device_name,fwd) features and move them away from the Doxygen privatesection?
A colleague recently discovered Group::disable(device_name,fwd), Group::enable(device_name,fwd) and Group::is_enabled(device_name,fwd) methods, which are public but in the Doxygen privatesection, so they are not documented.

This feature can be quite useful for a user when it has to send commands, write/read attributes to many different devices. It allows to disable/enable some of the members of the groups. This could be useful if some devices of the group are temporarily unavailable or in maintenance mode.

I create this issue to trigger a discussion. 
Tango users are not supposed to use the methods which are not documented, even though they are declared as public methods in the class, because their interface could be changed or the methods could also be removed without notice.
So my question to @taurel and @nleclercq is: do we have good reasons to hide this feature to the users?

If not, we could move it away from the privatesection and advertise it in Doxygen and in the Tango [readthedocs documentation related to groups](http://tango-controls.readthedocs.io/en/latest/development/client-api/cpp-client-programmers-guide.html?highlight=Tango%3A%3AGroup#group).

mliszcz:
Fixes #440.
A separate PR will come for tango-doc

### DIFF
--- a/cppapi/client/group.h
+++ b/cppapi/client/group.h
@@ -157,11 +157,15 @@ public:
               bool group_element_enabled = true);
   //- dtor
   virtual ~GroupReply ();
-  //- group_element_enabled accessor
+///@publicsection
+/**
+ * Check if the group element corresponding to this reply is enabled.
+ *
+ * @return true if corresponding element is enabled, false otherwise
+ */
   inline bool group_element_enabled () const {
     return group_element_enabled_m;
   }
-///@publicsection
    //- enable/disable exception - returns the previous mode
 /**
  * Enable/Disable exception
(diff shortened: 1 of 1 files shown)


mliszcz:
@bourtemb could you check if it is ok now?

cppapi/client/group.h
```diff
@@ -1044,15 +1044,33 @@ class Group : public GroupElement
   std::vector<std::string> get_device_list (bool fwd = true);
   //-
 
+/**
+ * Enables group element
+ *
+ * Enables group element. The element will participate in all group operations.
+ *
+ * @param [in] device_name name of the element, can contain wildcards (*).
+ * If more than one device matches the pattern, only the first one enabled.
```
bourtemb: "only the first one enabled." => "only the first one is enabled." or "only the first one will be enabled."
Same comment for the disable method.

mliszcz: THanks! Fixes as proposed

mliszcz:
Please do not merge yet, `GroupReply::group_element_enabled` needs to be documented as well

cppapi/client/group.h
```diff
@@ -157,11 +157,17 @@ class GroupReply
               bool group_element_enabled = true);
   //- dtor
   virtual ~GroupReply ();
-  //- group_element_enabled accessor
+///@publicsection
+/**
+ * Check if the group element coresponding to this reply is enabled.
```
bourtemb: coresponding => corresponding (lines 162 and 164)

bourtemb: Not sure it is necessary to duplicate the line. I understand one is a summary and the other should be a more detailed explanation on what this method is doing. In this case, it is redundant since the summary is detailed enough...?
I don't mind keeping it like that but I find it cleaner to avoid duplication.

mliszcz: Both short and long description are shown when you click the function name in the html documentation, so I removed the long one. Also, we have other functions with only the short description, so the change seems to be ok. Thanks!

PR #581, opened 2019-07-30
Remove some compiler warnings

t-b:
I've compiled tango in debug mode with both gcc 9.1 and clang 8.0.
I then fixed the obvious and easy to fix warnings. My original motivation was to get into a state were we can use `-Werror` for a travis job to avoid introducing new warnings.

But that would require some more work, this PR is the first step in that direction though.

Logs with remaining warnings: 
- [output-clang-cppTango.txt](https://github.com/tango-controls/cppTango/files/3447107/output-clang-cppTango.txt)
- [output-gcc9-cppTango.txt](https://github.com/tango-controls/cppTango/files/3447109/output-gcc9-cppTango.txt)

### DIFF
--- a/cpp_test_suite/asyn/asyn_attr.cpp
+++ b/cpp_test_suite/asyn/asyn_attr.cpp
@@ -65,7 +65,7 @@ int main(int argc, char **argv)
 				assert( db == 5.55 );
 				finish = true;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				finish = false;
 				coutv << "Attribute not yet read" << std::endl;
@@ -103,7 +103,7 @@ int main(int argc, char **argv)
 				assert( enc_data.encoded_data.length() == 4 );
 				finish = true;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				finish = false;
 				coutv << "Attribute not yet read" << std::endl;
@@ -136,7 +136,7 @@ int main(int argc, char **argv)
 				assert( l == 5.55 );
 				finish = true;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				coutv << "Attribute not yet read" << std::endl;
 				nb_not_arrived++;
@@ -191,7 +191,7 @@ int main(int argc, char **argv)
 				finish = true;
 				delete received;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				finish = false;
 				nb_not_arrived++;
@@ -233,7 +233,7 @@ int main(int argc, char **argv)
 				finish = true;
 				delete received;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				coutv << "Attribute not yet read" << std::endl;
 				nb_not_arrived++;
@@ -313,7 +313,7 @@ int main(int argc, char **argv)
 				double db;
 				(*received) >> db;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				finish = false;
 				nb_not_arrived++;
@@ -357,7 +357,7 @@ int main(int argc, char **argv)
 				double db;
 				(*received) >> db;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				coutv << "Attribute not yet read" << std::endl;
 			}
--- a/cpp_test_suite/asyn/asyn_write_attr_multi.cpp
+++ b/cpp_test_suite/asyn/asyn_write_attr_multi.cpp
@@ -75,7 +75,7 @@ int main(int argc, char **argv)
 				device->write_attributes_reply(id);
 				finish = true;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				finish = false;
 				coutv << "Attributes not yet written" << std::endl;
@@ -105,7 +105,7 @@ int main(int argc, char **argv)
 				device->write_attributes_reply(id,200);
 				finish = true;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				coutv << "Attributes not yet written" << std::endl;
 				nb_not_arrived++;
@@ -154,7 +154,7 @@ int main(int argc, char **argv)
 				device->write_attributes_reply(id);
 				finish = true;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				finish = false;
 				nb_not_arrived++;
@@ -195,7 +195,7 @@ int main(int argc, char **argv)
 				device->write_attributes_reply(id,200);
 				finish = true;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
 			{
 				coutv << "Attributes not yet written" << std::endl;
 				nb_not_arrived++;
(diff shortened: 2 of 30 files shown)


bourtemb:
> /home/firma/devel/cppTango/cppapi/server/idl/tangoSK.cpp:2100:57: warning: unused parameter ‘cd’ [-Wunused-parameter]
 2100 | _0RL_lcfn_6fe2f94a21a10053_b0000000(omniCallDescriptor* cd, omniServant* svnt)

This one can be removed by generating tangoSK.cpp with omniidl coming from omniORB >= 4.2.2

cpp_test_suite/asyn/asyn_write_attr_multi.cpp
```diff
@@ -267,15 +267,16 @@ int main(int argc, char **argv)
 		finish = false;
 		bool failed = false;
 		nb_not_arrived = 0;
-		long nb_except,faulty_idx;
+		long nb_except = 0;
+    long faulty_idx = 0;
```
bourtemb: Indentation

cppapi/client/filedatabase.cpp
```diff
@@ -713,7 +713,11 @@ std::string FileDatabase::parse_res_file(const std::string &file_name)
 		    	 			break;
 
 	           			default:
-                     				return "COLON or -> expected at line " + StartLine;
+                         {
```
bourtemb: indentation

cppapi/client/filedatabase.cpp
```diff
@@ -791,7 +795,11 @@ std::string FileDatabase::parse_res_file(const std::string &file_name)
 		 			break;
 
 	       			default:
-                 			return "SLASH or -> expected at line " + StartLine;
+                      {
```
bourtemb: indentation

cppapi/client/filedatabase.cpp
```diff
@@ -844,12 +852,20 @@ std::string FileDatabase::parse_res_file(const std::string &file_name)
 	     			break;
 
 	   		default:
-             			return "SLASH or -> expected at line " + StartLine;
+              {
```
bourtemb: indentation

cppapi/client/filedatabase.cpp
```diff
@@ -844,12 +852,20 @@ std::string FileDatabase::parse_res_file(const std::string &file_name)
...
+                TangoSys_MemStream desc;
+                desc << "SLASH or -> expected at line " << StartLine;
+                return desc.str();
+              }
 			}
 			break;
 
       		default:
-        		return "Invalid resource name get  instead of STRING al line " + StartLine;
+              {
```
bourtemb: indentation

t-b:
@bourtemb Thanks for the review. I've fixed the indentation issues. Not so easy to replicate proper indentation with mixed-indentation being the norm here ;)

bourtemb:
> @bourtemb Thanks for the review. I've fixed the indentation issues. Not so easy to replicate proper indentation with mixed-indentation being the norm here ;)

I know and you're right to complain ;-)...
We never dared to unify this until now because it will imply changes in so many files and lines and would make it a bit more difficult to identify changes between different versions, even though some IDEs like CLion provide some options to ignore such changes when comparing 2 versions.

I think the best way to tackle this problem would be to provide some configuration files for clang-format or CLion, defining clearly the coding style rules we want. And then to have a tool applying automatically these rules to all our files and to do a massive commit.
We did some work with @ingvord using CLion code styles settings (https://github.com/tango-controls/cppTango/blob/tango-9-lts/.idea/codeStyleSettings.xml) but we never applied it to all of the files. I always tried to minimize the differences when proposing a patch because it makes it more difficult to review if all the indentation changed.

cppapi/server/jpeg/jpeg_bitstream.h
```diff
@@ -84,7 +84,6 @@ class OutputBitStream {
 #endif
    unsigned char *bufferPtr;
    unsigned char *numbits;
-   unsigned char  bScratch[4];
```
bourtemb: It looks like this bScratch field is used when JPG_USE_ASM_PB macro is defined. 
This macro seems to be linked to --enable-jpegmmx/--disable-jpgmmx options given to configure script of the Tango source distribution.
So, we should probably surround bScratch with some
#ifdef JPG_USE_ASM_PB directives

cppapi/client/filedatabase.h
```diff
@@ -199,17 +199,12 @@ class FileDatabase
...
-  	int 			pos_buf;
   	int 			CrtLine;
   	int 			StartLine;
   	char 			CurrentChar;
   	char 			NextChar;
 
-  	bool 			DELETE_ENTRY;
   	std::string 			word;
-
-	FileDatabaseExt	*ext;
```
bourtemb: The ext data member should be kept here because it is there to help us to extend this class without breaking the binary compatibility in the future.
For this version, I think it is fine to remove the other fields because we will break the binary compatibility anyway in the next version but I think we should keep the ext data member.

By the way, this reminds me that since the next version will not be binary compatible with 9.3.3, we should take this opportunity to move all the features (data members/ methods) which have been implemented in some ext classes into their corresponding Tango class.

cppapi/server/jpeg/jpeg_bitstream.h
t-b: @bourtemb Done. This was also the reason why the 32bit appveyor tests failed.

cpp_test_suite/asyn/asyn_attr.cpp
```diff
@@ -65,7 +65,7 @@ int main(int argc, char **argv)
 				assert( db == 5.55 );
 				finish = true;
 			}
-			catch (AsynReplyNotArrived )
+			catch (AsynReplyNotArrived&)
```
Ingvord: Should not be this **const** as well?

cppapi/client/helpers/DeviceProxyHelper.h
```diff
@@ -424,7 +424,7 @@ namespace Tango
 		//  Overloaded commands to  avoid usage of DevVarXX ARRAY for argout
 		//---------------------------------------------------------------------------
 		template <class _IN>
-			void internal_command_inout (const std::string& cmd_name, const _IN& argin, DevVarDoubleStringArray* argout,std::string file= __FILE__, int line= __LINE__)
+			void internal_command_inout (const std::string& TANGO_UNUSED(cmd_name), const _IN& TANGO_UNUSED(argin), DevVarDoubleStringArray* TANGO_UNUSED(argout), std::string file= __FILE__, int line= __LINE__)
```
Ingvord: Should we maybe remove these unused parameters at all?

cppapi/client/filedatabase.h
t-b: Done

cpp_test_suite/asyn/asyn_attr.cpp
t-b: You are right in principle but I think the current usage patterns tell us otherwise

```
$git grep "DevFailed" | grep catch | grep const | wc -l
21
firma@thomas:~/devel/cppTango (remove-some-compiler-warnings)
$git grep "DevFailed" | grep catch | grep -v const | wc -l
550
```

And if you look at the class structure of DevFailed/CorbaException there are no non-const member functions.

cppapi/client/helpers/DeviceProxyHelper.h
t-b: `TANGO_UNUSED` instead of just skipping the name has the advantage that we still have the name which tells us what is supposed to be passed in.

But I don't have a strong opinion here, so I can also change it.

cpp_test_suite/asyn/asyn_attr.cpp
Ingvord: Ok, got it! Thanks!

t-b:
The compile error in travis is real and fixed in https://github.com/tango-controls/cppTango/pull/587.

cppapi/client/helpers/DeviceProxyHelper.h
bourtemb: I vote for using TANGO_UNUSED for this specific example so we have a better understanding of the meaning of the arguments.
Sometimes we can get also this kind of warnings in methods where some arguments are used only when compiling with some specific compiler versions or with some specific compiler flags (-Dxxxx).
In this case it is also good to use TANGO_UNUSED.

t-b:
@Ingvord Merge?

t-b:
@bourtemb any objections with merging this?

PR #587, opened 2019-08-01
CMakeLists.txt: Fix compilation problems on debian buster

t-b:
Since 60fcadac (CMakeLists.txt: Check for more modern -std= flags as
well, 2019-07-25) we are using the most modern -std flags avaiable.

And since 4e259462 (.travis: Run tests on debian 10 as well, 2019-06-05)
we use debian buster for testing as well.

The combination of the two features broke compilation on debian buster
as it's gcc supports C++17 but we actually don't have valid C++17 code
as omniorb uses throw specifications, which are not valid anymore.

The error looks like

/usr/include/omniORB4/internal/orbOptions.h:95:57: error: ISO C++17 does not allow dynamic exception specifications
     virtual void visit(const char* value,Source source) throw (BadParam) = 0;

...

So let's not use C++17 and C++20 for now.

### DIFF
--- a/configure/CMakeLists.txt
+++ b/configure/CMakeLists.txt
@@ -176,23 +176,19 @@ endif(WIN32)
 
 if(NOT WIN32)
     include(CheckCXXCompilerFlag)
-    CHECK_CXX_COMPILER_FLAG("-std=c++20" COMPILER_SUPPORTS_CXX20)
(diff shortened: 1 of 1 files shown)


t-b:
This is a known issue upstream, see http://www.omniorb-support.com/pipermail/omniorb-list/2019-May/032064.html (scroll down).

PR #589, opened 2019-08-21
Fix for Unstable event::pipe_event testcase (#524)

Ingvord:
I have played a little bit with the code in #589 

And here are some of my thoughts:

1) If we want to get rid off sleeps that basically means in the code below `device->subscribe_event` becomes a blocking function which is blocked till the client executes first callback. This is technically achievable but will rely on sync call. Which I am strictly against (there must not be any sync calls during subscription what so ever, but that is a different story)

```cpp
//
// subscribe to a pipe event
//
		int eve_id1 = device->subscribe_event("RWPipe",Tango::PIPE_EVENT,&cb);
//
// The callback should have been executed once
//
		assert (cb.cb_executed == 1);

```

2) due to 1) I would use conditional variables in callback functions to fix this (and maybe others) test(s). I think it is OK, even though C++11 etc as this change will affect only tests and not library code

3) We can think about improving hand-shake, but I do not see clear benefit vs complexity. The nature of PUB/SUB is that PUB lives its own live sending events, SUBs occasionally appear/disappear receiving some of the events. It MUST be OK for a SUB to miss some events 

I will continue with conditional variables approach, while the rest (especially 3) requires severe discussion


Ingvord:
The test is now much more stable. I ran it continuously for a quite long period:

```
In PipeEventTestSuite__loop::test_pipe_event__loop:
/home/ingvord/Projects/org.tango/git/cppTango/cpp_test_suite/cxxtest/include/../../new_tests/cxx_pipe_event.cpp:298: Error: Expected (8 == cb_executed), found (8 != 7)
TEST FAILED in:
192 iteration of test case
4 iteration of test suite


Process finished with exit code 255
```

So it did fail, but I guess we can leave with that for now.

t-b:
@Ingvord I like the new test and its approach. *But* it is using C++11 and AFAIR we still allow C++98 only compilers. Or has something changed?

mliszcz:
I don't like the idea how mutex locking and waiting on conditional is mixed with the test case logic. This "supporting" behavior could be encapsulated in a dedicated class.
Also, I think we should have a separate PR for porting the testcases to CxxTest (if we agree on doing so - this needs further discussions).

I did something similar (a bit more generic) for another testcase in https://github.com/tango-controls/cppTango/pull/558 (still open, not much feedback received).

IMO the way forward for improving tests quality and stability is:
* decide if we can use use C++11 in tests (per Thomas' comment)
* implement an utility class which encapsulates the conditional variable and handles synchronization
* gradually apply this new utility class to all the tests

BTW, there is an empty commit (https://github.com/tango-controls/cppTango/commit/dcb06630fda510d8b925e50ca252937234f8f04c) that I think can be removed.

bourtemb:
> Also, I think we should have a separate PR for porting the testcases to CxxTest (if we agree on doing so - this needs further discussions).

This needs further discussions indeed.

> IMO the way forward for improving tests quality and stability is:
> 
>     * decide if we can use use C++11 in tests (per Thomas' comment)

This is something we already talked about. There is no problem in introducing C++11 in tests code.
I remember a discussion where we even said that we could use C++ 11 in kernel code after the release of a stable Tango 9.3 Source distribution.
But I didn't find any kernel meeting minutes where this was written.

>     * implement an utility class which encapsulates the conditional variable and handles synchronization
> 
>     * gradually apply this new utility class to all the tests

This sounds like a good idea.

> BTW, there is an empty commit ([dcb0663](https://github.com/tango-controls/cppTango/commit/dcb06630fda510d8b925e50ca252937234f8f04c)) that I think can be removed.

The Pull Requests from @ingvord are always developed in a way where the commits are squashed during the merge so this is not a problem.

Ingvord:
@t-b , @mliszcz , @bourtemb 

Thanks for the input! Really appreciate that!

>> Also, I think we should have a separate PR for porting the testcases to CxxTest (if we agree on doing so - this needs further discussions).

>This needs further discussions indeed.

Porting all tests seems like a lot of dumb work without big benefit. I would rather migrate gradually: once someone needs to do something with a test, try to migrate it. Like in this PR. Complicates code review indeed, but on the other hand it is quite rare occasion

> - decide if we can use use C++11 in tests (per Thomas' comment)

I feel everyone is agree that tests code can support anything. It is not a part of the distribution, and I doubt non-developers will even try to compile it non-speaking about running it. In fact would be nice to ensure that TangoSourceDistribution does not compile tests which is also good as save a lot of time compilation.

> - implement an utility class which encapsulates the conditional variable and handles synchronization
> - gradually apply this new utility class to all the tests

+1, I would not do this in this PR though. Optional solution - merge this one, implement class in a dedicated PR, refactor this one and other tests in further PRs

> The Pull Requests from @Ingvord are always developed in a way where the commits are squashed during the merge so this is not a problem.

Yeah, I prefer to separate work on a PR and the master branch. For me the output of a PR is usually a single commit into the master branch, while in the PR there are maybe a lot of "experimental" commits

cpp_test_suite/new_tests/cxx_pipe_event.cpp
```diff
@@ -0,0 +1,315 @@
...
+        public:
+            EventCallBack(PipeEventTestSuite__loop* outer):
+            root_blob_name(""),
+            nb_data(0),
+            outer(outer)
+            {}
+            std::string root_blob_name;
+            size_t	nb_data;
+            PipeEventTestSuite__loop* outer;
+            void push_event(Tango::PipeEventData* event_data){
```
mliszcz: Inconsistend brace placement. Should be in next line.

cpp_test_suite/new_tests/cxx_pipe_event.cpp
```diff
@@ -0,0 +1,315 @@
...
+                        else
+                        {
+                            coutv << "Error sent to callback" << std::endl;
+//			Tango::Except::print_error_stack(event_data->errors);
+                        }
+                }
+                catch (...)
+                {
+                    coutv << "EventCallBack::push_event(): could not extract data !\n";
+                }
```
mliszcz: This whole block should be moved 4 spaces right to be aligned with corresponding `try`.

cpp_test_suite/new_tests/cxx_pipe_event.cpp
```diff
@@ -0,0 +1,315 @@
...
+            EventCallBack(PipeEventTestSuite__loop* outer):
+            root_blob_name(""),
+            nb_data(0),
+            outer(outer)
+            {}
+            std::string root_blob_name;
+            size_t	nb_data;
+            PipeEventTestSuite__loop* outer;
+            void push_event(Tango::PipeEventData* event_data){
+                {
```
mliszcz: Unnecessary scope. Can be removed.

cpp_test_suite/new_tests/cxx_pipe_event.cpp
```diff
@@ -0,0 +1,315 @@
...
+
+        int eve_id1 = device1->subscribe_event("RWPipe",Tango::PIPE_EVENT,&cb);
+
+//
+// The callback should have been executed once
+//
+        {
+            std::unique_lock<std::mutex> lk(m);
+            cv.wait_for(lk, std::chrono::milliseconds(1000));
+        }
```
mliszcz: Can you extract this to a function? Maybe just a lambda on top of the testcase, like:

```cpp
auto wait_for_event = [&lk, &m, &cv](int millis)
{
            std::unique_lock<std::mutex> lk(m);
            cv.wait_for(lk, std::chrono::milliseconds(1000));
}
```

And then just call it instead of copy-pasting the whole block.

mliszcz (changes requested):
Ok then. I've added some minor comments (whitespaces and duplications). @Ingvord please take a look. Other than that, the PR looks good to me.

@bourtemb @Ingvord any chance you can take a look at #558? Thanks!

mliszcz:
One more comment:

>> The Pull Requests from @Ingvord are always developed in a way where the commits are squashed during the merge so this is not a problem.
>
> Yeah, I prefer to separate work on a PR and the master branch. For me the output of a PR is usually a single commit into the master branch, while in the PR there are maybe a lot of "experimental" commits

I don't care about intermediate or experimental commits (I also make such commits), and it's a good idea to squash related commits that are not useful on their own. But in my opinion the outcome of this PR should still be two separate commits: one that replaces every sleep() with waiting on a CV and other one that ports the refactored testase to CxxTest.

bourtemb:
> I doubt non-developers will even try to compile it non-speaking about running it. In fact would be nice to ensure that TangoSourceDistribution does not compile tests which is also good as save a lot of time compilation.

TangoSourceDistribution does not compile the tests.

t-b:
I would actually prefer the solution in #524 if I may be so bold.

PR #593, opened 2019-10-18
Bump release nb

bourtemb:
We said the next cppTango release will not be binary compatible with the previous version.
Increase the release number to 10.0.0 in preparation of this new release.
Update CHANGELOG.md with latest commits since cppTango 9.3.3.

This is not the Tango 10 we talk about since a while, but it would be logical to call it like that since we break the binary compatibility.
We could also call it 9.4.0 and change the SONAME of the generated library...
But 10.0.0 is probably cleaner to follow the most used versioning conventions?
Comments are welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,7 @@ endif()
 
 #convenient versions
 set(LIBRARY_VERSION "${MAJOR_VERSION}.${MINOR_VERSION}.${PATCH_VERSION}")
-set(SO_VERSION "${MAJOR_VERSION}")
+set(SO_VERSION "${MAJOR_VERSION}${MINOR_VERSION}")
 
 set(TANGO_HOST $ENV{TANGO_HOST})
 include(configure/CMakeLists.txt)


t-b:
We can use 10 for the next version. Of course some people will get confused because it is not the big thing we are talking about since ages.

I don't see an advantage over using 9.4 though. The binary compatiblity is something which does not need to be part of the first digit of the version number. As long as we change the SONAME ;)

mliszcz:
Hi all,
copying my comments from discussion with Reynald on slack:

I think if we break binary compatibility, major version shall be increased. As far as I know, binary compatibility is not in scope of e.g. semver specification (https://semver.org/), but from end-user perspective it is a breaking change. Also it may impact e.g. debian packaging.

I'm not an expert here but I'm just trying to read debian guidelines: https://www.debian.org/doc/debian-policy/ch-sharedlibs.html It says: *To determine the soversion, look at the SONAME of the library, stored in the ELF SONAME attribute. It is usually of the form name.so.major-version (for example, libz.so.1)*

So this raises some questions:
* can SONAME be e.g. "libtango.so.9.4"?
* Will it be installed as "libtango.so.9.4.minor.patch" (with 4 dots)?
* What should be the name of debian package? "libtango9" or "libtango94"?

Currently we have:
```
lrwxrwxrwx 1 root root         13 Oct  9 16:07 libtango.so -> libtango.so.9
lrwxrwxrwx 1 root root         17 Oct 14 16:03 libtango.so.9 -> libtango.so.9.3.3
-rw-r--r-- 1 root root   52316536 Oct 14 16:02 libtango.so.9.3.3
```

Later if we add:
```
-rw-r--r-- 1 root root          0 Oct 28 12:03 libtango.so.9.4
-rw-r--r-- 1 root root          0 Oct 28 12:03 libtango.so.9.4.0
```

* Should libtango.so point to libtango.so.9 or libtango.so.9.4?
* Is it safe to change libtango.so.9 to point to libtango.so.9.4 instead of libtango.so.9.3.3? - no, but user won't know
* And so on ...

In my opinon we should go with 10.0 version in tango-9-lts branch as it will be less surprising to the users, and for the "new" Tango (aka v10) we shall find a new name (and not rename to v11).

bourtemb:
> Agree with Michal, binary compatibility is not in the scope of semantic versioning. But, in my view, this still has to be Tango 9, stated it is still source-compatible (e.g. no major, incompatible, change in the API). 

Indeed, it is still source compatible... more or less, actually.
You'll need to add `using namespace std;` somewhere in your code to be compatible with the old code.

> Thus, Major.Minor.Patch should read 9.4.0 and SONAME.... needs to be libtango.so.9 as per Debian guidelines...

I think SONAME should not be libtango.so.9 in any case since it is not binary compatible.

> What about sticking with 9, that is in line with semver and SONAME rules and make a BIG announcement that 9.4.0 is NOT binary compatible?
> Tango 10 is already something different in the community...

Then, what about making a BIG announcement that Tango 10 will be different than what the Tango community is expecting and find another name as Michal proposed for the big major update of Tango?

lorenzopivetta:
> Then, what about making a BIG announcement that Tango 10 will be different than what the Tango community is expecting and find another name as Michal proposed for the big major update of Tango?

That will work as well, but I'm not sure what "another name" means... is it about rebranding TANGO? Or just the filename of the library? Meaning it will read different than libtango with TANGO still being called TANGO?

mliszcz:
@bourtemb @lorenzopivetta 
I was just thinking about a name like "tango-*something*", like tango-next, tango-kernel, tango-core, for the c++ library so that you can have both installed on the same system:
* libtango.so.9, libtango.so.10, etc. ...
* and libtangowhatever.so.10, libtangowhatever.so.11

This way we will avoid conflicts in .so names and debian package names (packages are named like libtango8, libtango9).

Of course this is assuming that we want to support and develop both 9-lts and "big" v10 in parallel. We don't know how many major versions of of 9-lts branch we will have (assuming that for each breaking change we will need a new major version).

andygotz:
This is a tricky question. I could imagine that the using namespace std could be part of the LTS 9 version but does this break binary compatible as well? Changing the API for LTS seems against the aim of LTS. The other question I have is do we know how many changes are planned? If we make a new version what are the plans to make it binary compatible for the future? We might make multiple changes which are not binary compatible. What are the plans / ideas on this?

Making a new release with a different name e.g. post-9 or pre-10 for example sounds like a possibility but we need to define the goals i.e. will it replace 9 LTS, do we continue maintaining the LTS 9 version?

Other questions are how many binary incompatible changes are planned? If none can we say this is exceptional?

How do other packages handle binary incompatibilities and LTS support?

Sorry for asking more questions than providing solutions ...

Andy

lorenzopivetta:
Just adding some thoughts to the discussion...

> Changing the API for LTS seems against the aim of LTS.

This is a key point to me... I mean, if we're not changing the API, in my view this is still Tango 9, and should stick to "9 series", e.g. 9.4.0 (even if requires recompiling...). On the contrary... it should be Tango 10... but then this change may (should?) not go into 9 LTS... But I recognize it is not exactly straightforward.
Concerning binary compatibility, I remember the discussion during the Tango meeting in Hamburg, where the conclusion was that we might drop binary compatibility for Tango 10 if required by a brand new, and advantageous, architectural design... so, if I'm correct, it was an option for v10 (well, in my understanding, likely to be true for v10).

bourtemb:
Here is an example of a coming pull request which will break the binary compatibility:
https://github.com/tango-controls/cppTango/pull/574

It should still be source code compatible but will not be binary compatible.

https://github.com/tango-controls/cppTango/pull/581 as well introduced some low severity problems with the binary compatibility because some fields have been removed from the FileDatabase class, changing its size.

https://github.com/tango-controls/cppTango/pull/480 requires a change in the Tango IDL. This is also a good example of a coming PR for which we would like to change the major version.

I propose to use tools like the [abi-compliance-checker](https://lvc.github.io/abi-compliance-checker/) to check the binary compatibility regularly.
We will need to change the debug compilation flags to use "-g -Og" options because this is required by this tool to get pertinent analysis results.

bourtemb:
> #581 as well introduced some low severity problems with the binary compatibility because some fields have been removed from the FileDatabase class, changing its size.

In fact, I think this one should not generate any problem because this FileDatabase class should be used only internally (it is not documented in the Doxygen documentation).

t-b:
My intention in the cleanup PRs was to not-break ABI/API compatibility. If that is the case I'm happy to make fixup PR.

I think we should only allow ABI breaking if we really add functionality like in #480. Just some cleanups should not be the reason for having to change the ABI.

bourtemb:
The version number for the next release was discussed during a Tango Kernel Teleconf meeting:
https://github.com/tango-controls/tango-kernel-followup/blob/master/2019/2019-11-15/Minutes.md#cpptango-next-release-number

> The next version of the Tango C++ library, which will not be binary compatible, will be named 9.4 and the soname will be changed. This solution was preferred over naming it Tango 10, which would have been confusing for the users, because Tango 10 is supposed to be the first version of Tango on the way to replace CORBA.
> 
> New Debian package names will have to be used for this version of the C++ library which will not be binary compatible. A name like libtango94 could be used. A similar solution has already been adopted by other projects, as Thomas Braun showed during the meeting.

I think we just need to agree on the soname name to be used: 
Would libtango.so.94 be fine?
Better ideas?

t-b:
`libtango.so.94` is fine with me.

andygotz:
libtango.so.94 sounds good to me too.

@t-b can you point some of the other projects which have a similar naming scheme? Just curious.

t-b:
@andygotz libjpeg has, see https://packages.debian.org/buster/arm64/libjpeg62-turbo/filelist

PR #597, opened 2019-10-22
Add debian buster (for 9.3-backports branch)

t-b:


.travis.yml
```diff
@@ -4,14 +4,11 @@ language: cpp
...
   matrix:
-    - OS_TYPE=debian9 CMAKE_BUILD_TYPE=DEBUG SONAR_SCANNER=OFF COVERALLS=OFF
-    - OS_TYPE=debian9 CMAKE_BUILD_TYPE=RELEASE SONAR_SCANNER=OFF COVERALLS=OFF
-    - OS_TYPE=debian8 CMAKE_BUILD_TYPE=Debug SONAR_SCANNER=ON COVERALLS=ON
-    - OS_TYPE=debian8 CMAKE_BUILD_TYPE=RELEASE SONAR_SCANNER=OFF COVERALLS=OFF
-    - OS_TYPE=debian7 CMAKE_BUILD_TYPE=RELEASE SONAR_SCANNER=OFF COVERALLS=OFF
-#    - OS_TYPE=debian7 CMAKE_BUILD_TYPE=DEBUG
-#    - OS_TYPE=win32
-# etc
+    - OS_TYPE=debian10 CMAKE_BUILD_TYPE=Release SONAR_SCANNER=OFF COVERALLS=OFF NON_STOCK_CPPZMQ=ON
```
bourtemb: What could be even more interesting in the case of this branch (and maybe for the master branch too?) would be to test on Debian Buster in Release mode at least (and maybe on stretch too) with NON_STOCK_CPPZMQ=OFF since this is the use case for which this 9.3-backports branch was originally created

bourtemb: > NON_STOCK_CPPZMQ=ON means that a custom cppzmq is installed. =OFF means that we take the system default. 

This is exactly what I had understood.

> So we only use a custom cppzmq for debian 8, for all the others we use the system cppzmq.

Debian 8 is the only one with NON_STOCK_CPPZMQ=OFF, which means Debian 8 is the only one using the system default according to your first sentence and to my understanding.

t-b:
PR is currently blocked until #598 is merged. After that this PR needs these changes as well.

t-b:
Integrated #598.

PR #599, opened 2019-10-22
Fix coveralls

t-b:
I think we still want coveralls.io support or?

In case yes, this PR enables it again. And also installs git in the docker images as that gives use more metadata for the report.

And some docker images cleanup as well.


t-b:
Thanks both. Works again, see https://coveralls.io/builds/26476119.

PR #605, opened 2019-10-24
Add warning error testing with clang

t-b:


cppapi/client/eventconsumer.h
```diff
@@ -772,7 +772,16 @@ class DelayedEventSubThread: public omni_thread
 	EventType       et;
 	CallBack        *callback;
 	EventQueue      *ev_queue;
+
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunused-private-field"
+#endif
 	bool            stateless;
```
mliszcz: I could not find a place where this one is used. Maybe clang is right and it can be removed?

t-b: @mliszcz I've tried removing it, but it is used in the ctor initalizer list, https://github.com/tango-controls/cppTango/pull/605/files/922d4df802b3987d80bad4f669c43b6e482446f0#diff-b8e05bd512a1ddaf88c34d6cb0153bbbL764.

mliszcz: Yes, but other than that, it is not used. You can remove it from constructor initializer list as well (and remove constructor parameter `bool _stateless` or wrap it in `TANGO_UNUSED` if it's a public API).

There's no point in storing it if we are not using it.

The parameter could be removed also from call sites, but that is a different issue, out of this PR's scope.

bourtemb: > > Yes, but other than that, it is not used. You can remove it from constructor initializer list as well (and remove constructor parameter `bool _stateless` or wrap it in `TANGO_UNUSED` if it's a public API).

I think we can remove from the constructor parameter list. This class is only used internally.

> The parameter could be removed also from call sites, but that is a different issue, out of this PR's scope.

The constructor seems to be called at only one location (in EventConsumer::subscribe_event() in cppapi/client/event.cpp). There is even a test before in this subscribe_event() method ensuring stateless parameter is always true.

t-b: @mliszcz 

> There's no point in storing it if we are not using it.

Yes of course, but this PR is not about refactoring the code. Nevertheless I've removed the private variable `stateless_` and wrapped the parameter via TANGO_UNUSED as suggested.

t-b: @bourtemb Our messages crossed. I'll remove the ctor parameter as well.

bourtemb:
It looks like some other projects ended up disabling this Clang warning:
https://github.com/dealii/dealii/issues/3705#issuecomment-269190302

I think we are in a similar use case here. We don't get any linker error in our case.

mliszcz:
I think the questions are:
1. is `MultiAttrProp` a public template?
2. what will happen if user will instantiate the template with "unsupported" type?

If answer to first is "no" or to second is "nothing", then I see no point in leaving the check.

bourtemb:
> I think the questions are:
> 
>     1. is `MultiAttrProp` a public template?

Yes it is: https://tango-controls.github.io/cppTango-docs/classTango_1_1MultiAttrProp.html

>     2. what will happen if user will instantiate the template with "unsupported" type?

I guess it will get a linker error.

> If answer to first is "no" or to second is "nothing", then I see no point in leaving the check.

bourtemb:
> > ```
> > 2. what will happen if user will instantiate the template with "unsupported" type?
> > ```
> 
> I guess a linker error will occur.

I confirm. A linker error occurs in this case:

obj/TangoClient.o: In function `Tango::MultiAttrProp<MyClass>::MultiAttrProp()':
/usr/include/tango/attrprop.h:558: undefined reference to `Tango::ranges_type2const<MyClass>::enu'
collect2: error: ld returned 1 exit status

t-b:
I've resolved conflicts.

t-b:
@bourtemb Is it good as is or should I ignore the warning in CmakeLists.txt?

bourtemb:
@t-b , I am in favour of reverting the changes from https://github.com/tango-controls/cppTango/commit/62ecfea66c62f772d5c0e39a81284f754f72e5ac and ignoring the warning for this part of the code.

t-b:
@bourtemb Merge?

t-b:
@bourtemb Thanks. Should be fixed now.

t-b:
@mliszcz Good?

Issue #495, opened 2018-11-05
Incorrect getting of the list of devices from a pattern for Tango::Group if using a remote host.


elkinvg:
incorrect getting of the list of devices from a pattern for Tango::Group if using a remote host.

Tango 9.2.5
OmniORB 4.2.1
0MQ 4.3.1

For group from localhost

```
[elkin@tango-linac test_for_group]$ ./test_for_group "sys/tg_test/*"
Getting information from Tango::Group

         Number of devices in pattern: 1

         List of Devices:
                sys/tg_test/1

command_inout for Tango::Group


         Device Name: sys/tg_test/1
         State: 1

```

For group from remote host.

The correct list of devices from group but without `REMOTE_TANGO_HOST`.

```
[elkin@tango-linac test_for_group]$ ./test_for_group "tango://{{REMOTE_TANGO_HOST}}:10000/sys/tangotest/*"
Getting information from Tango::Group

         Number of devices in pattern: 6

         List of Devices:
                sys/TangoTest/1
                sys/TangoTest/2
                sys/tangotest/3
                sys/tangotest/33
                sys/TangoTest/4
                sys/TangoTest/crio

command_inout for Tango::Group

device sys/tangotest/1 not defined in the database !
Failed to execute command_inout on device sys/database/2, command DbImportDevice
Can't connect to device sys/tangotest/1

```

[pattern_from_remote_host.zip](https://github.com/tango-controls/cppTango/files/2548483/pattern_from_remote_host.zip)

bourtemb:
Hi,

Groups indeed, when created using patterns, do not seem to support (yet) devices coming from a TANGO_HOST different than the TANGO_HOST of the client. This can be very confusing indeed!

Please be careful when using patterns to define a group because the group will be created with all the devices matching the pattern **and** which are **exported** at the moment of the group creation.
If you have a device defined in the database but not exported at the moment of the group creation, this device will not be part of the group. This might be an issue in many use cases.
There is an issue already reported about this: https://github.com/tango-controls/cppTango/issues/303

Thank you very much for the bug report and the provided code sample.

elkinvg:
Thank you for your reply.

Now I usually solve this problem as shown below.

```cpp
string pattern = "tango://hostname:10000/name/of/device*";

// [0] tango://hostname:10000/}}
// [1] name/of/device*
std::pair<std::string> hostAndDevice = splitDeviceName(pattern);

group = new Tango::Group("name");
group->add(pattern);

[...]

PR #608, opened 2019-10-29
Add support for group devices in remote TANGO_HOST (attempt 2)

mliszcz:
Correction for following scenario:
* There are two Tango Database instances, H1 and H2.
* Device "my/test/device" is hosted in a server running in the first instance (TANGO_HOST=H1)
* Client application is runing in tje second instance (TANGO_HOST=H2)
* Client application creates a group using qualified wildcard names (`tango://H1/my/test/*`)

Bug:
* `tango://H1/my/test/*` is resolved to `my/test/device` instead of `tango://H1/my/test/device`

Fixes https://github.com/tango-controls/cppTango/issues/495

### DIFF
--- a/cpp_test_suite/environment/setup.sh
+++ b/cpp_test_suite/environment/setup.sh
@@ -2,36 +2,58 @@
 
 echo "Setup test environment"
 
-#run mysql-tango docker
-docker run --name mysql_db \
-    -e MYSQL_ROOT_PASSWORD=root \
-    -d tangocs/mysql:9.2.2 --sql-mode=""
+function run_mysql_container {
+    docker run \
+        --rm \
+        --name $1 \
+        -e MYSQL_ROOT_PASSWORD=root \
+        -e MYSQL_INITDB_SKIP_TZINFO=1 \
+        -d \
+        tangocs/mysql:9.2.2 \
+        --sql-mode="" \
+        --innodb=OFF \
+        --default-storage-engine=MyISAM
+}
 
-#run tango-cs docker
-CONTAINER=$(docker run --name tango_cs \
-    -p 10000:10000 \
-    -e TANGO_HOST=127.0.0.1:10000 \
-    -e MYSQL_HOST=mysql_db:3306 \
-    -e MYSQL_USER=tango \
-    -e MYSQL_PASSWORD=tango \
-    -e MYSQL_DATABASE=tango \
-    --link mysql_db:mysql_db \
-    -d tangocs/tango-cs:latest)
+function run_tango_container {
+    docker run \
+        --rm \
+        --name $1 \
+        -e TANGO_HOST=127.0.0.1:10000 \
+        -e MYSQL_HOST=$2:3306 \
+        -e MYSQL_USER=tango \
+        -e MYSQL_PASSWORD=tango \
+        -e MYSQL_DATABASE=tango \
+        --link $2:$2 \
+        -d \
+        tangocs/tango-cs:latest
+}
 
-echo "CONTAINER=$CONTAINER"
+run_mysql_container mysql_db
+run_mysql_container mysql_db2
+run_tango_container tango_cs mysql_db
+run_tango_container tango_cs2 mysql_db2
 
-IPADDR=$(docker inspect -f '{{ .NetworkSettings.IPAddress }}' $CONTAINER)
-#export TANGO_HOST
-TANGO_HOST=$IPADDR:10000
+IPADDR=$(docker inspect -f '{{ .NetworkSettings.IPAddress }}' tango_cs)
+IPADDR2=$(docker inspect -f '{{ .NetworkSettings.IPAddress }}' tango_cs2)
+
+export TANGO_HOST=$IPADDR:10000
+export TANGO_HOST2=$IPADDR2:10000
+
 echo "TANGO_HOST=$TANGO_HOST"
-export TANGO_HOST
 
 echo "Create tango_host file"
-echo "#!/bin/bash" > tango_host
-echo "export TANGO_HOST=$TANGO_HOST" >> tango_host
+
+cat << EOF > tango_host
+#!/bin/bash
+export TANGO_HOST=$TANGO_HOST
+export TANGO_HOST2=$TANGO_HOST2
+EOF
 
 echo "Wait till tango-cs is online"
-#TODO notification?
-sleep 30
-
-#ctest
+if hash tango_admin 2>/dev/null; then
+    tango_admin --ping-database 30
+    TANGO_HOST=$TANGO_HOST2 tango_admin --ping-database 30
+else
+    sleep 30
+fi
--- a/cpp_test_suite/new_tests/cxx_group.cpp
+++ b/cpp_test_suite/new_tests/cxx_group.cpp
@@ -827,6 +827,39 @@ public:
 		CxxTest::TangoPrinter::restore_unset("double_attr_value");
 	}
 
+	/* Verifies that a group can contain devices from a remote TANGO_HOST
+	 * (a Tango instance different from client's default TANGO_HOST).
+	 * An issue was reported when resolving names containing wildcards. */
+
+	void test_use_devices_from_remote_tango_host()
+	{
+		const std::string original_tango_host = std::getenv("TANGO_HOST");
+		const std::string external_tango_host = std::getenv("TANGO_HOST2");
+
+		const bool force_update = true;
+
+		TS_ASSERT_EQUALS(0, setenv("TANGO_HOST", external_tango_host.c_str(), force_update));
+		ApiUtil::instance()->cleanup();
+
+		Group group("group");
+		group.add("tango://" + original_tango_host + "/" + device1_name + "*");
+
+		GroupCmdReplyList command_results = group.command_inout("State");
+		GroupCmdReply command_result = command_results[0];
+
+		TS_ASSERT(not command_results.has_failed());
+		TS_ASSERT_EQUALS(1u, command_results.size());
+
+		TS_ASSERT(not command_result.has_failed());
+
+		DevState state;
+		TS_ASSERT(command_result >> state);
+		TS_ASSERT_EQUALS(ON, state);
+
+		TS_ASSERT_EQUALS(0, setenv("TANGO_HOST", original_tango_host.c_str(), force_update));
+		ApiUtil::instance()->cleanup();
+	}
+
 };
 #undef cout
 #endif // GroupTestSuite_h
--- a/cppapi/client/group.cpp
+++ b/cppapi/client/group.cpp
@@ -130,7 +130,7 @@ DeviceNames GroupElementFactory::resolve_remote_device_names(
 
 DeviceNames GroupElementFactory::add_host_and_port_to_device_names(
     const DeviceNames& names,
-    std::string& host,
+    const std::string& host,
     int port)
 {
     DeviceNames full_names;
@@ -138,16 +138,16 @@ DeviceNames GroupElementFactory::add_host_and_port_to_device_names(
 
     for (DeviceNames::const_iterator name = names.begin(); name != names.end(); ++name)
     {
-        full_names.push_back(build_full_device_name(host.c_str(), port, name->c_str()));
+        full_names.push_back(build_full_device_name(host, port, *name));
     }
 
     return full_names;
 }
 
 std::string GroupElementFactory::build_full_device_name(
-    const char* host,
+    const std::string& host,
     int port,
-    const char* device_name)
+    const std::string& device_name)
 {
     std::ostringstream buffer;
     buffer << "tango://" << host << ":" << port << "/" << device_name;
--- a/cppapi/client/group.cpp
+++ b/cppapi/client/group.cpp
@@ -32,6 +32,7 @@
 //=============================================================================
 
 #include <group.h>
+#include <cstdio>
 
 //-----------------------------------------------------------------------------
 // LOCAL DEBUGGING MACRO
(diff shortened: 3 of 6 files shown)


mliszcz:
Test (04f0d49) is failing without correction (c645525). You need to check travis logs. Job exit status is not working currently in CI.

cpp_test_suite/environment/setup.sh
```diff
@@ -2,36 +2,58 @@
...
+#!/bin/bash
+export TANGO_HOST=$TANGO_HOST
+export TANGO_HOST2=$TANGO_HOST2
+EOF
 
-#ctest
\ No newline at end of file
+echo "Wait till tango-cs is online"
+if hash tango_admin 2>/dev/null; then
+    tango_admin --ping-database 30
```
bourtemb: Great improvement!

cpp_test_suite/new_tests/cxx_group.cpp
```diff
@@ -827,6 +827,39 @@ class GroupTestSuite: public CxxTest::TestSuite
...
+		const std::string original_tango_host = std::getenv("TANGO_HOST");
+		const std::string external_tango_host = std::getenv("TANGO_HOST2");
+
+		const bool force_update = true;
+
+		TS_ASSERT_EQUALS(0, setenv("TANGO_HOST", external_tango_host.c_str(), force_update));
+		ApiUtil::instance()->cleanup();
+
+		Group group("group");
+		group.add("tango://" + original_tango_host + "/" + device1_name + "*");
```
bourtemb: Nice to have set up a test with 2 tango databases.
At first sight, it looks like a device named device1_name exists in both Tango control systems (tests are executed with "--device1=test/debian8/10" parameter).
It would have been interesting to have a way to ensure that we are really sending the state command to the right device.
Talking to at least 1 device having its name in only one DB would have been interesting.
Another way could be to ensure that the devices $TANGO_HOST/test/debian8/10 and $TANGO_HOST2/test/debian8/10 return different values for the state.... or to read an attribute for which we could set a specific value before the test

cppapi/client/group.cpp
```diff
@@ -48,63 +49,109 @@ bool GroupReply::exception_enabled = false;
 //=============================================================================
 // class GroupElementFactory
 //=============================================================================
-GroupElements GroupElementFactory::instanciate (const std::string& p, int tmo_ms)
+GroupElements GroupElementFactory::instanciate(
+    const std::string& name_or_pattern,
+    int timeout_millis)
```
bourtemb: Good initiative to rename the parameters

cppapi/client/group.cpp
```diff
@@ -48,63 +49,109 @@ bool GroupReply::exception_enabled = false;
...
+    return full_names;
+}
+
+std::string GroupElementFactory::build_full_device_name(
+    const char* host,
+    int port,
+    const char* device_name)
+{
+    char buffer[255];
+    std::sprintf(buffer, "tango://%s:%d/%s", host, port, device_name);
```
bourtemb: I would use snprintf instead of sprintf here.
Here we limit the length of the fully qualified tango device name to 255 characters.
This should be enough in most cases but we have seen some use cases where people had very long host names.
Can we use C++ strings instead of a limited length C string here?

cppapi/client/group.cpp
```diff
@@ -854,7 +901,7 @@ void Group::remove_i (const std::string& p, bool fwd)
 #endif
   if (name_equals(p)) {
 #if defined(_LOCAL_DEBUGGING)
-    cout << "Group::remove_i::failed to remove " << _p << " (can't remove self)" << std::endl;
+    cout << "Group::remove_i::failed to remove " << p << " (can't remove self)" << std::endl;
```
bourtemb: well spotted

cppapi/client/group.cpp
mliszcz: The problem is that `snprintf` is available in C++11 and above.
Another problem is that the port is an integer. To convert integer to string in a portable way without using `std::to_string` (C++11 or 14), we need some kind of `*printf`.

So if we want to be memory-safe here, I see two options:
1. use `sprintf` just to convert the port number to a string. 11 characters should be enough for 32-bit number.
2. use `stringstream`

Do you have any preference here?

cpp_test_suite/new_tests/cxx_group.cpp
mliszcz: No, the device is defined only in the first (original) system (in `tango_cs` container).
The second one (`tango_cs2` container) is empty.

The devices are created by `conf_devtest` program invoked from `pre_test.sh` script just once, with TANGO_HOST pointing to the original system.

Please check the travis log for the commit that added the test:

```
Exception : 

Tango exception
Severity = ERROR 
Error reason = DB_DeviceNotDefined
Desc : device test/debian8/10 not defined in the database !
Origin : DataBase::ImportDevice()

...

In GroupTestSuite::test_use_devices_from_remote_tango_host:
../../new_tests/cxx_group.cpp:834: Error: Test failed: Exception thrown from test

TEST FAILED in:
1 iteration of test case
1 iteration of test suite
```

You can also verify this using the example provided in #495 after running `make start-tango`. It is failing without the correction:
```
TANGO_HOST=$TANGO_HOST2 LD_LIBRARY_PATH=~/Documents/cppTango/build ~/Documents/grouptest/a.out "tango://$TANGO_HOST/test/debian*/*"
```

cpp_test_suite/environment/setup.sh
mliszcz: This is probably 3rd PR where I'm adding this. Maybe finally we will get it on master ;)

cpp_test_suite/new_tests/cxx_group.cpp
bourtemb: > No, the device is defined only in the first (original) system (in `tango_cs` container).
> The second one (`tango_cs2` container) is empty. 
> The devices are created by `conf_devtest` program invoked from `pre_test.sh` script just once, with TANGO_HOST pointing to the original system.

Good. I missed that part indeed.

> You can also verify this using the example provided in #495 after running `make start-tango`. It is failing without the correction

This I already did and indeed #495 seems to be solved with this PR.

cpp_test_suite/environment/setup.sh
t-b: @mliszcz Just push the change out into a separate PR so we can get the easy changes merged fast.

mliszcz: @t-b here is a simple enough PR with just this kind of changes: https://github.com/tango-controls/cppTango/pull/557
But let's ignore that one for now and focus on current one.

cppapi/client/group.cpp
t-b: Yes using a stream, e.g. TangoSys_OMemStream is better than having another buffer overflow possibility.

t-b: And then the function can also take std::string arguments.

cppapi/client/group.cpp
```diff
@@ -48,63 +49,109 @@ bool GroupReply::exception_enabled = false;
...
+
+    DeviceNames device_names;
+    device_names.reserve(db_data.size());
+    db_data >> device_names;
+    return add_host_and_port_to_device_names(device_names, db_host, db_port);
+}
+
+DeviceNames GroupElementFactory::add_host_and_port_to_device_names(
+    const DeviceNames& names,
+    std::string& host,
```
t-b: non-const reference argument.

cppapi/client/group.h
```diff
@@ -575,14 +573,24 @@ class GroupAttrReplyList : public std::vector<GroupAttrReply>
 //=============================================================================
 class GroupElementFactory
 {
-  friend class Group;
-
+public:
   //- instanciatethe GroupElement which name matches the specified pattern with the specified timeout
   //- timeout = -1 => do not change the timeout
   static GroupElements instanciate (const std::string& p, int tmo = -1);
```
t-b: tmo->timeout_millis.

cppapi/client/group.h
```diff
@@ -575,14 +573,24 @@ class GroupAttrReplyList : public std::vector<GroupAttrReply>
...
 
+  static GroupElements create_group_elements(const DeviceNames&, int timeout_millis);
+  static DeviceNames resolve_device_names(const std::string&);
+  static DeviceNames resolve_local_device_names(const std::string&);
+  static DeviceNames resolve_remote_device_names(
+    int db_port,
+    std::string& db_host,
+    std::string& name_or_pattern_without_host);
+  static DeviceNames add_host_and_port_to_device_names(const DeviceNames&, std::string& db_host, int db_port);
+  static std::string build_full_device_name(const char*, int, const char*);
```
t-b: Please don't skip the names of the parameters. They should also match the ones in the implementation.

cpp_test_suite/new_tests/cxx_group.cpp
```diff
@@ -827,6 +827,39 @@ class GroupTestSuite: public CxxTest::TestSuite
...
+
+		GroupCmdReplyList command_results = group.command_inout("State");
+		GroupCmdReply command_result = command_results[0];
+
+		TS_ASSERT(not command_results.has_failed());
```
t-b: Question:
Do we want to use alternative operator names like `not` in tango? See https://en.cppreference.com/w/cpp/language/operator_alternative.

cppapi/client/group.cpp
```diff
@@ -48,63 +49,109 @@ bool GroupReply::exception_enabled = false;
...
+        std::string name_or_pattern_without_host = "";
+        parse_name(name_or_pattern, db_host, db_port, name_or_pattern_without_host);
+
+        const bool is_local_device = db_host.empty();
+
+        if (is_local_device)
+        {
+            return resolve_local_device_names(name_or_pattern);
+        }
+        else
```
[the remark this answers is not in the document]
mliszcz: This will look like a "guard" expression then:
```cpp
if (some_precondition)
{
    return handle_special_case()
}
work_as_usual()
```
If both branches are equally important and equally likely to happen I like to keep the `else`. See https://www.fluentcpp.com/2018/02/20/on-using-guards-in-cpp/, especially "Don’t sacrifice symmetry to reduce nesting" section for a deeper explanation.

cpp_test_suite/new_tests/cxx_group.cpp
bourtemb: This can cause some compilation errors when using some old compilers on Windows. But since we don't compile the tests on Windows, this should not be an issue here.
In this specific case, the code is even more clear when using this alternative operator.

cppapi/client/group.cpp
```diff
@@ -48,63 +49,109 @@ bool GroupReply::exception_enabled = false;
...
+    DeviceNames device_names;
+    device_names.reserve(db_data.size());
+    db_data >> device_names;
+    return device_names;
+}
+
+DeviceNames GroupElementFactory::resolve_remote_device_names(
+    int db_port,
+    std::string& db_host,
+    std::string& name_or_pattern_without_host)
```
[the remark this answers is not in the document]
mliszcz: I don't get it. Could you please clarify what is wrong here and what should I do?

cppapi/client/group.cpp
mliszcz: I've updated added a commit to address this issue. Could you please check?

As for the:
> And then the function can also take std::string arguments.

I plan to move this function to some common file in another PR, because building full name is repeated in a few places. It's more generic now. I prefer `const char*` over `std::string` if possible. You can convert `std::string` to `const char*` at no runtime cost but it's not possible the other way around.

cppapi/client/group.h
mliszcz: Thanks! Done.

cppapi/client/group.h
mliszcz: Done.

cpp_test_suite/new_tests/cxx_group.cpp
mliszcz: @t-b are you ok with leaving it as it is now?

cppapi/client/group.cpp
bourtemb: > I've updated added a commit to address this issue. Could you please check?

This commit https://github.com/tango-controls/cppTango/pull/608/commits/35a547dee8e421ab8f0a7dff13e924a37d942c6f looks good to me.
Using TangoSys_OMemStream made sense a while ago it seems according to the documentation (See the explanations for Line 7 under the code sample in this documentation section https://tango-controls.readthedocs.io/en/latest/development/device-api/device-server-writing.html#the-methods-used-for-the-devreaddirection-command).
I don't think we still have this issue because std::ostringstream is used directly in many other parts of the code.

mliszcz: Yes, currently `TangoSys_OMemStream` is always `std::ostringstream`. I see no benefit in using it any longer.

cppapi/client/group.cpp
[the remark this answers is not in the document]
mliszcz: Won't work. Of course I agree with your comment, but this function calls other functions that require (probably with no particular reason) non-const references.

```
ApiUtil.h:
	int	get_db_ind(std::string &host,int port);

Database.h:
	DbDatum get_device_exported(std::string &filter);
```

I don't want to change them in this PR. I think this is part of a bigger problem with C++ Tango where `const` is missing in most of the public API functions. We should have a separate issue for that.

Would it be ok for you if I open an separate issue and drop this comment?

cppapi/client/group.cpp
mliszcz: I've changed the code as you suggested. Please check if it's ok now.

cppapi/client/group.cpp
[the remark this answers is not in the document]
t-b: @mliszcz Okay, granted this is not doable here.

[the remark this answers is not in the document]
mliszcz: Thanks! I've raised #622 to track this.

mliszcz:
@bourtemb could you please check if the changes are ok for you and can be merged?

mliszcz:
@bourtemb is the PR ok from your POV?

cpp_test_suite/new_tests/cxx_group.cpp
```diff
@@ -827,6 +827,39 @@ class GroupTestSuite: public CxxTest::TestSuite
...
 	}
 
+	/* Verifies that a group can contain devices from a remote TANGO_HOST
+	 * (a Tango instance different from client's default TANGO_HOST).
+	 * An issue was reported when resolving names containing wildcards. */
+
+	void test_use_devices_from_remote_tango_host()
+	{
+		const std::string original_tango_host = std::getenv("TANGO_HOST");
+		const std::string external_tango_host = std::getenv("TANGO_HOST2");
```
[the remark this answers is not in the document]
mliszcz: Thanks!

> This will throw an std::logic_error exception ("basic:string::_M_construct null not valid") if TANGO_HOST or TANGO_HOST2 environment variable is not defined.

I'd say this is exactly what we want. All I can do is to re-throw the exception with a better message. But this will come at the expense of test case readability (extra logic and conditions not important to the scenario). In my opinion latter is more important for developers.

Still it would be nice to get the backtrace, but this is something to be handled on the CxxTest side.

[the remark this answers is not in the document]
bourtemb: > I'd say this is exactly what we want. All I can do is to re-throw the exception with a better message. But this will come at the expense of test case readability (extra logic and conditions not important to the scenario). In my opinion latter is more important for developers.

I agree with you and this is why I approved this PR. I think we don't need to do anything in this specific case.

PR #612, opened 2019-10-30
Compile with -Og option in Debug mode

bourtemb:
when this option is supported by the compiler
This is needed to use [abi-compliance-checker](https://lvc.github.io/abi-compliance-checker) tool to check the binary
compatibility between 2 library releases.


t-b:
The test runs will be really slow with `-Og`. Maybe we should have a CMAKE switch for enabling this flag?

bourtemb:
We get a strange error on Travis with gcc-latest:

```
[ 15%] Building CXX object cppapi/client/CMakeFiles/client_objects.dir/lockthread.cpp.o
cd /home/tango/src/build/cppapi/client && /usr/local/bin/c++  -D_REENTRANT -I/home/tango/include -I/home/tango/src/cppapi/client -I/home/tango/src/cppapi/client/helpers -I/home/tango/src/cppapi/server -I/home/tango/src/cppapi/server/idl -I/home/tango/src/log4tango/include -I/home/tango/src/build/cppapi/server -I/home/tango/src/build/log4tango/include  -std=c++14 -g   -Og -Wall -Wextra -pedantic -Werror -fPIC -Winvalid-pch -include /home/tango/src/build/cppapi/server/tango.h -o CMakeFiles/client_objects.dir/lockthread.cpp.o -c /home/tango/src/cppapi/client/lockthread.cpp
/home/tango/src/cppapi/client/lockthread.cpp: In member function 'virtual void Tango::LockThread::run(void*)':
/home/tango/src/cppapi/client/lockthread.cpp:133:22: error: 'cmd_received' may be used uninitialized in this function [-Werror=maybe-uninitialized]
  133 |    compute_sleep_time(cmd_received);
      |    ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [cppapi/client/CMakeFiles/client_objects.dir/build.make:391: cppapi/client/CMakeFiles/client_objects.dir/lockthread.cpp.o] Error 1
make[2]: Leaving directory '/home/tango/src/build'
make[1]: *** [CMakeFiles/Makefile2:1409: cppapi/client/CMakeFiles/client_objects.dir/all] Error 2
make[1]: Leaving directory '/home/tango/src/build'
make: *** [Makefile:166: all] Error 2
make: Leaving directory '/home/tango/src/build'
```

cppapi/client/lockthread.cpp has not changed...
The only apparent difference is this -Og flag  instead of -O0

bourtemb:
> The test runs will be really slow with `-Og`. Maybe we should have a CMAKE switch for enabling this flag?

Indeed, compilation is slower.
When reading the [gcc release notes](https://gcc.gnu.org/gcc-4.8/changes.html) and [documentation](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#Optimize-Options), it sounds like a good idea to use this -Og option when compiling the library in Debug mode to offer a better debugging experience to our users.
What do you think?

> It addresses the need for fast compilation and a superior debugging experience while providing a reasonable level of runtime performance. Overall experience for development should be better than the default optimization level -O0.

> Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0. 

It looks like the fast compilation they are talking about is not that fast indeed.

t-b:
@bourtemb Yes looks like `-Og` is better suited for debugging. I'm currently compiling with gcc 9.2.0 to see the list of additional warnings.

t-b:
[output-gcc-cppTango.txt](https://github.com/tango-controls/cppTango/files/3789499/output-gcc-cppTango.txt) all of them are `-Wmaybe-uninitialized`.

mliszcz:
@bourtemb  and @t-b 
I've had a quick look and since `-Og` is basically `-O1` with some optimizations disabled, I was able to reproduce the issue with `-O1 -fno-inline-functions-called-once`,

I the error pasted by Reynald I suspect that by default the compiler is inlining `get_command`
```cpp
			if (sleep != 0)
				received = get_command(sleep);
			else
				received = LOCK_TIME_OUT;
```

So that the optimizer sees that `received` is always either `LOCK_TIME_OUT` or `LOCK_COMMAND`.

In this case, `switch` is exhaustive and `cmd_received` is always initialized:
```cpp
			switch (received)
			{
			case LOCK_COMMAND:
				execute_cmd();
				cmd_received = true;
				break;

			case LOCK_TIME_OUT:
				one_more_lock();
				cmd_received = false;
				break;
			}

			compute_sleep_time(cmd_received);
```

If you don't inline `get_command`, value of `received` can be arbitrary integer, hence the warning.

t-b:
[output-clang-cppTango.txt](https://github.com/tango-controls/cppTango/files/3789765/output-clang-cppTango.txt).

t-b:
@mliszcz Yes the compiler is too stupid to figure out that the switch is exhaustive.

@bourtemb If you want to proceed now I would disable `-Wmaybe-uninitialized` for gcc and `-Wsometimes-uninitialized` for clang in debug mode. We can then fix these in a follow up PR.

t-b:
@bourtemb Your example is correct. That was one of the reasons why "enum class" was invented for C++11. But for the warning here in tango the compiler could now that get_command always returns a valid enum.

mliszcz:
Hi
@bourtemb What you have in your example is valid c++ code. If you are interested in standard's wording, you may check this SO answer: https://stackoverflow.com/a/39455421

@t-b I think GCC decided not to analyze function call in this case for performance reasons. In general you don't know how much code (in nested functions) you will need to analyze to find all possible values. This will increase compile time. And even then you may find a call to a function which is defined in another translation unit or shared library ...

Anyway, in my opinion the issue is on cppTango side and the variables should be initialized upon definition.

As for the PR, I second what Thomas proposed - we can have a dedicated CMake flag as this will be needed probably for just one job - the ABI checker.

bourtemb:
> Anyway, in my opinion the issue is on cppTango side and the variables should be initialized upon definition.

I agree.

> As for the PR, I second what Thomas proposed - we can have a dedicated CMake flag as this will be needed probably for just one job - the ABI checker.

From what I understood from the gcc documentation. -Og is recommended over -O0 for debugging purposes, so I would rather use this all the time (when supported by the used gcc version) in DEBUG mode to improve the debugging experience, not only for the ABI checker.

t-b:
@bourtemb I've fixed the gcc warnings in your branch.

t-b:
And I now fixed the last warning.

bourtemb:
> And I now fixed the last warning.

Thanks @t-b! It looks like there are still some remaining warnings.

t-b:
@bourtemb Should be fixed.

t-b:
So the IDL patches only work with the most recent omniidl.

So we have a couple of possibilities here:

1. Only do the patching for recent omniidl versions
2. Ignore the warning for these source files alltogether and remove the patching
3. Only do the patching when WARNING_AS_ERRORS is active and assume that this uses a latest omniidl

I actually favour two now.

bourtemb:
bb8a0857d83fe4881b97c54d30b87ab7e6546893 does not seem to compile on Debian7, 8 and 9.

t-b:
@bourtemb Yes, see https://github.com/tango-controls/cppTango/pull/612#issuecomment-553498567.

bourtemb:
> So the IDL patches only work with the most recent omniidl.
> 
> So we have a couple of possibilities here:
> 
>     1. Only do the patching for recent omniidl versions
> 
>     2. Ignore the warning for these source files alltogether and remove the patching
> 
>     3. Only do the patching when WARNING_AS_ERRORS is active and assume that this uses a latest omniidl
> 
> 
> I actually favour two now.

I agree with you guys. I would go for solution 2 too.

t-b:
@bourtemb Finally fixed.

t-b:
Huzzah!

cppapi/server/device_4.cpp
```diff
@@ -40,7 +40,28 @@
...
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#endif
+
+#if defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wsometimes-uninitialized"
+#pragma clang diagnostic ignored "-Wunused-variable"
+#endif
+
```
bourtemb: There are many files including tango.h header file.
Only cppapi/server/device_4.cpp and cppapi/server/device_5.cpp got a patch like this.
cppapi/server/device.cpp includes tango.h too for instance.
@t-b , Do you know why only cppapi/server/device_4.cpp and cppapi/server/device_5.cpp needed this patch?
This is not clear to me.

t-b: @bourtemb I presume because only these two cpp files use the code which triggers the warnings.

t-b: @bourtemb You can't review and approve because you opened the PR. Please merge if it is good.

bourtemb: Github also gives me the possibility to review...

This PR is fixing all the warnings when using the latest gcc but it is also introducing the following warning for Debian < 10:

```
/home/tango/src/cpp_test_suite/cxxtest/include/../../new_tests/cxx_old_poll.cpp:906:32: warning: unknown option after '#pragma GCC diagnostic' kind [-Wpragmas]

 #pragma GCC diagnostic ignored "-Wcatch-value"

                                ^~~~~~~~~~~~~~~
```

This is not a critical issue of course. There are some other warnings reported in these old Debian releases.

When testing with CLion, I noticed there are currently some inconsistencies in the values of the CMAKE_BUILD_TYPE CMake variable.
There are several locations in the cppTango CMakefiles where CMAKE_BUILD_TYPE value is compared to "DEBUG" or "RELEASE" values instead of comparing it to "Debug" or "Release".
CMAKE_BUILD_TYPE is even set to "RELEASE" by default.
I think this should be changed to use "Release" and "Debug" values, which are actually some of the possible values documented in the official CMake documentation and used by CLion:
https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html
This issue can be observed in the Travis builds where CMAKE_BUILD_TYPE is set to "Release".
As you can see, -Og option is used when it should not be in this use case:
https://travis-ci.org/tango-controls/cppTango/jobs/612484597

bourtemb: https://github.com/bourtemb/cppTango/commit/923ecfeeb6cea6d581faba598dbd9423d63b64f8 commit in my fork should solve these CMAKE_BUILD_TYPE inconsistencies issues.
I can create a dedicated PR for that.

t-b: @bourtemb Yes old gcc compilers (< 8.0) don't know about that "Wcatch-value". I could add some version magic and I'm also wondering if it is worth the trouble.

PR #615, opened 2019-10-31
configure/test_cppzmq_disconnect.cpp: Be compatible with older zmq.hp…

t-b:
…p versions as well

We used to ship a version of zmg.hpp with cppTango, see 5144cb80 (Add
again zmq.hpp before tagging release 9.2.5 to be consistent with 9.2.5
version from SVN Sourceforge (#348), 2017-01-11).

And in that version there was no default constructor for zmq::context_t
available. The constructor taking an integer parameter is still
available in newer versions so we can just use that here.

### DIFF
--- a/configure/test_cppzmq_disconnect.cpp
+++ b/configure/test_cppzmq_disconnect.cpp
@@ -2,7 +2,7 @@
 
 int main(int, char**)
 {
-  zmq::context_t c;
+  zmq::context_t c(1);
   zmq::socket_t s(c, ZMQ_REQ);
   s.disconnect("some endpoint");
 }


configure/test_cppzmq_disconnect.cpp
```diff
@@ -2,7 +2,7 @@
 
 int main(int, char**)
 {
-  zmq::context_t c;
+  zmq::context_t c(1);
```
mliszcz: Magic number. I suggest to use `ZMQ_IO_THREADS_DFLT` from `libzmq`'s `zmq.h`.

t-b: @mliszcz And fail if that define does not exist anymore? That does not sound clever.

bourtemb: This piece of code is there only to check whether the disconnect method is supported by the used cppzmq version. We will just try to compile this code. We will never execute it. What we want here is just to create a zmq::context_t object using a constructor which is present in recent cppzmq versions and present as well in older versions.
What we need is just an int parameter, whatever the value in order to compile.
So I would keep it simple and avoid using some Constant value which might not be defined.

mliszcz: How about adding a constant:
```cpp
const int num_of_io_threads = 1;
zmq::context_t c(num_of_io_threads);
```
My concern is that the reader may not know what this `1` means if he is not using cppzmq on a daily basis. For instance, my impression was that it was some kind of context id.

mliszcz: @bourtemb @t-b sorry! I've just noticed it's a cmake check. You can ignore my previous comment. I'm ok with leaving magic number here.

PR #616, opened 2019-11-01
appveyor.yml: Add inline deployment on tag push

### LINKED ISSUE #602: appveyor: Add support for automatic uploading of releases to github
> @bourtemb Regarding "Add Windows installers as assets to the GitHub release": Is that currently done with a script (where is that?) or if not, is it worth having one?

I think it was done manually the last time. Automating this process would be much better of course.

> Appveyor does also support that out of the box. See https://www.appveyor.com/docs/deployment/github/.

Good idea. I think we should try that.

_Originally posted by @bourtemb in https://github.com/tango-controls/cppTango/issues/594#issuecomment-545540317_

t-b:
Automate the deployment from appveyor.

We want to publish all artifacts as a new draft prelease on tag push
only.

I've published 9.3.4-rc1 from an environment in appveyor. This is an alternative solution to the inline deployment method.

For playing around I've used an OAuth token from my user. But we can not merge it as it is as using a token from a human user account is unsafe. This OAuth token allows read/write access to all repositories.

I propose to create a machine user which has only access to the cppTango repo and generate an OAuth token for her and use that here. See https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users and https://help.github.com/en/github/authenticating-to-github/creating-a-personal-access-token-for-the-command-line?query=machine%20user.

Close #602.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -195,3 +195,20 @@ artifacts:
   - path: ./*.exe
   - path: ./*.zip
   - path: ./*.msi
+
+deploy:
+- provider: GitHub
+  tag: $(APPVEYOR_REPO_TAG_NAME)
+  description: "Release $(APPVEYOR_REPO_TAG_NAME)"
+  auth_token:
+    # User: cppTango-bot
+    # Personal Access token with repo scope, encrypted via https://ci.appveyor.com/tools/encrypt
+    secure: KnvPAbdhmiTQxTLOGq7nyNs6d1JYbKVKspPgigzQ3fDrI+nkaxqWCw6UmRkS2Gqe
+  artifact: /.*/
+  repository: tango-controls/cppTango
+  draft: true
+  prerelease: true
+  force_update: true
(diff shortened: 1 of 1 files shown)


bourtemb:
> I propose to create a machine user which has only access to the cppTango repo and generate an OAuth token for her and use that here.

Sounds good to me.

t-b:
Updated. @bourtemb Could you grant the new machine user https://github.com/cppTango-bot access to the cppTango repo (and only to that, not to the full orga).

I've already created  a personal access token for her.

bourtemb:
> Updated. @bourtemb Could you grant the new machine user https://github.com/cppTango-bot access to the cppTango repo (and only to that, not to the full orga).

@t-b , invitation sent to cppTango-bot.
Please tell me if the permissions are not correct.

t-b:
@bourtemb Can you bend the rules and merge with just one approval?

PR #624, opened 2019-11-13
Cmake cleanups

### LINKED ISSUE #619: Don't create and compile IDL files in the source tree
The source is currently tainted by

```
$LC_ALL=C git clean -fdxn
Would remove server/idl/tango.h
Would remove server/idl/tangoDynSK.cpp
Would remove server/idl/tangoSK.cpp
```

this should be done in the build directory.

t-b:
Close #619.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,7 @@
 cmake_minimum_required(VERSION 2.8.12)
+
+project(cppTango)
+
 include(CTest)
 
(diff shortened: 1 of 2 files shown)


bourtemb:
Sorry! I think I made a mistake while testing this PR... It seems to work as expected...

t-b:
@bourtemb Yes, these files should already be created in the binary directory.

PR #637, opened 2019-12-09
Document contributing process

t-b:
- [x] Adapt the repo setting to something like 
grafik. @bourtemb I think this is one for you, as I'm not mighty enough.

The CODEOWNERS file is used for automatically requesting a review from the given users. If someone want's to see this list changed now, I'm open to any discussion. I've just used the minimal version suggested at the last kernel telco.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,14 @@
 
 # TANGO distributed control system - shared library
 
-TANGO is an object-oriented distributed control system. In TANGO all objects are representations of devices, which can be on the same computer or distributed over a network. Communication between devices can be synchronous, asynchronous or event driven.
+TANGO is an object-oriented distributed control system. In TANGO all objects
+are representations of devices, which can be on the same computer or
+distributed over a network. Communication between devices can be synchronous,
+asynchronous or event driven.
 
-The object model in TANGO supports methods, attributes and properties. TANGO provides an API which hides all the details of network access and provides object browsing, discovery, and security features.
+The object model in TANGO supports methods, attributes and properties. TANGO
+provides an API which hides all the details of network access and provides
+object browsing, discovery, and security features.
 
 For more information please visit [www.tango-controls.org](http://www.tango-controls.org).
 
--- a/README.md
+++ b/README.md
@@ -64,21 +64,26 @@ See [INSTALL file](INSTALL.md).
 
 See [corresponding wiki page](https://github.com/tango-controls/cppTango/wiki/Contribution-Guide)
 
-# Usage workflow (draft)
+# Pull request acceptance and merging
 
+You have created a change to cppTango. 🎉
 
-1. In its most basic form, we adopt the githubflow : the default branch is master, which is always in a deployable state (i.e. must never be broken); work on new features is done using temporary feature branches (which may live in the same repo or in a fork)
-2. Reporting issues: use github issues
-3. Code contributions: use Pull requests. Pull requests can be associated with issues. Trivial fixes can even be done from the web (a temporary branch is created automatically to use the Pull request infrastructure)
-4. Code review: most projects (all except single-developer ones) should only allow commits to the master branch after peer review. This can be enforced by the convention that all commits to master must be done via a Pull request and the pull request approved by a person different from the author of the commit.
-5. Releases (named versions):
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,16 +30,22 @@
...
 
 For more information please visit [www.tango-controls.org](http://www.tango-controls.org).
 
-# cppTango - tango-9-lts branch
-This repository is the result of svn2git migration of Tango source code previously hosted on tango-cs Sourceforge project. It is combined from api/cppapi, api/log4tango, api/cpp_test_suite anf api/cpp_test_ds
+# Important branches
 
-cppTango tango-9-lts branch is the branch dedicated to Tango 9 Long Term Support version.
+- tango-9-lts: Tango 9 Long Term Support (LTS) version
+- 9.3-backports: Last stable LTS version
```
bourtemb: - tango-9-lts: Tango 9 Long Term Support (LTS) development
- 9.3-backports: Tango 9.3 binary compatible LTS version development

README.md
```diff
@@ -58,21 +64,26 @@ See [INSTALL file](INSTALL.md).
...
+# Pull request acceptance and merging
+
+You have created a change to cppTango. 🎉
+
+And now you want to get these changes merged? Very nice!
+
+In order to give you the best possible experience here are a few hints for the
+path forward:
+- All CI tests have to pass. If you have changed the behaviour of the code, you
+  should add new tests as well. You don't need to exectute the tests locally,
```
bourtemb: exectute  => execute

README.md
```diff
@@ -58,21 +64,26 @@ See [INSTALL file](INSTALL.md).
...
+  CI is the reference anyway. So just create a PR and let CI handle that.
+- Make your PR easy to review. This starts with explaining what it wants to
+  achieve and ends with splitting the changes into logical commits where each
+  commit describe why they are changing the code.
```
bourtemb: describe why they are => describes why it is

README.md
```diff
@@ -58,21 +64,26 @@ See [INSTALL file](INSTALL.md).
...
+- Follow the coding style. This is at the moment messy at best, but still
+  we don't want to get worse.
```
bourtemb: we don't want to get worse =>  we don't want it to get worse

README.md
```diff
@@ -58,21 +64,26 @@ See [INSTALL file](INSTALL.md).
...
+- Your PR needs approval from two of the code owners listed [here](https://github.com/tango-controls/cppTango/blob/tango-9-lts/CodeOwners).
+- Be prepared to adapt your pull request to the review responses. Code review
+  is done for ensuring a higher code quality and communicating implementations
+  details to newcomers and not for annoying anyone or slowing done development.
```
bourtemb: slowing done => slowing down

bourtemb:
By the way, in the github settings you are suggesting, with the suggested CODEOWNER file, does it require an approval of @t-b **and** @bourtemb or having only one approval from @t-b **or** @bourtemb is enough as long as there are at least 2 approvals (from any reviewer) in total?

t-b:
@bourtemb Thanks for the feedback. I've addressed the typos.

> By the way, in the github settings you are suggesting, with the suggested CODEOWNER file, does it require an approval of @t-b and @bourtemb or having only one approval from @t-b or @bourtemb is enough as long as there are at least 2 approvals (from any reviewer) in total?

The way I understand it, it is the latter. Two approvals in total were one needs to be from a codeowner.

bourtemb:
> The way I understand it, it is the latter. Two approvals in total were one needs to be from a codeowner.

Which is nice because then we wouldn't be blocked if one of the codeowners is not available.

bourtemb:
We (@andygotz and myself) propose to add @mliszcz as well in the codeowners file.

t-b:
@bourtemb Added.

bourtemb:
I think we should try to use the command `[skip ci] ` in our commit messages when the commit is only affecting the documentation and cannot have an impact on the tests.
It looks like travis and appveyor are both following this convention:
https://www.appveyor.com/docs/how-to/filtering-commits/
https://docs.travis-ci.com/user/customizing-the-build/#skipping-a-build

bourtemb:
I changed the github settings for tango-9-lts branch.
We should do something similar in 9.3-backports branch (update README, add CODEOWNERS and then add a github rule for that branch too).

t-b:
@bourtemb See https://github.com/tango-controls/cppTango/pull/639.

PR #638, opened 2019-12-10
Do not require "sed" to enhance generated files

ltjax:
This makes it a lot easier to build from source on windows.


t-b:
@ltjax Thanks for the PR.

It is definitly a good thing to get rid of sed here. The new patching seems trigger a compilation error though, see https://travis-ci.org/tango-controls/cppTango/jobs/623197049?utm_medium=notification&utm_source=github_status.

And unfortunately your PR crossed development with #624 which just got merged. So please rebase your changes here.

ltjax:
Yea, I saw it failing on linux. I'll look into it!

bourtemb:
Generated file is different from what we had before.
Some \n are still missing for the AttributeValue_5 and DevPipeData cases
AttributeValue_4 case seems to be ok

ltjax:
> Generated file is different from what we had before.
> Some \n are still missing for the AttributeValue_5 and DevPipeData cases
> AttributeValue_4 case seems to be ok

You're right. I forgot those when fiddling around with escaping in CMake. Added them.

PR #642, opened 2019-12-17
Allow different paths for CPPZMQ and ZMQ

ltjax:
On Windows. Was previously assumed to be equal. Needed this because those dependencies sit in different folders for my build.


configure/cmake_win.cmake
```diff
@@ -162,8 +162,8 @@ install(DIRECTORY $ENV{OMNI_BASE}/include/omniVms DESTINATION include COMPONENT
 install(FILES $ENV{OMNI_BASE}/include/omniconfig.h DESTINATION include COMPONENT headers)
 install(FILES $ENV{OMNI_BASE}/include/omnithread.h DESTINATION include COMPONENT headers)
 install(FILES $ENV{ZMQ_BASE}/include/zmq.h DESTINATION include COMPONENT headers)
-install(FILES $ENV{ZMQ_BASE}/include/zmq.hpp DESTINATION include COMPONENT headers)
-install(FILES $ENV{ZMQ_BASE}/include/zmq_addon.hpp DESTINATION include COMPONENT headers)
+install(FILES $ENV{CPPZMQ_BASE}/include/zmq.hpp DESTINATION include COMPONENT headers)
```
mliszcz: I like the idea of separating zmq and cppzmq, but wouldn't this break other people's builds if they do not set CPPZMQ_BASE? Would it be possible to fall-back to using ZMQ_BASE if CPPZMQ_BASE is not defined? - just a suggestion, I have no experience in windows builds and do not know if this will be a problem.

ltjax: Yes. And yes, it is possible. IMO, let them break. This was broken to begin with, and the CMake system is not very old yet. To be honest, the whole build system right now is super unportable and seems to be tailored towards one specific build scenario only.
The problem of _not being able to build the software_ is better remedied with a **manual** or an example. This example can be the appveyor build, or if you want, my conan build. You should not already stack compatibility code on this to keep a broken solution running.

mliszcz: Fine with me, but let's wait for comments from people using windows.

bourtemb: > Yes. And yes, it is possible. IMO, let them break. This was broken to begin with, and the CMake system is not very old yet. To be honest, the whole build system right now is super unportable and seems to be tailored towards one specific build scenario only.

You're right, it was tailored for appveyor builds with the original goal to generate some zip files and Windows installers containing the Tango library and include files and all the dependencies (lib and include files too).
I think if we really want to help our users to use CMake to build cppTango themselves on Windows, then we need to change several things.
One of them being to stop installing the dependencies in the cppTango CMakefiles as you suggested.
On Linux, we don't provide any longer zmq*.hpp files for instance. We should probably do the same on Windows by default too. This could be an option?
As you suggested, @ltjax, this dependencies installation could be moved to a separate CMake script, for instance.

Using conan might also simplify our life... I'm not familiar with it yet.
Appveyor could also probably be updated to use conan and to create conan packages, I guess you would be interested in that?

It would be interesting to get some feedback from our users on Windows.

ltjax: Thanks for the explanation!
Conan really streamlines dependency handling, and I recommend getting into it.
Yes, I am interested in getting binary deployment like that working on appveyor, but I fear I might need a little more budget from the guys at KARA/ANKA (or you?) to do that. For now, my focus is to get one old windows tango device up and running again, and the working recipe is almost as good for that as a full binary release.

appveyor.yml
```diff
@@ -259,16 +259,17 @@ install:
...
   - cmd: set ZMQ_BASE=C:/projects/zeromq/
+  - cmd: set CPPZMQ_BASE=C:/projects/zeromq/
   - cmd: set IDL_BASE=%IDL_BIN%
   - cmd: set OMNI_BASE=C:/projects/omniorb/
   - cmd: set PTHREAD_WIN=C:/projects/pthreads-win32/
   - cmd: if not defined USE_PCH set USE_PCH=OFF
   #- cmd: cmake -G "%CMAKE_GENERATOR%" -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE -DBUILD_SHARED_LIBS=TRUE .
   #- cmd: cmake -G "%CMAKE_GENERATOR%" -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE .
-  - cmd: cmake -G "%CMAKE_GENERATOR%" -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE -DIDL_BASE="%IDL_BASE%" -DOMNI_BASE="%OMNI_BASE%" -DZMQ_BASE="%ZMQ_BASE%" -DPTHREAD_WIN=%PTHREAD_WIN% -DUSE_PCH=%USE_PCH% .
+  - cmd: cmake -G "%CMAKE_GENERATOR%" -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE -DIDL_BASE="%IDL_BASE%" -DOMNI_BASE="%OMNI_BASE%" -DZMQ_BASE="%ZMQ_BASE%" -DCPPZMQ_BASE="%ZMQ_BASE%" -DPTHREAD_WIN=%PTHREAD_WIN% -DUSE_PCH=%USE_PCH% .
```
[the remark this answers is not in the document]
t-b: I'm in favour of separating `ZMQ_BASE` and `CPPZMQ_BASE` as well. And I don't think we need to care about windows developers compiling on their own machines here. The only current user of the cmake compilation on windows is appveyor, so as long as that works, this is good.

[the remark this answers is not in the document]
t-b: @ltjax What do you think about the separation of `ZMQ_BASE` and `CPPZMQ_BASE`?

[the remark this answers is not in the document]
ltjax: Sure. Sorry it took me a while to make this change, I'm switching between like 10 computers and OSes :-)

configure/cmake_win.cmake
```diff
@@ -165,8 +165,8 @@ install(DIRECTORY $ENV{OMNI_BASE}/include/omniVms DESTINATION include COMPONENT
...
+install(FILES $ENV{CPPZMQ_BASE}/include/zmq_addon.hpp DESTINATION include COMPONENT headers)
```
bourtemb: There was this comment from @mliszcz at some point in this PR:

> 
> 
> I like the idea of separating zmq and cppzmq, but wouldn't this break other people's builds if they do not set CPPZMQ_BASE? Would it be possible to fall-back to using ZMQ_BASE if CPPZMQ_BASE is not defined? - just a suggestion, I have no experience in windows builds and do not know if this will be a problem.

What he was suggesting has been added in appveyor.yml but I think he also meant in the general case when compiling on Windows. 
What happens if CPPZMQ_BASE is not defined? If I understood well, I think he was suggesting to initialize CPPZMQ_BASE to ZMQ_BASE in this use case, but in the CMake file.

ltjax: Yes, I understood & answered that. My stance is still to let it break instead of of cluttering a pretty new, and mostly untested build system for backwards compatibility. TBH, with all the pain I had to go through to get it to compile, I would be pretty impressed with anyone else who build it on windows from scratch.

PR #651, opened 2020-01-10
fix build on PPC

lorenzopivetta:
Add support for cmake option -DPPC=true to disable mmx optimization and cpp_test_suite. Required for 9.3-backports.

### DIFF
--- a/configure/cmake_linux.cmake
+++ b/configure/cmake_linux.cmake
@@ -4,8 +4,10 @@ add_library(tango SHARED    $<TARGET_OBJECTS:log4tango_objects>
                             $<TARGET_OBJECTS:client_objects>
                             $<TARGET_OBJECTS:idl_objects>
                             $<TARGET_OBJECTS:jpeg_objects>
-                            $<TARGET_OBJECTS:jpeg_mmx_objects>
                             $<TARGET_OBJECTS:server_objects>)
+if (NOT PPC)
+    add_library(tango SHARED $<TARGET_OBJECTS:jpeg_mmx_objects>)
+endif (NOT PPC)
 target_link_libraries(tango PUBLIC ${ZMQ_PKG_LIBRARIES} ${OMNIORB_PKG_LIBRARIES} ${OMNICOS_PKG_LIBRARIES} ${OMNIDYN_PKG_LIBRARIES} ${CMAKE_DL_LIBS})
 target_compile_options(tango PRIVATE -fPIC)
 target_include_directories(tango PUBLIC ${ZMQ_PKG_INCLUDE_DIRS} ${OMNIORB_PKG_INCLUDE_DIRS} ${OMNIDYN_PKG_INCLUDE_DIRS})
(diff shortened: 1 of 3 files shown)


lorenzopivetta:
Argh. As it is does not work on X86... anyone can close this. I'll make a new one.

configure/cmake_linux.cmake
```diff
@@ -4,8 +4,10 @@ add_library(tango SHARED    $<TARGET_OBJECTS:log4tango_objects>
                             $<TARGET_OBJECTS:client_objects>
                             $<TARGET_OBJECTS:idl_objects>
                             $<TARGET_OBJECTS:jpeg_objects>
-                            $<TARGET_OBJECTS:jpeg_mmx_objects>
                             $<TARGET_OBJECTS:server_objects>)
+if (NOT PPC)
+    add_library(tango SHARED $<TARGET_OBJECTS:jpeg_mmx_objects>)
+endif (NOT PPC)
```
t-b: @lorenzopivetta 

You can not declare the tango target multiple times.

I think you want something like

```diff
git diff .
diff --git a/configure/cmake_linux.cmake b/configure/cmake_linux.cmake
index 945e2d19..23e86f14 100644
--- a/configure/cmake_linux.cmake
+++ b/configure/cmake_linux.cmake
@@ -4,7 +4,7 @@ add_library(tango SHARED    $<TARGET_OBJECTS:log4tango_objects>
                             $<TARGET_OBJECTS:client_objects>
                             $<TARGET_OBJECTS:idl_objects>
                             $<TARGET_OBJECTS:jpeg_objects>
-                            $<TARGET_OBJECTS:jpeg_mmx_objects>
+                            $<$<NOT:$<BOOL:${PPC}>>:$<TARGET_OBJECTS:jpeg_mmx_objects>>
                             $<TARGET_OBJECTS:server_objects>)
 target_link_libraries(tango PUBLIC ${ZMQ_PKG_LIBRARIES} ${OMNIORB_PKG_LIBRARIES} ${OMNICOS_PKG_LIBRARIES} ${OMNIDYN_PKG_LIBRARIES} ${CMAKE_DL_LIBS})
 target_compile_options(tango PRIVATE -fPIC)
@@ -21,7 +21,7 @@ add_library(tango-static STATIC $<TARGET_OBJECTS:log4tango_objects>
                                 $<TARGET_OBJECTS:client_objects>
                                 $<TARGET_OBJECTS:idl_objects>
                                 $<TARGET_OBJECTS:jpeg_objects>
-                                $<TARGET_OBJECTS:jpeg_mmx_objects>
+                                $<$<NOT:$<BOOL:${PPC}>>:$<TARGET_OBJECTS:jpeg_mmx_objects>>
                                 $<TARGET_OBJECTS:server_objects>)
 target_link_libraries(tango-static PUBLIC ${ZMQ_PKG_LIBRARIES} ${OMNIORB_PKG_LIBRARIES} ${OMNICOS_PKG_LIBRARIES} ${OMNIDYN_PKG_LIBRARIES} ${CMAKE_DL_LIBS})
 target_include_directories(tango-static PUBLIC ${ZMQ_PKG_INCLUDE_DIRS} ${OMNIORB_PKG_INCLUDE_DIRS} ${OMNIDYN_PKG_INCLUDE_DIRS})
```

t-b: And I also think that we should not use a self-defined symbol like `PPC` here, something like `TANGO_JPEG_MMX` (defaulting to true) would better suit the purpose.

lorenzopivetta: Hi @t-b , yes I realized it as soon as the tests failed. Need some cmake background ;-) !

lorenzopivetta: > And I also think that we should not use a self-defined symbol like `PPC` here, something like `TANGO_JPEG_MMX` (defaulting to true) would better suit the purpose.

Looking around, I'd put the default in 'configure/CMakeLists.txt', something like:
`if (NOT TANGO_JPEG_MMX)
    message("Building MMX support; disable with -DTANGO_JPEG_MMX=false")
    set(TANGO_JPEG_MMX true)
endif(NOT TANGO_JPEG_MMX)
`
Is it the right place?

PR #652, opened 2020-01-11
Travis: Add testing with ubuntu 20.04 (aka focal)

### LINKED ISSUE #649: Add testing on latest ubuntu-lts version to travis
ESRF is moving to ubuntu-lts so we need to test on that as well.

t-b:
Closes #649.


.travis/ubuntu-20.04/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM ubuntu:focal
```
mliszcz: This dockerfile is exactly the same as for debian 10 (and maybe others). Do you think if we can re-use debian 10 dockerfile here just by adding build arg? Like:
```dockerfile
ARG base=debian:buster
FROM ${base}
...
```

And then in CI script:
```bash
docker build --build-arg base="ubuntu:focal" ...
```

t-b: I can certainly chat that, but do you think the added complexity is worth the reduced duplication?

mliszcz: It was just a proposal. Sometimes we need to modify these files and it would be simpler just to edit one file instead. But we can have a separate PR for that if we decide that this is necessary. In my opinion, we can have multiple files as we don't change them very frequently.

t-b: @mliszcz Okay so I leave it as is for now.

PR #653, opened 2020-01-11
CMakeLists.txt: Require C++11 and cmake 3.7

### LINKED ISSUE #626: Raise required C++11 and cmake versions
As decided today the we will nowish require C++11 and cmake 3.7 on the tango-9-lts branch. A C++14 supported compiler will be required around 8/2020. In practical terms this means we will require at least gcc 4.8.1, [link](https://gcc.gnu.org/projects/cxx-status.html), or clang 3.3. [link](https://clang.llvm.org/cxx_status.html#cxx11).

For the supported linux distributions this means we will not build on wheezy anymore. The windows builds should not be affected, as MSVC 2015 already supports C++11, see https://devblogs.microsoft.com/cppblog/c111417-features-in-vs-2015-rtm/.

It would be nice to add some docker images for other required linux distributions (centos, ubuntu lts, ...) so that we can check that we at least build on these. But I need to know which versions I should add.

t-b:
- [x] Fix CI fallout
- [x] Remove C++ feature #ifdef handling in tango_config.h
- [x] Use cmake testing for `cxx_relaxed_constexpr`, see also https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES
- [x] Use cmake 3.7 on systems with too old cmake to ensure that it works with this version.
- [x] Output cmake version on travis/appveyor
- [x] Finalize

Closes #626.


mliszcz:
+1 for this PR, but I'm not sure what is the purpose of checking for relaxed `constexpr` support if we still need to be compatible with C++11. We are just doubling implementation effort this way (If we *need* certain computation to happen at compile time we still need to provide `#else` clause with template-based solution for C++11-only compilers).

t-b:
@mliszcz My idea is that we only use `relaxed constexpr` and not the C++11 one. The plan is to move to C++14 in august, so I don't think we need to have constexpr for C++11 and C++14.

cppapi/client/devapi_attr.tpp
```diff
@@ -34,9 +34,7 @@
 #ifndef _DEVAPI_ATTR_TPP
 #define _DEVAPI_ATTR_TPP
 
-#ifdef HAS_TYPE_TRAITS
 	#include <type_traits>
```
bourtemb: Indentation

cppapi/server/attrsetval.tpp
```diff
@@ -34,9 +34,7 @@
 #ifndef _ATTRSETVAL_TPP
 #define _ATTRSETVAL_TPP
 
-#ifdef HAS_TYPE_TRAITS
 	#include <type_traits>
```
bourtemb: indentation

cppapi/server/w_attrsetval.tpp
```diff
@@ -33,9 +33,7 @@
 #ifndef _WATTRSETVAL_TPP
 #define _WATTRSETVAL_TPP
 
-#ifdef HAS_TYPE_TRAITS
 	#include <type_traits>
```
bourtemb: indentation

cppapi/client/api_util.cpp
```diff
@@ -60,22 +58,15 @@ ApiUtil *ApiUtil::_instance = NULL;
...
 
 void _t_handler(TANGO_UNUSED(int signum))
 {
-#ifdef HAS_THREAD
     std::thread t(_killproc_);
     t.detach();
-#else
-    _KillProc_ *t = new _KillProc_;
-    t->start();
-#endif
```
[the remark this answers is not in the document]
t-b: Done.

t-b:
Indentation is fixed.

t-b:
@mliszcz @bourtemb Ready for review. I've not done any automated modernization using [clang-tidy/clang-modernize](https://clang.llvm.org/extra/clang-tidy/index.html) at this point as the PR is already big enough.

And I also did not want to step on everyones toes who has a PR open.

.travis/debian8/Dockerfile
```diff
@@ -11,15 +11,19 @@ RUN sed -i '/jessie-updates/d' /etc/apt/sources.list  # Now archived
...
   lsb-release                            \
   omniidl                                \
   libomniorb4-dev                        \
   libcos4-dev                            \
   libomnithread3-dev                     \
-  libzmq3-dev
+  libzmq3-dev                            \
+  wget
+
+RUN wget --no-check-certificate https://cmake.org/files/v3.10/cmake-3.10.0-Linux-x86_64.sh -O /tmp/cmake-install.sh \
```
bourtemb: Maybe we should install CMake 3.7 instead of 3.10 on Travis Debian 8 to be sure that it works fine with CMake 3.7?

configure/CMakeLists.txt
```diff
@@ -174,34 +192,14 @@ if(${FAILED})
...
     CHECK_CXX_COMPILER_FLAG("-Og" COMPILER_SUPPORTS_OG)
     if(COMPILER_SUPPORTS_CXX14)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
-    elseif(COMPILER_SUPPORTS_CXX11)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-    elseif(COMPILER_SUPPORTS_CXX0X)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x")
-    else()
-        message("The compiler ${CMAKE_CXX_COMPILER} has no C++11/14 support. You may not benefit from performance optimizations.")
     endif()
```
mliszcz: What `-std` flag will be used for compilers that do not support c++14, e.g. gcc 4.8.1)? If I recall, `-std=c++11` was never a default setting in gcc and if this is true, the compiler will run in c++98 mode, right? On Debian 8 we compile in c++14 mode so actually there is no job in CI that checks the c++11 mode.

bourtemb:
> What -std flag will be used for compilers that do not support c++14, e.g. gcc 4.8.1)? If I recall, -std=c++11 was never a default setting in gcc and if this is true, the compiler will run in c++98 mode, right? On Debian 8 we compile in c++14 mode so actually there is no job in CI that checks the c++11 mode.


I think 
`set(CXX_STANDARD_REQUIRED 11)`
instruction in CMakeLists.txt is ensuring the code is compiled with at least c++11 enabled.
https://cmake.org/cmake/help/v3.1/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD

configure/CMakeLists.txt
mliszcz: I think @bourtemb answered this question. Resolving.

.travis/debian8/Dockerfile
t-b: I orignally though about that as well but debian 9 has cmake 3.7.2, see https://packages.debian.org/stretch/cmake, which is close enough to 3.7 I think.

t-b:
I'm not 100% sure that `set(CXX_STANDARD_REQUIRED 11)` will really fail cmake if it the compiler does not understand c++11. But something like the check in #659 does enforce that. And that PR is planned to be merged after this one.

.travis/debian8/Dockerfile
bourtemb: > 
> 
> I orignally though about that as well but debian 9 has cmake 3.7.2, see https://packages.debian.org/stretch/cmake, which is close enough to 3.7 I think.

Fine by me.

cppapi/server/tango_config.h
```diff
@@ -414,17 +236,6 @@
...
-  #define TANGO_NORETURN
-#endif
-
-// C++14 style constexpr
-#ifdef HAS_CONSTEXPR
-  #define TANGO_CONSTEXPR constexpr
-#else
-  #define TANGO_CONSTEXPR
-#endif
+#define TANGO_NORETURN [[noreturn]]
```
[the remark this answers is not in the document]
t-b: @bourtemb I thought about that as well. So it is okay if the symbol `TANGO_NORETURN` disappears? In that case I would cleanup that whole and move the relevant rest file into tango_const.in.h, basically killing `tango_config.h` afterwards.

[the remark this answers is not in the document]
bourtemb: > So it is okay if the symbol `TANGO_NORETURN` disappears?

I think it's fine. TANGO_NORETURN was added by you in 2018 and is currently used only in except.h in cppTango code.
I don't think our users are aware of this symbol and are using it.

t-b:
@mliszcz I would merge this one after #568 as your PR is older.

t-b:
@mliszcz @bourtemb Rebased.

PR #657, opened 2020-01-16
Fix build on ppc 3

lorenzopivetta:
Add cmake option TANGO_JPEG_MMX to enable/disable MMX support (default true). Self-defined symbol PPC still required to disable building cpp_test_suite (does not compile on Ubuntu 7.10 PPC)


configure/CMakeLists.txt
```diff
@@ -210,5 +210,13 @@ if(NOT WIN32)
         endif()
     endif()
 endif()
+
+if (TANGO_JPEG_MMX STREQUAL "false")
+  message("MMX support disabled")
+else (TANGO_JPEG_MMX STREQUAL "false")
+  message("Building MMX support; use '-DTANGO_JPEG_MMX=false' to disable")
+  set(TANGO_JPEG_MMX true)
+endif (TANGO_JPEG_MMX STREQUAL "false")
```
mliszcz: I think it would be simpler to use CMake's option (and the usual ON/OFF values) here.
Please check the documentation: https://cmake.org/cmake/help/latest/command/option.html

See some examples:
https://github.com/tango-controls/cppTango/blob/3343f94d4151ab1dbb7eee7f5adb5cf48678db4d/CMakeLists.txt#L7-L9
https://github.com/tango-controls/cppTango/blob/3343f94d4151ab1dbb7eee7f5adb5cf48678db4d/configure/CMakeLists.txt#L235

bourtemb:
Travis CI failed on Debian 8 with the following error:
```
CMake Error at configure/cmake_linux.cmake:3 (add_library):

  Unrecognized generator expression:

    $<$<BOOL:true>:$<TARGET_OBJECTS:jpeg_mmx_objects>>

Call Stack (most recent call first):

  CMakeLists.txt:43 (include)

CMake Error at configure/cmake_linux.cmake:20 (add_library):

  Unrecognized generator expression:
```

So it looks like this PR is requiring a minimal CMake version which is more recent than the one provided on Debian 8 (3.0.2 I think).
So we need to agree on this minimal CMake version and update the following line in our main CMakeList.txt if we agree we should raise the minimum required CMake version:

https://github.com/tango-controls/cppTango/blob/tango-9-lts/CMakeLists.txt#L1

The build is passing on Debian 7 because CMake 3.10 is installed in our Docker Debian 7 because the default one was too old.

t-b:
@bourtemb That should not be a problem as #653 will require a newer cmake version. We just need to merge #653 first.

t-b:
@bourtemb According to https://cmake.org/cmake/help/v2.8.12/cmake.html it has bool support.

bourtemb:
I think it can be ok to require a more recent CMake version even for the 9.3-backports branch because it should still be possible to install more recent versions of CMake on old systems as we did when we installed CMake 3.10 on our Travis Debian 7 Docker.

configure/CMakeLists.txt
lorenzopivetta: Much better, indeed! I'll fix it Monday.

lorenzopivetta:
We had cmake 3.6.0 _compiled_ on those old Ubuntu 7.10 PPC

bourtemb:
> @bourtemb According to https://cmake.org/cmake/help/v2.8.12/cmake.html it has bool support.

I think the issue we are facing here is the lack of support for nested CMake generator expressions in old CMake versions.

configure/CMakeLists.txt
lorenzopivetta: Fixed in https://github.com/tango-controls/cppTango/pull/657/commits/acc962bcec595d1e28a0a6a7158e616e317d19d9. Tested on PPC and x86_64 (Ubuntu 18.04)

t-b:
We already have automake support for disabling jpegmmx, see https://github.com/tango-controls/TangoSourceDistribution/blob/9ba331bab50ff7d9f2ca47207080a0c71ff3c7ca/assets/configure.ac#L423

lorenzopivetta:
> We already have automake support for disabling jpegmmx, see https://github.com/tango-controls/TangoSourceDistribution/blob/9ba331bab50ff7d9f2ca47207080a0c71ff3c7ca/assets/configure.ac#L423

Yes. That used to work. But isn't automake been dropped in favor of cmake? Am I missing something?

t-b:
@lorenzopivetta For cppTango yes we only use cmake. But for the tango source distribution we still use automake (and don't cmake), but 9.3.4 will be the last tango source distribution release using automake.

lorenzopivetta:
> Okay for not having to deal with nested operator expressions here is what I would do:

@t-b thanks for looking at this. But, is the version with nested operators definitely dropped?
I'll try this one and let you know.

t-b:
nested operators works only on tango-9-lts branch as that allows cmake 3.7, 9.3-backports only allows cmake 2.8.12.

t-b:
@lorenzopivetta I think this PR is obsolete as we included that in #674 and #668. Please reopen if I'm wrong.

PR #659, opened 2020-01-21
Remove fallback snprintf implementation

t-b:
Since the default C++ version was raised to 11 we now know that we have a working snprintf solution.
Therefore we can drop the alternative implementation which was not touched in 18 years.


log4tango/config/config.cmake
```diff
@@ -56,6 +56,10 @@ LOG4TANGO_CHECK_COMPILER_FEATURE("config/check_namespace.cpp" NAMESPACES)
 #snprintf
 LOG4TANGO_CHECK_COMPILER_FEATURE("config/check_snprintf.cpp" SNPRINTF)
 
+if(NOT LOG4TANGO_HAVE_SNPRINTF)
+  message(FATAL_ERROR "snprintf could not be found, aborting.")
+endif()
```
mliszcz: Do we still need this check if presence of snprinf is guaranteed by the C++ standard required by tango (CXX_STANDARD_REQUIRED)?

t-b: You are right, let's remove the check for `snprintf`.

t-b: @mliszcz Updated.

PR #662, opened 2020-01-22
Automate abi api compliance check

### LINKED ISSUE #613: Automate abi compliance check
libzmq project has tried something to automate abi compliance checking process (recent travis builds are failing because of that though).
Maybe we could do something similar as what they did:
https://github.com/zeromq/libzmq/blob/02f7dca678095099be8774844fec693f390b2223/builds/abi-compliance-checker/ci_build.sh
https://github.com/zeromq/libzmq/blob/9bb6b2142385b78d47811b73745b9c21ec366106/.travis.yml#L136-L143

t-b:
- [x] Use gcc-latest
- [x] Rebase once https://github.com/tango-controls/cppTango/pull/668 is merged
- [x] Use github actions for executing the job and serving the artefact
Close #613.


t-b:
The test error is unrelated, the output looks like

```
Creating compatibility report ...

Binary compatibility: 100%

Source compatibility: 100%

Total binary compatibility problems: 0, warnings: 0

Total source compatibility problems: 0, warnings: 0

Report: compat_reports/libtango/85fbfd3aebddb629e7fcb8e15106c4add89e42b6_to_29fbad5248830849a04062bcdbc91ba99c18d397/compat_report.html
```

I've not yet found a good place for this kind of artefact. Ideas?

bourtemb:
> I've not yet found a good place for this kind of artifact. Ideas?

@t-b, as suggested by @mliszcz for another problem, the best would probably be to use [Github actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/persisting-workflow-data-using-artifacts) for this kind of job.

Github actions support artifacts.

t-b:
Rebased. Regarding the artefacts. Do we really want to move to github actions just because of the artifacts? According to [1] they only store artifacts for 90 days for PRs ("For pushes and pull requests, GitHub stores artifacts for 90 days.") and I don't think this is enough in general.

Bintray would be the obvious alternative. Otherwise I can probably also come up with free S3 storage.

[1]: https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts

bourtemb:
> Rebased. Regarding the artefacts. Do we really want to move to github actions just because of the artifacts? According to [1](https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts) they only store artifacts for 90 days for PRs ("For pushes and pull requests, GitHub stores artifacts for 90 days.") and I don't think this is enough in general.
> 
> Bintray would be the obvious alternative. Otherwise I can probably also come up with free S3 storage.

I don't think we need to keep the ABI compatibility reports for ages so for the use case we were talking in this issue it should be fine.
I think we will have a look at the report only when an incompatibility will have been reported. So hopefully 90 days should be enough and if not, I think we can re-run the github action or equivalent to generate it again?

t-b:
@bourtemb I'll give github actions a try for that.

configure/CMakeLists.txt
```diff
@@ -198,6 +198,7 @@ if(NOT WIN32)
     # throw specifications and these are not supported anymore.
     CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX14)
     CHECK_CXX_COMPILER_FLAG("-Og" COMPILER_SUPPORTS_OG)
+    CHECK_CXX_COMPILER_FLAG("-g" COMPILER_SUPPORTS_G)
```
mliszcz, line 201: Can we remove these checks and assume the flags are always available?

GCC: supports '-Og -g' since 4.8 (Debian 8 - oldest OS we support in CI has GCC 4.9)
Clang: supports '-Og -g' since 4.0.0

configure/CMakeLists.txt
```diff
@@ -219,6 +220,18 @@ if(NOT WIN32)
...
+if(TANGO_ABI_API_CHECKING)
+  if(NOT COMPILER_SUPPORTS_OG OR NOT COMPILER_SUPPORTS_G)
+    message(FATAL_ERROR "Your compiler does not support -g or -Og but this is required for ABI/API compliance checking.")
+  endif()
+
+  if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+    message(FATAL_ERROR "Requires Debug release type for ABI/API compliance checking.")
+  endif()
+
+  add_compile_options(-Og -g)
```
mliszcz, line 232: How about removing TANGO_ABI_API_CHECKING options and enabling '-Og -g' for all debug builds unconditionally? Currently we enable only  '-Og' for all debug builds.

.travis/install_tango.sh
```diff
@@ -2,5 +2,10 @@
 
 set -e
 
+if [[ "$CHECK_ABI_API_COMPLIANCE" == "ON" ]]
```
mliszcz, line 5: Consider moving this check to `.travis.yml` so that it is done once instead of in each file run in CI. Something like (not sure about exit code):
```yaml
script:
  - set -e
  - >
      [[ "$CHECK_ABI_API_COMPLIANCE" == "ON" ]]
      && docker exec -w /home/tango/src cpp_tango .travis/check-ABI-API-compliance.sh
      || true
 - proceed as usual
```

.travis/check-ABI-API-compliance.sh
```diff
@@ -0,0 +1,49 @@
...
+  make
+  abi-dumper libtango.so -o ${base}/libtango-${prefix}-${revision}.dump -lver ${revision}
+  cd "${base}"
+}
+
+new_revision=$(git rev-parse HEAD)
+
+# old_revision is the first parent commit not in the history of this branch
+# see the help of git merge-base for the gritty detailed explanation
+old_revision=$(git merge-base --fork-point tango-9-lts HEAD)
```
mliszcz, line 33: Aren't we interested more in whether the current branch is binary compatible with the PR target branch? I mean maybe we should trigger this job only for PRs and only for those those targeting branch 9.3-backports.

Travis sets some environmental variables that allow to determine whether this is a PR build and what is the name of the target branch.

mliszcz (commented):
Hi @t-b. I understand this is still work in progress, but I added some comments for you to consider that may simplify this PR a bit.

configure/CMakeLists.txt
mliszcz: BTW I just noticed it is compiling with '-g -Og' already:
```
[ 16%] Building CXX object cppapi/server/CMakeFiles/server_objects.dir/eventsupplier.cpp.o
cd /home/michal/Documents/cppTango/build/cppapi/server && /usr/bin/c++  -D_REENTRANT -I/home/michal/Documents/cppTango/cppapi/client -I/home/michal/Documents/cppTango/cppapi/client/helpers -
I/home/michal/Documents/cppTango/cppapi/server -I/home/michal/Documents/cppTango/cppapi/server/idl -I/home/michal/Documents/cppTango/log4tango/include -I/home/michal/Documents/cppTango/build
/cppapi/server -I/home/michal/Documents/cppTango/build/log4tango/include  -std=c++14 -g   -Og -Wall -Wextra -pedantic -fPIC -Winvalid-pch -include /home/michal/Documents/cppTango/build/cppap
i/server/tango.h -o CMakeFiles/server_objects.dir/eventsupplier.cpp.o -c /home/michal/Documents/cppTango/cppapi/server/eventsupplier.cpp
```

My configuration: ubuntu bionic (docker):
```
cmake .. -DCMAKE_INSTALL_PREFIX=/usr/local -DUSE_PCH=ON -DCMAKE_BUILD_TYPE=Debug -DIDL_BASE=$HOME/Documents/tangoid
```

.travis/install_tango.sh
t-b: I'll be implenting that via github actions, so this will go away.

configure/CMakeLists.txt
t-b: I don't think assuming that it is present is good practice. What happens if someone tries to compile it with other compilers (Intel/HP UX, etc)?

configure/CMakeLists.txt
t-b: Always using `-g -Og` for debug is a good idea.

configure/CMakeLists.txt
t-b: I'm correcting myself, this is in the GNU/CLANG branch. Will do.

.travis/check-ABI-API-compliance.sh
t-b: @mliszcz 

> Aren't we interested more in whether the current branch is binary compatible with the PR target branch? 

Yes we are exactly interested in that. As I switched now to github actions, I can't use the travis environment variable. I also would like the script to be runnable locally as well.

I think it is nice to have it for tango-9-lts as well just to teach us how compliance really works. The check is non-fatal only when targeting tango-9-lts, it will be fatal when I do the backport for 9.3-backports.

t-b: Found it. Github action also allows to query the target branch. See `base_ref` in contexts, https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context.

t-b:
@bourtemb @mliszcz Ready.

bourtemb (approved):
Thanks @t-b 
Looks good to me.

Do you know whether these errors reported in the [log ](https://github.com/tango-controls/cppTango/runs/762479009?check_suite_focus=true#step:7:283) are important?:
```
Reading debug-info
WARNING: a "Struct" type with no attributes detected in the DWARF dump (19012805)
Reading v-tables
ERROR: missed type name (241)
ERROR: missed type name (1881813)
ERROR: missed type name (751092)
ERROR: missed type name (751101)
ERROR: missed type name (751162)
ERROR: missed type name (751171)
ERROR: missed type name (1879520)
```

t-b:
@bourtemb These are from the ABI dumper tool. This could be https://github.com/lvc/abi-dumper/issues/21. I'll have a look.

t-b:
@bourtemb I've tried the hint from the issue passing `-lambda` but the warnings don't dissapear.

.travis/check-ABI-API-compliance.sh
mliszcz: I was thinking about different thing. Consider below example:
```
      merge-base   tango-9-lts   refs/pull/ID/merge
          |            |                |
          V            V                V
    a1----a2-----------a3---------------a4
           \                            /
            b1-------------------------b2
                                       A
                                       |
                                 PR branch HEAD
```
Now we're checking that b2 (PR HEAD) is binary compatible with a2 (merge base). I think that for user (who is using latest version of target branch, that is a3, which may not even be binary compatible with a2) it is more interesting if a4 will still be compatible after we merge the PR.

We should be able to extract the merged ref from action trigger, see [GITHUB_REF here](https://help.github.com/en/actions/reference/events-that-trigger-workflows#pull-request-event-pull_request), or maybe something like `refs/pull/${{ github.event.number }}/merge` would work. I've also seen `github.event.after` used as a ref [here](https://github.com/actions/checkout/pull/102), although I do not know how it works.

On the other hand, what you currently did is also useful as it provides information whether the current branch introduces ABI breaking changes. I have no strong opinion on which of these two options is better.

mliszcz (approved):
In general looks good. I just replied to your comment regarding refs which we are comparing to detect ABI breakage. I do know which is the preferred way so I wouldn't mind if you just drop that comment. Thanks!

.travis/check-ABI-API-compliance.sh
bourtemb: Ideally, we want to check the binary compatibility with a previous official release.
So we want to check if a4 (result of the merge) is binary compatible with the latest release tag we want to check (9.3.4-rc6 or even 9.3.3 right now, but might change in the future).

bourtemb:
> @bourtemb I've tried the hint from the issue passing `-lambda` but the warnings don't dissapear. I would leave it like it is.

@t-b did you try the other hint. Use the latest version from the master branch of abi-dumper tool?
I don't like much this approach because if they introduce a bug in their master branch, we get it.
Their latest release was done in 2017 and it looks like there a a few changes since then (11 commits compared to their latest release), including some warning fixes.

.travis/check-ABI-API-compliance.sh
t-b: @mliszcz Thanks, now I understand. I wanted the script also to locally work, so I avoided the additional headache of working with the merged result.

@bourtemb Comparing with a tagged release is also something we would like to know yes.

t-b:
@bourtemb No I did not try latest master of abi-dumper. Feel free to create an issue and assign me if I should give it a try.

I'll merge that now.

PR #664, opened 2020-01-23
[Backport] Avoid undefined interface warning msvc

t-b:
Undefine interface for Microsoft Visual Studio version 14.0.25431.01 Update 3
comming with _MSC_VER = 1900 to fix compilation error:
 "error C2236: unexpected token 'struct'. Did you forget a ';'?


t-b:
@mliszcz `interface` is only a define for MSVC 1900 or later. The mid-term solution for tango-9-lts would probably be to rename the struct member and drop the `#ifdef`.

PR #641, opened 2019-12-16
Fix race conditions between polling threads and user threads pushing events

mliszcz:
Changes proposed in comment in another PR: https://github.com/tango-controls/cppTango/pull/635#issuecomment-565041260

PR #665, opened 2020-01-23
[Backport] Fix race conditions between polling threads and user threads pushing events

mliszcz:
Backport of #641


mliszcz:
Hi @t-b and @bourtemb. I've just merged the corresponding PR to the mainline branch. This one can be reviewed now. One Appveyor job has already passed. It will take few hours to complete all of them, but in my opinion it is safe to merge it as soon as it is reviewed so that we do not block 9.3.4-rc4 release.

PR #668, opened 2020-01-23
CMakeLists.txt: Add option to disable building the test suite

t-b:
@lorenzopivetta I needed to disable the tests as well, so I just extracted it into this PR.

PR #670, opened 2020-01-24
Add switch to disable building tests

lorenzopivetta:
CMakeLists.txt: add cmake switch to disable building the test suite (target 9-backports)


lorenzopivetta:
> Appveyor.yml needs to be adapted so that it passes `-DTANGO_COMPILE_TEST_SUITE=OFF`

@t-b I'm not familiar with appveyor config. Do you mean something like:
[appveyyor.txt](https://github.com/tango-controls/cppTango/files/4108016/appveyyor.txt)

t-b:
@lorenzopivetta Yes!

configure/CMakeLists.txt
```diff
@@ -212,3 +212,5 @@ if(NOT WIN32)
 endif()
 include(GNUInstallDirs)
 include(configure/coveralls.cmake)
+
+option(TANGO_COMPILE_TEST_SUITE "Compile the test suite" ON)
```
mliszcz: LGTM but just my 2c: the standard variable for enabling tests in CMake projects is `BUILD_TESTING` an IMO we should follow that. It is defined in [CTest module](https://cmake.org/cmake/help/latest/module/CTest.html) which looks more or less like:

```cmake
option(BUILD_TESTING "Build the testing tree." ON)

if(BUILD_TESTING)
  # ...
  # Make sure testing is enabled
  enable_testing()
  #...
endif()
```

According to [`add_test()` documentation](https://cmake.org/cmake/help/latest/command/add_test.html), it's no-op when `enable_testing()` was not called.

I think it would be enough to remove explicit call to `enable_testing()` from main Cmake config file and pass `-DBUILD_TESTING=OFF` to achieve  what you wanted (cxxtests would still be generated though).

BTW, to skip building tests without any code changes, you could just:
```bash
make tango tango-static
make install/fast
```

mliszcz: Sure! It should be `if(BUILD_TESTING)` but I see you changed that in #668. Thanks!

t-b:
I'm closing this in favour of https://github.com/tango-controls/cppTango/pull/668, @lorenzopivetta thanks for the initiative!

PR #673, opened 2020-01-27
Switch to python3 for tests

t-b:
I did not upgrade cxxtest completely as I think this is way too much effort. IMHO we should move away from cxxtest.


t-b:
@bourtemb Done, thanks for the review.

@mliszcz I've also fixed the obvious warnings.

PR #676, opened 2020-02-05
Add switch to disable mmx build lts

lorenzopivetta:
Add switch to disable building jpeg MMX support. Targets 9-lts.


lorenzopivetta:
@bourtemb  fixed, assuming that we want 4 spaces indentation...

PR #687, opened 2020-02-18
Fix omniidl error checks

ltjax:
Both error checks did not work before. Now they do. Note that CMake automatically evaluates the variable for `if(variable)` or `if(NOT variable)`, so when evaluating it before you are basically double-dereferencing when you shouldn't.


t-b:
@bourtemb I've removed the "Candidate for Backports" label as this error is not in the 9.3-backports branch.

bourtemb:
@t-b, the second error check seems to be present in 9.3-backports:
https://github.com/tango-controls/cppTango/blob/9.3-backports/cppapi/server/idl/CMakeLists.txt#L27

PR #692, opened 2020-03-13
Always allow to change attribute alarm thresholds

### LINKED ISSUE #401: Cannot change alarm range for memorized attribute
Trying to change the alarm range for a memorized attribute gives the following exception:

alarm_min_max

Note that the current set_value is 1.2 and the user is trying to change the max_alarm to 1.5.

- so, the feature seems to be buggy since it should be possible to change the alarm in this situation
- I think it should always be possible to change the alarm/warning/value ranges of an attribute. So, IMHO, the feature should simply be dropped.

mliszcz:
Fixes #401

I couldn't find one so I also added a test that checks if one can set alarm threshold below current attribute value (the attribute goes immediately into alarm state).

### DIFF
--- a/cpp_test_suite/new_tests/cxx_mem_attr.cpp
+++ b/cpp_test_suite/new_tests/cxx_mem_attr.cpp
@@ -166,6 +166,54 @@ public:
 		TS_ASSERT(s_val_2 == 10);
 	}
 
+/*
+ * Test for changing min and max alarm threshold of a  memorized attribute
+ * with neither min_ nor max_value specified. Such scenario used to fail
+ * (#401), with misleading error message claiming that memorized value
+ * is above new limit.
+ */
+
+    void test_change_max_alarm_threshold_of_memorized_attribute()
+    {
+        const char* attr_name = "Short_attr_w";
+        const DevShort attr_value = 20;
+
+        unset_attribute_min_max_value(attr_name);
+
+        DeviceAttribute value(attr_name, attr_value);
+        TS_ASSERT_THROWS_NOTHING(device1->write_attribute(value));
+
+        auto config = device1->get_attribute_config(attr_name);
+        config.alarms.max_alarm = std::to_string(attr_value + 1);
+        AttributeInfoListEx config_in = { config };
+        TS_ASSERT_THROWS_NOTHING(device1->set_attribute_config(config_in));
+    }
+
+    void test_change_min_alarm_threshold_of_memorized_attribute()
+    {
+        const char* attr_name = "Short_attr_w";
+        const DevShort attr_value = -20;
+
+        unset_attribute_min_max_value(attr_name);
+
+        DeviceAttribute value(attr_name, attr_value);
+        TS_ASSERT_THROWS_NOTHING(device1->write_attribute(value));
+
+        auto config = device1->get_attribute_config(attr_name);
+        config.alarms.min_alarm = std::to_string(attr_value - 1);
+        AttributeInfoListEx config_in = { config };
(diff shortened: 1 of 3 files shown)


mliszcz:
```
28/93 Test #28: CXX::cxx_mem_attr .................***Failed    4.77 sec
Starting DevTest/test
DevTest/test PID: 7825
Starting FwdTest/test
FwdTest/test PID: 7870
Running 3 tests
Testing Mem Attr :

	One memorized attribute failing during init cmd --> OK

Exception : 

Tango exception
Severity = ERROR 
Error reason = API_AttrOptProp
Desc : Device test/debian8/10-> Attribute : Short_attr_w
This attribute is memorized and the memorized value (20) is above the new limit!!
Origin : Attribute::throw_min_max_value()

In MemAttrTestSuite::test_change_max_alarm_threshold_of_memorized_attribute:
/home/tango/src/cpp_test_suite/cxxtest/include/../../new_tests/cxx_mem_attr.cpp:189: Error: Expected (device1->set_attribute_config(config_in)) not to throw, but it did
```

cpp_test_suite/new_tests/cxx_mem_attr.cpp
```diff
@@ -166,6 +166,54 @@ class MemAttrTestSuite: public CxxTest::TestSuite
 		TS_ASSERT(s_val_2 == 10);
 	}
 
+/*
+ * Test for changing min and max alarm threshold of a  memorized attribute
+ * with neither min_ nor max_value specified. Such scenario is reported to
```
bourtemb:
```suggestion
 * with neither min_ nor max_value specified. Such scenario has been reported to
```

bourtemb:
```suggestion
 * with neither min_ nor max_value specified. Such scenario used to
```

bourtemb:
Just wanted to report here something I noticed during my tests. 
I noticed that the alarm thresholds do not seem to do anything on Tango Write Only attributes (no change of quality if someone sets the attribute to a value above the max alarm threshold).
This seems to be expected according to the documentation (https://tango-controls.readthedocs.io/en/latest/development/device-api/attribute-alarms.html#the-level-alarms) since the alarm thresholds seem to make sense only for tango attribute numerical read types.

So, setting the alarm threshold on `Short_attr_w ` attribute does not really seem to make sense.

I saw the mem_attr test updated in this PR is using a Write Only attribute.
This is not representative of the usage from our users (Write Only attributes are not recommended).
I understand why you've chosen this attribute. It is because it was the one used by one of the tests on memorized attributes.
There does not seem to be many different memorized attributes provided by DevTest class.
They all seem to be Write Only attributes (`Short_attr_w`, `String_attr_w`, `Boolean_attr_w`), except Enum_attr_rw which is a DevEnum READ_WRITE attribute.

I tested the fix on a DevLong READ_WRITE attribute on one of my classes and it works as expected.

cpp_test_suite/new_tests/cxx_mem_attr.cpp
mliszcz: Thanks! Fixed as you proposed

mliszcz:
> I understand why you've chosen this attribute. It is because it was the one used by one of the tests on memorized attributes.

@bourtemb exactly, this is why I picked this attribute. I did not want to add new attribute for this simple test-case and I did not want to memorize any existing attribute as it may impact other tests. Even if property change is discarded at later point, the code that used to cause the issue was executed also for write-only attributes.

PR #696, opened 2020-03-17
Ensure location transparency for image and spectrum attributes

### LINKED ISSUE #416: Remote vs local transparency difference when reading spectrum or image attributes
When a Tango client tries to read a spectrum or image attribute which is not polled from a device exported by the same device server process as the client, by default (when Attribute::set_value() release parameter is set to false), omniORB and Tango will not do any copy of the data buffer and will give the spectrum/image buffer data pointer to the client without transferring ownership of this buffer to the client.
In this case, if the client modifies directly the data pointed by the pointer it got, it will also modify the data on the server side's buffer (dangerous!!).
In the case of reading a spectrum or image polled attribute, Tango is already making a copy of the data and providing the copy/transferring ownership to the client. So in this case, the client can do whatever it wants with the data pointer it got. It is the same in the case of a remote client.

Here are some parts from the answer we got from omniORB's maintainer on this topic:

> The C++ sequence mapping intentionally breaks local/remote transparency with regard to the release flag, for efficiency purposes.
> [...]
> Really, the summary is that the ability to construct a sequence with release set to false is a short-cut that can help performance, at the expense of a loss of local/remote transparency.
> 
> If it is essential that client code does not need to care about this, then yes, the implementation should avoid using sequences with release set to false. That will come with a bit of a performance cost, which may or may not be a problem for your application.
> 
> The alternative is for clients to become aware of this situation. If a client wants to modify the data in a sequence it has received, it can look at its release flag. If release is false, it knows the sequence is "borrowing" the buffer, so it should create a new sequence object with the copy constructor, because that copies the buffer as well.

This issue was created in order to get always the same behaviour on the client side, meaning, the client can do whatever it wants with the data it got after reading a spectrum or image attribute. No matter whether the read was done on a local (same process) or remote device.
In order to avoid performance penalties, the best would be that Tango would detect whether the client is local or remote and if the client is local, Tango would always force a copy. In the remote call case, we should stick with the current behaviour and avoid an unnecessary copy.

The code creating the result sequences is located in DATA_IN_OBJECT(A,B,C,D) macro in server/device.h file.

mliszcz:
Fixes #416

### DIFF
--- a/cppapi/server/dintrthread.h
+++ b/cppapi/server/dintrthread.h
@@ -58,7 +58,7 @@ struct _ShDevIntrTh
 	bool			cmd_pending;	// The new command flag
 	DevIntrCmdCode	cmd_code;		// The command code
 	bool			th_running;		// Thread running flag
-	#if _MSC_VER > 1900
+	#if _MSC_VER >= 1900
 	#ifdef interface
 	#undef interface
 	#endif


mliszcz:
```
In AttrMiscTestSuite::test_location_transparency:
/home/tango/src/cpp_test_suite/cxxtest/include/../../new_tests/cxx_attr_misc.cpp:1092: Error: Test failed: Failure, check server logs.
/home/tango/src/cpp_test_suite/cxxtest/include/../../new_tests/cxx_attr_misc.cpp:1092: Error: Assertion failed: result
	Destroying suite at Wed Mar 18 08:15:31 2020
```

Before (same address in server read callback and at the client - command - side):
```
[read_SpectrumAttrWithSharedBuffer] att.get_buffer()=0x7f8834016da8
[cmd_check_location_transparency] seq->release()=0
[cmd_check_location_transparency] seq->get_buffer()=0x7f8834016da8
[cmd_check_location_transparency] *seq[0]=0x7f8834016da8
[cmd_check_location_transparency] *data[0]=0x7f8834016da8
```

After:
```
[read_SpectrumAttrWithSharedBuffer] att.get_buffer()=0x7fcdc8016da8
[cmd_check_location_transparency] seq->release()=1
[cmd_check_location_transparency] seq->get_buffer()=0x7fcdc4008610
[cmd_check_location_transparency] *seq[0]=0x7fcdc4008610
[cmd_check_location_transparency] *data[0]=0x7fcdc8016da8
```

PR #697, opened 2020-03-18
Ensure location transparency for image and spectrum attributes

### LINKED ISSUE #416: Remote vs local transparency difference when reading spectrum or image attributes

mliszcz:
Fixes #416.


mliszcz:
```
In AttrMiscTestSuite::test_location_transparency:
/home/tango/src/cpp_test_suite/cxxtest/include/../../new_tests/cxx_attr_misc.cpp:1092: Error: Test failed: Failure, check server logs.
/home/tango/src/cpp_test_suite/cxxtest/include/../../new_tests/cxx_attr_misc.cpp:1092: Error: Assertion failed: result
	Destroying suite at Wed Mar 18 08:15:31 2020
```

Before (same address in server read callback and at the client - command - side):
```
[read_SpectrumAttrWithSharedBuffer] att.get_buffer()=0x7f8834016da8
[cmd_check_location_transparency] seq->release()=0
[cmd_check_location_transparency] seq->get_buffer()=0x7f8834016da8
[cmd_check_location_transparency] *seq[0]=0x7f8834016da8
[cmd_check_location_transparency] *data[0]=0x7f8834016da8
```

After:
```
[read_SpectrumAttrWithSharedBuffer] att.get_buffer()=0x7fcdc8016da8
[cmd_check_location_transparency] seq->release()=1
[cmd_check_location_transparency] seq->get_buffer()=0x7fcdc4008610
[cmd_check_location_transparency] *seq[0]=0x7fcdc4008610
[cmd_check_location_transparency] *data[0]=0x7fcdc8016da8
```

mliszcz:
I did some refactoring and removed ~400 lines of duplicated code so that functional change was applied just in one place instead of ~30.

AttributeValue_X to DeviceAttribute conversion functions extracted in this PR can be re-used also in EventConsumer::attr_to_device calls (to save extra ~600 lines), but I need more time to work on this (I'm seeing some instability in tests). If I'll manage to fix it, a dedicated PR will come.

t-b:
@mliszcz Please ping me once this is ready to review.

mliszcz:
Hi @t-b. I think this can be reviewed now. Last Appveyor build failed because I used `not` instead of `!`. I just fixed this and half of the appveyor jobs already passed. You may want to wait for the remaining jobs (32-bit builds), but I do not expect any surprises here.

As for the PR, this is the only functional change: https://github.com/tango-controls/cppTango/pull/697/commits/c52e23451c222fae686681ad7bdc81900af1d9fe

The rest is just refactoring to remove code duplication.

If backport is needed, we'd need to apply the change to all cases in switch here: https://github.com/tango-controls/cppTango/blob/9.3-backports/cppapi/client/api_util.tpp#L83-L92

Thanks in advance for your review!

t-b (changes requested):
8a989798 (Refactor AttributeValue to DeviceAttr conversion, 2020-03-17)

Is there a benefit of defining the templates instead of just declaring them?

```
$git diff .
diff --git a/cppapi/server/data_type_utils.h b/cppapi/server/data_type_utils.h
index f5444068..7420015e 100644
--- a/cppapi/server/data_type_utils.h
+++ b/cppapi/server/data_type_utils.h
@@ -18,7 +18,7 @@ struct TangoTypeTraitsBase
 };

 template <AttributeDataType Type>
-struct TangoTypeTraits {};
+struct TangoTypeTraits;

 #define ENTRY(TYPE_ENUM, SEQ_VAR_TYPE) \
     template <> struct TangoTypeTraits<TYPE_ENUM> : TangoTypeTraitsBase<SEQ_VAR_TYPE> {};
@@ -53,7 +53,7 @@ struct TangoTypeOpsBase
 };

 template <AttributeDataType Type>
-struct TangoTypeOps {};
+struct TangoTypeOps;

 #define ENTRY(TYPE_ENUM, DA_SEQ, UNION) \
     template <> struct TangoTypeOps<TYPE_ENUM> : TangoTypeOpsBase< \
```

I tend to only declare these kind of templates, as the non-specialized versions
should never be used. But then again maybe MSVC barfs at that.

49cd5211 (Move code from api_util.tpp to api_util.cpp, 2020-03-17)

Intermediate commit.

7c6eeba6 (Replace not operator with bang (!) for MSVC, 2020-03-17)

Please fold that into 8a989798.

471bfe52 (Refactor AttributeValue_3 to DeviceAttr conversion, 2020-03-17)

Good!

2b73e529 (Rename and move attr_to_device_base, 2020-03-17)

Intermediate commit.

1ef106ff (Add test for checking location transparency, 2020-03-18)

Good, TSM_ASSERT is also the correct choice.

c52e2345 (Enforce location transparency in attribute access, 2020-03-18)

Good! We are sacrificing some performance here, but I think this is a good
change nevertheless.

72caa0dd (Move DeviceAttr. conversions to separate header, 2020-03-18)

Intermediate commit.

2426de6d (Add conversions from ZmqAttributeValue to DevAttr, 2020-03-18)

Nice!

b72f759b (Remove duplications in AttrVal conversion handling, 2020-03-18)

Yeep.

96aaa4ea (Correct test_change_max_alarm_threshold_below_current_value, 2020-05-12)

Good.

ba22c3c1 (Correct access to AttrValUnion when storing state, 2020-05-28)

This fix should be folded into the commit introducing the code.

b022f18c (Support const& in attr_val_union(AttributeValue&), 2020-05-28)

intermediate commit.

d3e97c20 (Rename some functions in data_type_utils.h, 2020-05-28)

intermediate commit.

b5c23b8a (Add detail ns in device_attribute_conversions.h, 2020-05-28)

intermediate commit.

a576cfca (Move CopySeq tempate parameter to function argument, 2020-05-28)

intermediate commit.

0eda909e (Replace not with bang for MSVC compatibility, 2020-05-29)

intermediate commit.

In general this is a very nice improvement!
I also like the `_copy`/`_nocopy` variants. I would even move the
`attribute_value_to_device_attribute_5` and
`attribute_value_to_device_attribute_3` into the detail namespace.

For reviewing it would be easier to avoid intermediate commits. As intermediate
commits means I have to learn several locations and APIs during review instead of the final
one. But this is of course open for discussion how we want to handle that.

Issue #241, opened 2016-06-08
Read value for attribute xxx has not been updated


tango-tickets-migrator:
If you use the read_attributes (Note the s) call to read at the same moment two (or more) attributes with
att2 pushing event for att1, you received the error "read value for attribute att1 has not been updated" when
reading att1

Fixed in SVN

Reported by: taurel

Original Ticket: [tango-cs/bugs/799](https://sourceforge.net/p/tango-cs/bugs/799)

hardion:
To be closed (sorry for the spam)

bourtemb:
Fixed in Tango 9.2.5

PR #698, opened 2020-03-19
Safe interworking between read_attribute() and push_event()

### LINKED ISSUE #443: push_archive_event() or push_change_event() methods causing a crash in some specific situations
We have observed some device servers crashes when _DeviceImpl::push_archive_event()_ or _Device::Impl::push_change_event()_ methods are invoked in some specific situations.

For instance when these methods are invoked after having invoked attr.set_value() inside a _read_MyAttributeName(Tango::Attribute &attr)_ method.
Here is a code example for an attribute named _Value_:

```
void TestArchive::read_Value(Tango::Attribute &attr)
{
    DEBUG_STREAM << "TestArchive::read_Value(Tango::Attribute &attr) entering... " << endl;
    /*----- PROTECTED REGION ID(TestArchive::read_Value) ENABLED START -----*/
    Tango::DevLong64 x = 0x5555333377771111;
    *attr_Value_read = x;
    // No problem when push_<archive|change>_event is invoked before attr.set_value()
    //push_archive_event(attr.get_name(), &x);
    //	Set the attribute value
    attr.set_value(attr_Value_read);
    push_archive_event(attr.get_name(),&x); // A crash will occur after exiting TestArchive::read_Value
    /*----- PROTECTED REGION END -----*/	//	TestArchive::read_Value
}

```

A similar problem could arise when reading an attribute _X_ is triggering push_archive_event() or push_change_event() on another attribute _Y_ and if these 2 attributes are read together in the same read_attributes(). In this particular case, a crash will occur only if _Y_ is read before _X_.
This case may happen in Tango 9 when the 2 attributes are polled at the same polling period, since Tango 9 is using read_attributes() by default to read all the attributes configured with the same polling period.

mliszcz:
This PR fixes two crashes reported in #443:
1. When Attr1 and Attr2 are read in single read_attributes({Attr1, Attr2}) and Attr2 pushes change event for Attr1 (order of attribute names passed to read_attributes is important here)
2. When attribute pushes a change event for itself from read callback after calling set_value.

In the first case, when Attr2 read callback pushes an event for Attr1, it overwrites and then releases CORBA sequence associated with Attr1. This is because sequence obtained from Attr1 read callback is not yet copied for network transfer when Attr2 read callback is called, i.e. it worked like this pseudocode:
```cpp
for(Attribute& a : attributes) { invoke_read_callback(a); }
for(Attribute& a : attributes) { store_sequence(a); }
```
This PR changes the behavior to be like:
```cpp
for(Attribute& a : attributes) { invoke_read_callback(a); store_sequence(a); }
```
This resolved the issue and allowed to revert correction for #241 which attempted to fix the same problem by keeping value flag set to true after deleting the sequence. Correction for #241 (330a68c) worked only for scenarios where attribute that pushed event was read first and resulted in a crash when attribute reading order was inverted.

In the second case, when push_event is called after set_value, we now get API_AttrValueNotSet exception instead of a crash. This is just result of reverting previous #241 correction. We cannot really 'fix' this scenario and make it work as one would expect because push_event and set_value are operating on the same buffer and push_event automatically releases that buffer when it is done with pushing the event. See also https://github.com/tango-controls/cppTango/issues/443#issuecomment-569268014.

I also took the opportunity to refactor whole Device_3Impl::read_attributes_no_except and split it into more manageable chunks.

Fixes #443, #241.

### DIFF
--- a/cpp_test_suite/cpp_test_ds/DevTest.h
+++ b/cpp_test_suite/cpp_test_ds/DevTest.h
@@ -157,6 +157,8 @@ public :
 	void read_DynEnum_attr(Tango::Attribute &att);
 	void read_ReynaldPoll_attr(Tango::Attribute &att);
 	void read_attr_asyn_write(Tango::Attribute &att);
+	void read_ReadWithPushAttr1(Tango::Attribute&);
+	void read_ReadWithPushAttr2(Tango::Attribute&);
 
 	void write_Short_attr_rw(Tango::WAttribute &att);
 	void write_Long64_attr_rw(Tango::WAttribute &att);
@@ -365,6 +367,8 @@ protected :
     double                                  Reynald_val;
 
     Tango::DevLong                          attr_asyn_write_val;
+    Tango::DevShort attr_with_push_1;
+    Tango::DevShort attr_with_push_2;
 };
 
 #endif
--- a/cpp_test_suite/cpp_test_ds/DevTest.h
+++ b/cpp_test_suite/cpp_test_ds/DevTest.h
@@ -159,6 +159,7 @@ public :
 	void read_attr_asyn_write(Tango::Attribute &att);
 	void read_ReadWithPushAttr1(Tango::Attribute&);
 	void read_ReadWithPushAttr2(Tango::Attribute&);
+	void read_PushItselfAfterSetAttr(Tango::Attribute&);
 
 	void write_Short_attr_rw(Tango::WAttribute &att);
 	void write_Long64_attr_rw(Tango::WAttribute &att);
@@ -369,6 +370,7 @@ protected :
     Tango::DevLong                          attr_asyn_write_val;
     Tango::DevShort attr_with_push_1;
     Tango::DevShort attr_with_push_2;
+    Tango::DevShort attr_push_itself_after_set;
 };
 
 #endif
--- a/cpp_test_suite/new_tests/cxx_attr_misc.cpp
+++ b/cpp_test_suite/new_tests/cxx_attr_misc.cpp
@@ -1114,6 +1114,9 @@ cout << "status = " << status << endl;
         std::vector<std::string> attributes = {"ReadWithPushAttr2", "ReadWithPushAttr1"};
         std::vector<DeviceAttribute>* result = nullptr;
         TS_ASSERT_THROWS_NOTHING(result = device1->read_attributes(attributes));
+        Tango::DevShort out;
+        TS_ASSERT_THROWS_NOTHING((*result)[0] >> out);
+        TS_ASSERT_THROWS_NOTHING((*result)[1] >> out);
         delete result;
     }
 
@@ -1122,8 +1125,30 @@ cout << "status = " << status << endl;
         std::vector<std::string> attributes = {"ReadWithPushAttr1", "ReadWithPushAttr2"};
         std::vector<DeviceAttribute>* result = nullptr;
         TS_ASSERT_THROWS_NOTHING(result = device1->read_attributes(attributes));
+        Tango::DevShort out;
+        TS_ASSERT_THROWS_NOTHING((*result)[0] >> out);
+        TS_ASSERT_THROWS_NOTHING((*result)[1] >> out);
         delete result;
     }
+
+/*
+ * Test that if attribute pushes an event for itself from its read callback
+ * and push_event is called after set_value, an exception is raised. Such
+ * scenario was reported to cause a crash (#443).
+ */
+
+    void test_read_attribute_that_pushes_event_for_itself_after_calling_set_value()
+    {
+        DeviceAttribute result;
+        TS_ASSERT_THROWS_NOTHING(result = device1->read_attribute("PushItselfAfterSetAttr"));
+
+        Tango::DevShort out;
+        TS_ASSERT_THROWS_ASSERT(
+            result >> out,
+            Tango::DevFailed& e,
+            TS_ASSERT(string(e.errors[0].reason.in()) == "API_AttrValueNotSet"));
+    }
+
 };
 #undef cout
 #endif // AttrMiscTestSuite_h
--- a/cpp_test_suite/new_tests/cxx_attr_misc.cpp
+++ b/cpp_test_suite/new_tests/cxx_attr_misc.cpp
@@ -1109,6 +1109,9 @@ cout << "status = " << status << endl;
  * used to fail with API_AttrValueNotSet as reported in #201.
  */
 
+    static constexpr Tango::DevShort READ_WITH_PUSH_ATTR_1_VALUE = 13;
+    static constexpr Tango::DevShort READ_WITH_PUSH_ATTR_2_VALUE = 17;
+
     void test_read_two_attributes_one_pushes_event_for_another_1st_pushes()
     {
         std::vector<std::string> attributes = {"ReadWithPushAttr2", "ReadWithPushAttr1"};
@@ -1116,7 +1119,9 @@ cout << "status = " << status << endl;
         TS_ASSERT_THROWS_NOTHING(result = device1->read_attributes(attributes));
         Tango::DevShort out;
         TS_ASSERT_THROWS_NOTHING((*result)[0] >> out);
+        TS_ASSERT_EQUALS(out, READ_WITH_PUSH_ATTR_2_VALUE);
         TS_ASSERT_THROWS_NOTHING((*result)[1] >> out);
+        TS_ASSERT_EQUALS(out, READ_WITH_PUSH_ATTR_1_VALUE);
         delete result;
     }
 
@@ -1127,7 +1132,9 @@ cout << "status = " << status << endl;
         TS_ASSERT_THROWS_NOTHING(result = device1->read_attributes(attributes));
         Tango::DevShort out;
         TS_ASSERT_THROWS_NOTHING((*result)[0] >> out);
+        TS_ASSERT_EQUALS(out, READ_WITH_PUSH_ATTR_1_VALUE);
         TS_ASSERT_THROWS_NOTHING((*result)[1] >> out);
+        TS_ASSERT_EQUALS(out, READ_WITH_PUSH_ATTR_2_VALUE);
         delete result;
     }
 
--- a/cpp_test_suite/new_tests/cxx_attr_misc.cpp
+++ b/cpp_test_suite/new_tests/cxx_attr_misc.cpp
@@ -1102,6 +1102,28 @@ cout << "status = " << status << endl;
         TS_ASSERT_EQUALS(Tango::ALARM, device1->state());
         TS_ASSERT_EQUALS(Tango::ATTR_ALARM, device1->read_attribute(attr_name).get_quality());
     }
+
+/*
+ * Tests that it is possible to read two attributes in a single network call
+ * if one of the attributes pushes an event for the other one. Such scenario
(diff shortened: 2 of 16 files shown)


t-b (commented):
All in all I think this is good. I just have some minor questions. But how would we integrate that into 9.3-backports?

815f59a9 (Add test for event push during multiple attr read, 2020-03-19)

Good.

22e8c7f8 (Split large read_attributes_no_except function, 2020-03-19)

I think call_read_attr_if_needed should have `nb_wanted_attr == 0` instead of `!= 0`.
Rest looks good. Very nice work!!

13ffd029 (Extract diagnostics handling from read_attributes, 2020-03-19)

Jeep!

46797c36 (Separate state/status handling in read_attributes, 2020-03-19)

Difficult one, but I'd say good.

758fecf3 (Reorder attribute read callbacks in read_attributes, 2020-03-19)

Is it true that the find_if calls find at most one matching entry? I'm asking because I was wondering
what happens if multiple matches are found.

f98bccc3 (Revert broken correction for #241, 2020-03-19)

Please cite in the future the commit introducing the breakage, in this case it was 330a68c6
(- Fix bug 799 (Read value for att xxx has not been updated) - Fix Valgrind
warning in DeviceProxy dtor for instance created using the default ctor,
2016-06-08)

aa0564a7 (Add test for calling push_event after set_value, 2020-03-19)

Good.

3d322985 (Remove Device_3Impl::reading_state_necessary, 2020-03-20)

Nice!

25bf51f8 (Refactor storing errors in read_attributes, 2020-03-20)

Okay

fb268276 (Remove redundant check from read_attributes, 2020-03-20)

We are never called with idx_in_multi_attr, good!

8e82a3f1 (Reduce indentation in Device3Impl::read_attributes, 2020-03-20)

Very nice!

dc47d152 (Extract function from Device_3Impl::read_attributes, 2020-03-20)

I think I understand it.

7ac15379 (Reorder calls in Device_3Impl::read_attributes, 2020-03-21)

Are you sure that the always_executed_hook() can be moved around like that?

066e486b (Remove wanted_attr vector from read_attributes, 2020-03-21)

Looks faitful.

fb27ced8 (Refactor handling of read_attributes, 2020-03-21)

Hmmh.

4e09f66e (Early calculate index in data for netwok transfer, 2020-03-21)

Okay.

3b3a4860 (Simplify access mode handling in read_attributes, 2020-03-21)

Much nicer!

41aade0b (Remove flag argument from Device3::read_attributes, 2020-03-21)

Can you explain why removing second_try does not change anything?

757da14f (Use RAII for storing device in SubDevDiag, 2020-03-21)

Very nice refactoring!

mliszcz:
Thanks @t-b for looking into that PR.

> 
> 
> All in all I think this is good. I just have some minor questions. But how would we integrate that into 9.3-backports?
>
Ideally we should just move call to `data_into_net_object(att,aid,index,w_type,true);` immediately after we invoke the read callback in the first loop over attributes (`attr_vect[att.get_attr_idx()]->read(this,att);`). The problem is that after the first loop there are some other loops which may alter the generated sequence (`att.set_rvalue();` and `dev_attr->add_write_value(att)`) or raise some errors that should prevent from storing the sequence in `aid`.

I do not see a simpler solution (without introducing extra ifs, copying and adding more complexity) than reorganizing how we iterate over attributes in this function. The functional change in this PR does not require breaking the ABI. We may do the same without extracting any helper functions or using C++11 features. It just was simpler to me to first split the function and then reorder things.

 
> [22e8c7f](https://github.com/tango-controls/cppTango/commit/22e8c7f81f27d5f4d6268e4c435f80c4facd8467) (Split large read_attributes_no_except function, 2020-03-19)
> 
> I think call_read_attr_if_needed should have `nb_wanted_attr == 0` instead of `!= 0`.
> Rest looks good. Very nice work!!
> 
Yes, good catch! I fixed it, probably accidentally, in 7ac153795. I'll not change 22e8c7f if it's ok for you.


> [758fecf](https://github.com/tango-controls/cppTango/commit/758fecf30d22da77ecde10f3f755b8e23f15edd4) (Reorder attribute read callbacks in read_attributes, 2020-03-19)
> 
> Is it true that the find_if calls find at most one matching entry? I'm asking because I was wondering
> what happens if multiple matches are found.
>
Yes, there will be either one or no matches. If there were multiple matching elements, it would mean that in `wanted_attr` there are two elements with the same `idx_in_names`. And this cannot hapen because we build this vector inside read_attributes_no_except function:
```cpp
		for (i = 0;i < nb_names;i++)
		{
			AttIdx x;
			x.idx_in_names = i;
			// later:
			wanted_attr.push_back(x);
```

Anyway, this list and the index is removed in subsequent commits.

> [f98bccc](https://github.com/tango-controls/cppTango/commit/f98bccc3e24dd4beebf609ba6fe61817e8ce8036) (Revert broken correction for #241, 2020-03-19)
> 
> Please cite in the future the commit introducing the breakage, in this case it was [330a68c](https://github.com/tango-controls/cppTango/commit/330a68c6af06eabfe10ce59f0e74a9c03f33e8d2)
> (- Fix bug 799 (Read value for att xxx has not been updated) - Fix Valgrind
> warning in DeviceProxy dtor for instance created using the default ctor,
> 2016-06-08)
> 
Thanks, I reworded that commit's message.

> [dc47d15](https://github.com/tango-controls/cppTango/commit/dc47d15263dc4c41d08e214e15c1dbb9f4c33a8a) (Extract function from Device_3Impl::read_attributes, 2020-03-20)
> 
> I think I understand it.
> 
The idea was to separate building the list of AttIdx from the list of wanted_(w)_attr, because I planned to completely remove wanted_attr lists and reduce AttrIdx structure to just one index. Having these concerns separated allows to deal with them separately.

> [7ac1537](https://github.com/tango-controls/cppTango/commit/7ac153795411a82224ed25ce0818d6f14066b666) (Reorder calls in Device_3Impl::read_attributes, 2020-03-21)
> 
> Are you sure that the always_executed_hook() can be moved around like that?
>
Thanks for this comment! I re-ordered things again in 7bdc4d3 so that we call `att.save_alarm_quality()` before invoking the hook. Otherwise we may see some misleading printouts (but no functional impact) if we call `check_alarm` from the hook. What matters for sure is that we cannot change the relative ordering between always_executed_hook, read_attr_hardware and read callback, and this is preserved in the commit.



> [4e09f66](https://github.com/tango-controls/cppTango/commit/4e09f66eadcbd67ae02b261f382487e013cbd40a) (Early calculate index in data for netwok transfer, 2020-03-21)
> 
> Okay.
> 


> [41aade0](https://github.com/tango-controls/cppTango/commit/41aade0b8c2c1a04db5c2b6238312499c5b4fb48) (Remove flag argument from Device3::read_attributes, 2020-03-21)
> 
> Can you explain why removing second_try does not change anything?
> 
It is really easy to get lost in all these indices. For each attribute we maintain thre indices: index in device attribute list, index in list of requested attributes passed to read_attributes and index in data structure where we store the read results. The last one is needed because on failure we retry the procedure (call again read_attributes) just for the failed attributes, and the output data structure stays the same.

Previously the function signature was: `read_attributes_no_except(names_from_device,aid,true,idx_in_back);`. If last but one parameter (second_try) was true, we used `idx_in_back` to determine at which index in `aid` we should store data for each attribute in the `names_from_device` list, like: `aid[idx_in_back[idx_in_names_from_device]]`. If `second_try` was false, we just ignored the mapping parameter and accessed `aid` like `aid[idx_in_names_from_device]`. Now I've changed it to always translate indices using `idx_in_back` vector (to remove the flag argument and extra ifs). In cases when previously we ignored `idx_in_back` (when `second_try` is `false`), we now set `idx_in_back` to identity mapping (`{0, 1, 2, ...}`) so that `aid[idx]` and `aid[idx_in_back[aid]` are equivalent. This identity mapping is generated in special `read_attributes_no_except` overload.

To sum up, previously we called:
* normal case: `read_attributes_no_except(all_names, aid, false, whatever)`
* failure: `read_attributes_no_except(failed_names, aid, true, indices_of_failed)`

Now we call:
* normal case: `read_attributes_no_except(all_names, aid, identity)` or overload: `read_attributes_no_except(all_names, aid)`
* failure: `read_attributes_no_except(failed_names, aid, indices_of_failed)`

t-b (dismissed):

cpp_test_suite/cpp_test_ds/DevTest.h
```diff
@@ -365,6 +368,9 @@ protected :
     double                                  Reynald_val;
 
     Tango::DevLong                          attr_asyn_write_val;
+    Tango::DevShort attr_with_push_1;
+    Tango::DevShort attr_with_push_2;
+    Tango::DevShort attr_push_itself_after_set;
```
bourtemb, line 373: attr_push_itself_after_set is not used it seems.

cpp_test_suite/new_tests/cxx_attr_misc.cpp
```diff
@@ -1102,6 +1102,53 @@ cout << "status = " << status << endl;
...
+ */
+
+    void test_read_two_attributes_one_pushes_event_for_another_1st_pushes()
+    {
+        std::vector<std::string> attributes = {"ReadWithPushAttr2", "ReadWithPushAttr1"};
+        std::vector<DeviceAttribute>* result = nullptr;
+        TS_ASSERT_THROWS_NOTHING(result = device1->read_attributes(attributes));
+        Tango::DevShort out;
+        TS_ASSERT_THROWS_NOTHING((*result)[0] >> out);
+        TS_ASSERT_THROWS_NOTHING((*result)[1] >> out);
```
bourtemb, line 1123: What about testing the returned read values of the attributes too?

cppapi/server/device_3.cpp
```diff
@@ -300,741 +340,645 @@ Tango::AttributeValueList_3* Device_3Impl::read_attributes_3(const Tango::DevVar
 // arguments:
 //		in :
 //			- names: The names of the attribute to read
-//			- aid : Structure with pointers for data to be returned to caller (several pointers acoording to
+//			- data : Structure with pointers for data to be returned to caller (several pointers acoording to
```
bourtemb: We can take this opportunity to fix this typo in the comment:
```suggestion
//			- data : Structure with pointers for data to be returned to caller (several pointers according to
```

bourtemb (commented):
Partial review for the moment (big PR) but some comments already.

cppapi/server/device_3.cpp
mliszcz: Fixed in https://github.com/tango-controls/cppTango/pull/698/commits/ad8e9efddfda3f5e8a63c43ae8c92156c30367d7

cpp_test_suite/new_tests/cxx_attr_misc.cpp
mliszcz: Checks added in https://github.com/tango-controls/cppTango/pull/698/commits/c17dea18042d6b5bf44f4e66b8b66fa282b92d80

cpp_test_suite/cpp_test_ds/DevTest.h
mliszcz: It should be used. There was a problem in attribute read callback, I was passing a pointer to stack value (which won't be copied if e.g. callback is called to evaluate device status). This is fixed now: https://github.com/tango-controls/cppTango/pull/698/commits/8ef86048505785f8fd50f9fa7a5409881f874848

t-b (approved):

Issue #586, opened 2019-08-01
Use automatic code formatting


t-b:
Inspired from https://github.com/tango-controls/cppTango/pull/581#issuecomment-516828044

Situation:
 The code base is currently manually formatted and indented. So it sometimes looks like

![grafik](https://user-images.githubusercontent.com/1185677/62293817-b273a580-b469-11e9-82c4-8e552fe837d4.png)

This is of course not nice as it forces the developer to manually indent using mixed indentation. And it is also a waste of developers time.

Proposal:
- [ ] Add clang-format file from https://github.com/AllenInstitute/ZeroMQ-XOP/blob/master/src/.clang-format
- [ ] Add a cmake target which calls clang-format
- [ ] Start discussion about how formatting should be done
- [ ] Make **1** commit changing all the C++ code to use the new formatting
- [ ] Document how users can ignore this commit when blaming using the new git blame feature available from git 2.23, see also https://github.com/gitster/git/commit/ae3f36dea16e51041c56ba9ed6b38380c8421816.
- [ ] Add a travis job which ensure that the PR matches our code formatting. Passing this check is *mandatory* for all PRs.
- [ ] Finalize formatting discussion

Ingvord:
+1

My 50 cents: auto-formatting procedure must be as simple as `Ctrl+Alt+L` in CLion

bourtemb:
> 
> 
> +1
> 
> My 50 cents: auto-formatting procedure must be as simple as `Ctrl+Alt+L` in CLion

It looks like CLion already supports clang-format:
https://www.jetbrains.com/help/clion/clangformat-as-alternative-formatter.html

PR #700, opened 2020-03-25
Upgrade llvm in warning as errors CI tests

t-b:


cppapi/server/fwdattribute.cpp
```diff
@@ -133,10 +133,10 @@ void FwdAttribute::set_att_config(const Tango::AttributeConfig_5 &conf)
 // Enum labels (in case of)
 //
 
-    enum_labels.clear();
```
bourtemb: This new indentation looks strange on github page showing the differences (this page).
It was not aligned before, but it still does not seems to be aligned with the code above (line 130) after.

cppapi/server/attrgetsetprop.cpp
```diff
@@ -440,8 +440,8 @@ void Attribute::add_config_5_specific(AttributeConfig_5 &conf)
 
 void Attribute::set_properties(const Tango::AttributeConfig &conf,std::string &dev_name,TANGO_UNUSED(bool from_ds),std::vector<AttPropDb> &v_db)
 {
-    if (name_lower == "state" || name_lower == "status")
-        return;
+  if (name_lower == "state" || name_lower == "status")
+    return;
```
bourtemb: 2 spaces indentation instead of 4?

cppapi/server/fwdattribute.cpp
bourtemb: I guess this is again due to the mixing between tabs and spaces.

bourtemb:
> Looks good! Could you please re-start failed jobs? 

I restarted them. I had to log out from Travis and log in again to be able to see the restart job buttons. Strange.

> Also, ubuntu 20.04 build started to fail, but this is unrelated to the changes you made in this PR.

Indeed. We have a an issue with 20.04 which is trying to install tzdata and his asking some questions during the install/configuration process.

t-b:
@bourtemb I'll look into that. How wide is a tab at cppTango? 4 spaces?

bourtemb:
> @bourtemb I'll look into that. How wide is a tab at cppTango? 4 spaces?

I would say 4 but I'm pretty sure it's not consistent all over the place. And in some files tabs are used instead of spaces.
The recommendation up to now was to fit the style of the surrounding code.
In this particular case, it looks like the surrounding code is using tabs, so tabs should be used here I would say?

Hopefully, when https://github.com/tango-controls/cppTango/issues/586 will be addressed, all these indentation issues will disappear.

t-b:
@bourtemb Thx.

t-b:
@mliszcz @bourtemb Updated.

t-b:
The important travis task passed, see https://travis-ci.org/github/tango-controls/cppTango/jobs/678571433.

PR #573, opened 2019-07-21
Fix #496 blind event clients after dev restart (attempt 2)

mliszcz:
This PR restores Attribute::client_lib contents from temporary storage during DevReset handling.

Per original issue, attribute shall be identified by its name instead of index in multiattribute. This change is included as well but I was unable to reproduce this issue in a test (dynamic attribute created with IOAddAttribute was somehow persisted during the reset).

Fixes #496.

Issue #361, opened 2017-04-12
missing attr_conf event after "Reset to default value"


gscalamera:
Using Jive, after right click on "Attribute config"->Description selecting "Reset to default value", the attr_conf event with the default value for the description is not sent

bourtemb:
Thanks Graziano for the report.
What jive does when the user clicks on "Reset to default value" is the following:

1. Delete the description for this attribute in the Database
2. Restart the device

From the tests I've done with a C++ event subscriber talking to a C++ server using Tango 9.2.5a, it looks like the event subscriber no longer receives any attribute configuration event after a restart device.

mliszcz:
Hi @bourtemb and @gscalamera 

I don't think this is a bug. Reynald already explained the behavior on Jive side.

The usual way of triggering Attr_Conf event from external client (like Jive) is to call CORBA method set_attribute_config_5. The server will then push Attr_Conf event to any registered clients (**regardless if there was a change in properties or not**).

What we can do:
1. At the very end of `DServer::restart` push events regardless if there was a change or not (consistent with current behavior of set_attribute_config_5)
2. or, before device is deleted, store current property values and after the restart fire events only for modified attributes.(probably more reasonable)

For option 1., it is as simple as:
```cpp
for (Attribute* attr : new_dev->get_device_attr()->get_attribute_list())
{
    new_dev->push_att_conf_event(attr);
}
```

Options 2. should not be much more complex.

What do you think?

lorenzopivetta:
+1 for option 1. Much cleaner, and I think it is reasonable clients get all-fresh events after a device restart... my 2c.

bourtemb:
+1 for option 1 too.

mliszcz:
Thanks Lorenzo and Reynald. I have PR and testcase ready. I'll wait with opening a PR until #573 is merged (it's a dependency required to pass the test and it will trigger merge conflicts).

PR #702, opened 2020-04-06
Backport fix for #496 and #361 - event clients lost after dev restart

mliszcz:
Summary of changes comparing to original commits:
* Correction for #496 (backport of #573):
  * 6c22521 - test only
  * 8712a7a - Event names are passed instead of event type enums
    to avoid changing `set_client_lib` signature
  * 7115111 - I needed to provide a new struct next to EventPar
    and implement a set of overloads that use the new struct
	and attribute_name field instead of attr_id
* Correction for #361 (dependent on previous one, backport of #694):
  * 3950063 - test only
  * b306962 - no changes

### DIFF
--- a/cpp_test_suite/new_tests/cxx_dserver_misc.cpp
+++ b/cpp_test_suite/new_tests/cxx_dserver_misc.cpp
@@ -14,6 +14,7 @@ using namespace std;
 #undef SUITE_NAME
 #define SUITE_NAME DServerMiscTestSuite
 
+template <typename TEvent>
 struct EventCallback : public Tango::CallBack
 {
     EventCallback()
@@ -21,7 +22,7 @@ struct EventCallback : public Tango::CallBack
         , num_of_error_events(0)
     {}
 
-    void push_event(Tango::EventData* event)
+    void push_event(TEvent* event)
     {
         num_of_all_events++;
         if (event->err)
@@ -237,7 +238,7 @@ cout << "str = " << str << endl;
      */
     void test_event_subscription_recovery_after_device_restart()
     {
-        EventCallback callback{};
+        EventCallback<Tango::EventData> callback{};
 
         std::string attribute_name = "event_change_tst";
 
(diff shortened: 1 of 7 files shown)


t-b:
Overall looks good.

The only thing is `get_event_param`/`set_event_param`. According to [1] "You can not ... add an overload (BC, but not SC: makes &func ambiguous), adding overloads to already overloaded functions is ok (any use of &func already needed a cast).". But the compliancy tool says good its good 
[compat-report.zip](https://github.com/tango-controls/cppTango/files/4558286/compat-report.zip).

I would tend to believe the tool's output.

[1]: https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B#The_Do.27s_and_Don.27ts

mliszcz:
@t-b thanks for pointing that out! I focused on binary compatibility but have not thought about source compatibility.

The tool is correct, the change is ABI compatible, because function signature is mangled into symbol's name in C++ ABI. So these two overloads are separate symbols pointing to different places in the code block of the library.

But, as documented in the link you referenced, any device server that does something like this, will stop compiling:
```cpp
auto setter = &Device::set_event_param;
```
While these methods are intended to be internal for the library (not documented in doxygen), nothing prevents users from calling them from their device server code.

Do you think we need to address this somehow?

bourtemb:
> But, as documented in the link you referenced, any device server that does something like this, will stop compiling:
> 
> ```c++
> auto setter = &Device::set_event_param;
> ```
> 
> While these methods are intended to be internal for the library (not documented in doxygen), nothing prevents users from calling them from their device server code.
> 
> Do you think we need to address this somehow?

get_event_param & set_event_param methods are not documented in Doxygen (because under `/// @privatesection` line).

so users are not supposed to use it.
So I think it is a source code incompatibility we can tolerate.

t-b:
If users are not supposed to use it we can break it ;)

PR #703, opened 2020-04-06
Fix #693: pipe events are lost after RestartServer command

### LINKED ISSUE #693: Pipe event subscriptions are lost after RestartServer command
Follow up of #573 / #496

Great work @mliszcz! Thanks!
It seems to work fine with archive, change and periodic events as well as for attribute configuration events and data ready events.
I noticed only one issue with the Pipe events after calling the RestartServer admin device command.
Pipe events seem to work fine after a device restart command.

_Originally posted by @bourtemb in https://github.com/tango-controls/cppTango/pull/573#issuecomment-598194237_

mliszcz:
Fixes #693.

### DIFF
--- a/cppapi/server/dserver.cpp
+++ b/cppapi/server/dserver.cpp
@@ -1993,8 +1993,9 @@ ServerEventSubscriptionState DServer::get_event_subscription_state()
 
             if (events.has_dev_intr_change_event_clients ||
                 ! events.attribute_events.empty())
-
-            result.emplace(device->get_name(), std::move(events));
+            {
+                result.emplace(device->get_name(), std::move(events));
+            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


cppapi/server/dserver.cpp
```diff
@@ -1985,29 +1981,30 @@ void DServer::mcast_event_for_att(std::string &dev_name,std::string &att_name,st
...
+    for (const auto& device_class : class_list)
+    {
+        for (const auto& device : device_class->get_device_list())
+        {
+            auto events = device->get_event_subscription_state();
+
+            if (events.has_dev_intr_change_event_clients ||
+                ! events.attribute_events.empty())
+
+            result.emplace(device->get_name(), std::move(events));
```
t-b: I'm not sure how to read that code.

Should it be

```
            if (events.has_dev_intr_change_event_clients ||
                ! events.attribute_events.empty())
             {
                 result.emplace(device->get_name(), std::move(events));
             }
```
?

t-b (changes requested):
Except one minor code style issue this looks good!

cppapi/server/dserver.cpp
mliszcz: Thanks! Correected as you suggested.

t-b (dismissed):

PR #704, opened 2020-04-10
Correct archive periodic event sending after polling restart

### LINKED ISSUE #675: Archive periodic events no longer received after attribute polling restart
Archive periodic events are no longer sent by the device server if the remote attribute polling is temporary stopped and restarted, archive events based on archive change event thresholds are still sent. It seems like there is no issue with standard PERIODIC events (not archive periodic events).

Here is how you can reproduce the issue with TangoTest device server for instance.

1. Start Tango Test device
2. Set the polling period to 1000 ms on ampli attribute
3. Configure the archive event properties for the ampli attribute with an archive absolute change threshold of 1, and archive period of 3000
4. Start a Tango client (or more) which subscribes to sys/tg_test/1/ampli attribute archive events
5. At this point, the clients receive archive events as expected (every 3s)
6. Stop the polling for ampli attribute
7. The clients receive an API_PollObjNotFound error event
8. Restart the polling for ampli attribute (set to 3000ms for instance)
9. The clients receive 1 archive event for ampli attribute.
10. Wait for some time longer than the archive period event configuration for ampli attribute
11. The archive periodic events are no longer received by the clients/sent by the server after this single event.
12. Set ampli attribute to a different value which should trigger archive change event
13. The clients receive the archive event triggered by the change but the archive periodic events are still not received.

Restarting the device server restores a correct behaviour. Restarting the device sys/tg_test/1 (restart device from jive or DevRestart('sys/tg_test/1') on TangoTest admin device as well as RestartServer command on TangoTest admin device) seems to solve the issue too at some point.

I tested with tango-9-lts branch and 9.3-backports branch, as well with 9.3.3 and the bug seems to be present in all these versions

mliszcz:
Fixes #675.

### DIFF
--- a/cpp_test_suite/new_tests/cxx_dserver_misc.cpp
+++ b/cpp_test_suite/new_tests/cxx_dserver_misc.cpp
@@ -2,9 +2,13 @@
 #define DServerMiscTestSuite_h
 
 #include "cxx_common.h"
+#include <thread>
+#include <chrono>
 
 #undef SUITE_NAME
 #define SUITE_NAME DServerMiscTestSuite
+
+using namespace std::chrono_literals;
 
 template <typename TEvent>
 struct EventCallback : public Tango::CallBack
@@ -286,6 +290,54 @@ cout << "str = " << str << endl;
         TS_ASSERT_EQUALS(0, callback.num_of_error_events);
 
         TS_ASSERT_THROWS_NOTHING(device1->unsubscribe_event(subscription));
+    }
+
+/*
+ * Test to check that client still receives periodic archive events after
+ * polling is stopped and started again for particular device. Such scenario
+ * used to fail as described in #675.
+ */
+    void test_archive_periodic_events_after_polling_restart()
+    {
+        constexpr auto poll_period = 1000ms;
+        constexpr auto time_buffer = 100ms;
+        constexpr auto poll_period_ms = std::chrono::milliseconds(poll_period).count();
+
+        std::string attribute_name = "PollLong_attr";
+
+        auto config = device1->get_attribute_config(attribute_name);
+        config.events.arch_event.archive_period = std::to_string(poll_period_ms);
+        AttributeInfoListEx config_in = { config };
+        TS_ASSERT_THROWS_NOTHING(device1->set_attribute_config(config_in));
+
+        TS_ASSERT_THROWS_NOTHING(device1->poll_attribute(attribute_name, poll_period_ms));
+
+        EventCallback<Tango::EventData> callback{};
+
+        TS_ASSERT_THROWS_NOTHING(device1->subscribe_event(
+            attribute_name,
+            Tango::ARCHIVE_EVENT,
+            &callback));
+
+        std::this_thread::sleep_for(poll_period + time_buffer);
+        TS_ASSERT_EQUALS(0, callback.num_of_error_events);
+        TS_ASSERT_EQUALS(2, callback.num_of_all_events);
+
+        TS_ASSERT_THROWS_NOTHING(device1->stop_poll_attribute(attribute_name));
+
+        std::this_thread::sleep_for(time_buffer);
+        TS_ASSERT_EQUALS(1, callback.num_of_error_events);
+        TS_ASSERT_EQUALS(3, callback.num_of_all_events);
+
+        TS_ASSERT_THROWS_NOTHING(device1->poll_attribute(attribute_name, poll_period_ms));
+
+        std::this_thread::sleep_for(poll_period + time_buffer);
+        TS_ASSERT_EQUALS(1, callback.num_of_error_events);
+        TS_ASSERT_EQUALS(4, callback.num_of_all_events);
+
+        std::this_thread::sleep_for(poll_period);
+        TS_ASSERT_EQUALS(1, callback.num_of_error_events);
+        TS_ASSERT_EQUALS(5, callback.num_of_all_events);
     }
 };
 #undef cout
--- a/cppapi/server/attribute.cpp
+++ b/cppapi/server/attribute.cpp
@@ -4461,6 +4461,7 @@ void Attribute::fire_archive_event(DevFailed *except)
 #else
 			gettimeofday(&now_timeval,NULL);
 #endif
+			now_timeval.tv_sec = now_timeval.tv_sec - DELTA_T;
 
 //
 // Eventually push the event (if detected)


cpp_test_suite/new_tests/cxx_dserver_misc.cpp
```diff
@@ -255,6 +255,50 @@ cout << "str = " << str << endl;
...
+
+        Tango_sleep(2);
+        TS_ASSERT_EQUALS(1, callback.num_of_error_events);
+        TS_ASSERT_EQUALS(3, callback.num_of_all_events);
+
+        TS_ASSERT_THROWS_NOTHING(device1->poll_attribute(attribute_name, 1500));
+
+        Tango_sleep(2);
+        TS_ASSERT_EQUALS(1, callback.num_of_error_events);
+        TS_ASSERT_EQUALS(4, callback.num_of_all_events);
```
bourtemb, line 336: It looks like we could get 5 events received in some cases here, depending on the synchronization of the polling thread.
At line 291, the attribute polling is restarted so it is polled every 1.5 s.
At line 293, we sleep 2 seconds.
During these 2 seconds, the polling thread has potentially the time to read twice the attribute, so, to send 2 events, isn't it?
I got 1 test failure like this once when I tried to reproduce it on an Ubuntu 18.04 VM:
`/home/reynald/GIT/mliszcz/cppTango/cpp_test_suite/cxxtest/include/../../new_tests/cxx_dserver_misc.cpp:295: Error: Expected (4 == callback.num_of_all_events), found (4 != 5)`

bourtemb:
As I wrote in another comment, I tried to execute the new test on an Ubuntu 18.04 VM.
I got the problem I mentioned earlier and I got also some failures due to the same problem as https://github.com/tango-controls/cppTango/issues/524

`/home/reynald/GIT/mliszcz/cppTango/cpp_test_suite/cxxtest/include/../../new_tests/cxx_dserver_misc.cpp:243: Error: Expected (2 == callback.num_of_all_events), found (2 != 1)`

I also got a crash once:

```
Program terminated with signal SIGSEGV, Segmentation fault.

(gdb) bt
#0  0x00007fa96ca8d760 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007fa96dda965f in Tango::ZmqEventConsumer::push_zmq_event (this=this@entry=0x55c7c3cfc020, ev_name="tango://172.17.0.5:10000/test/debian8/10/polllong_attr.idl5_archive", endian=endian@entry=1 '\001', event_data=..., error=<optimized out>, 
    ds_ctr=<optimized out>) at /home/reynald/GIT/mliszcz/cppTango/cppapi/client/zmqeventconsumer.cpp:2597
#2  0x00007fa96ddadaa4 in Tango::ZmqEventConsumer::process_event (this=this@entry=0x55c7c3cfc020, received_event_name=..., received_endian=..., received_call=..., event_data=...)
    at /home/reynald/GIT/mliszcz/cppTango/cppapi/client/zmqeventconsumer.cpp:591
#3  0x00007fa96ddaefa8 in Tango::ZmqEventConsumer::run_undetached (this=0x55c7c3cfc020, arg=<optimized out>) at /home/reynald/GIT/mliszcz/cppTango/cppapi/client/zmqeventconsumer.cpp:317
#4  0x00007fa96d2d74b8 in omni_thread_wrapper () from /usr/lib/x86_64-linux-gnu/libomnithread.so.4
#5  0x00007fa96d8936db in start_thread (arg=0x7fa968fb8700) at pthread_create.c:463
#6  0x00007fa96c58188f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) f 1
#1  0x00007fa96dda965f in Tango::ZmqEventConsumer::push_zmq_event (this=this@entry=0x55c7c3cfc020, ev_name="tango://172.17.0.5:10000/test/debian8/10/polllong_attr.idl5_archive", endian=endian@entry=1 '\001', event_data=..., error=<optimized out>, 
    ds_ctr=<optimized out>) at /home/reynald/GIT/mliszcz/cppTango/cppapi/client/zmqeventconsumer.cpp:2597
2597                                        callback->push_event(event_dat);
(gdb) p *event_dat
$2 = {<Tango::EventData> = {reception_date = {tv_sec = 1586868647, tv_usec = 531199, tv_nsec = 0}, device = 0x55c7c3ceab70, attr_name = "tango://172.17.0.5:10000/test/debian8/10/polllong_attr", event = "archive", attr_value = 0x55c7c3d0b980, 
    err = false, errors = {<_CORBA_Unbounded_Sequence<Tango::DevError>> = {<_CORBA_Sequence<Tango::DevError>> = {pd_max = 0, pd_len = 0, pd_rel = true, pd_bounded = false, pd_buf = 0x0}, <No data fields>}, <No data fields>}}, av_5 = 0x0, 
  event_data = 0x0}


```

mliszcz:
Thanks @bourtemb!
Today I was working on some other project (non-Tango) but I plan to get back to this issue still this week. I agree with your comment regarding the test - I can definitely improve it and make it more robust.

mliszcz:
Hi @bourtemb I reworked the test case as you suggested (now polling period is 1000ms and I'm waiting just 100ms for the events to be fired).

I also started replacing time calculations in polling thread with std::chrono, but it is more challenging than I initially thought. We have five different types to represent time in Tango (time_t, timeval, long (ms), double (s), Tango::TimeVal) and we convert between them a lot in the whole codebase. I already updated most of the code but I still have some troubles with aligning pollthread.cpp. I propose to review and merge this PR as a simple bugfix and I'll send a separate PR for replacing all those timevals with chrono.

cpp_test_suite/new_tests/cxx_dserver_misc.cpp
mliszcz: Changed. Now it is polled every 1s and after polling period I wait events for 100ms (so it is not possible for this buffer to overlap with next polling period).

andygotz:
@mliszcz I understand you are moving to using chrono which sounds like a good idea (is it the date and time standard in C++?). 

What is the time resolution of the time with chrono? Does it support sub-microsecond? It would be great to be able to support time with the precision of the local timing system which could be nanoseconds.

mliszcz:
Hi @andygotz 
C++ standard does not mandate any precision for chrono clocks, it is up to the library implementation.
The implementations I've checked are using clock source with highest precision available on given platform. On GNU systems, this means `clock_settime(CLOCK_REALTIME, ...)` which indeed has nanosecond precision (can confirm, just did a test on Ubuntu 18.04). But if your system (glibc) does not support `clock_settime`, they fall back to `gettimeofday()` (microsecond precision) or even `time(NULL)` which holds just seconds.

Here are relevant implementations of `std::chrono::system_clock::now()` from GNU's libstdc++ and from LLVM's libc++:
https://github.com/gcc-mirror/gcc/blob/releases/gcc-10.1.0/libstdc++-v3/src/c++11/chrono.cc#L51-L74
https://github.com/llvm/llvm-project/blob/llvmorg-10.0.0/libcxx/src/chrono.cpp#L53-L92

The main reason I started porting to chrono instead of using manual operation on timestamp structures is not to increase precision, but to fix any possible bugs we may have due to code like below being scattered all over the codebase (it is very easy to forget about subtracting DELTA_T or put one zero less or more when converting time).
https://github.com/tango-controls/cppTango/blob/9.3.3/cppapi/server/device_3.cpp#L1254-L1267

cpp_test_suite/new_tests/cxx_dserver_misc.cpp
```diff
@@ -287,6 +291,54 @@ cout << "str = " << str << endl;
...
     }
+
+/*
+ * Test to check that client still receives periodic archive events after
+ * polling is stopped and started again for particular device. Such scenario
+ * used to fail as described in #675.
+ */
+    void test_archive_periodic_events_after_polling_restart()
+    {
+        constexpr auto poll_period = 1000ms;
```
bourtemb, line 302: std::chrono_literals::ms is C++14 standard, right? (https://en.cppreference.com/w/cpp/chrono/operator%22%22ms)
If tango-9-lts branch requires C++14 standard, we should probably upgrade the CMakeLists.txt file (https://github.com/tango-controls/cppTango/blob/tango-9-lts/CMakeLists.txt#L5) to impose at least C++14 standard, or at least for the tests part (maybe it's already there but I didn't see it).

bourtemb (commented):
Looks good to me!
Just a comment on C++14 requirement.
The use case described in #675 seems to work fine now and the new test seems to pass more reliably now!
Good initiative to create #729, we should really try to use the std::chrono everywhere we are dealing with timestamps.

cpp_test_suite/new_tests/cxx_dserver_misc.cpp
t-b: In https://github.com/tango-controls/cppTango/issues/626#issue-523572962 we thought about enforcing C++14 around 8/2020 in cppTango.

I would prefer having the same requirements for the tests and cppapi, as that would at least theoretically allow us to get the tests running on windows as well at some point.

mliszcz: Thanks @bourtemb and @t-b for your comments. C++14 is not mandatory for this test, changing literals to explicit duration constructors is trivial. But since we agreed to move to C++14 now, let me mark your comments as resolved without changing the test.

Here is PR I opened to require C++14: https://github.com/tango-controls/cppTango/pull/732.

t-b (approved):

bourtemb (approved):

PR #716, opened 2020-04-30
Fix nb_endpoints in ZmqEventConsumer::connect_event_channel

bourtemb:
Since cppTango 9.3.0, ZmqEventSubscriptionChange is returning 2 extra
argout string values after the list of endpoints.
This patch proposes to take this correctly into account in
ZmqEventConsumer::connect_event_channel() method
This missing part was preventing some clients to successfully subscribe to
events in some specific network configurations on the server side (with server tango version >= 9.3.0).
For instance when the ORBendPoint was used on the server side using a host string value in the definition of the ORBendPoint (e.g. giop:tcp:myhostname:).
In this specific case ZmqEventSubscriptionChange is returning endpoints including the hostname, not IP addresses.
The bug was not observed when the host part of ORBendPoint was specified with
an IP address (In this case check_zmq_endpoint() was working properly).

### DIFF
--- a/cppapi/client/zmqeventconsumer.cpp
+++ b/cppapi/client/zmqeventconsumer.cpp
@@ -1294,21 +1294,29 @@ void ZmqEventConsumer::connect_event_channel(string &channel_name,TANGO_UNUSED(D
 // If the server has returned several possible ZMQ endpoints (because several NIC boards on server host), check which
 // one is correct
 //
+    size_t nb_endpoints = ev_svr_data->svalue.length();
 
-    size_t nb_endpoint = ev_svr_data->svalue.length();
-    nb_endpoint = nb_endpoint >> 1;
+    Tango::DevLong server_tango_lib_version = ev_svr_data->lvalue[0];
+    if(server_tango_lib_version >= 930)
+    {
+        // ZmqEventSubscriptionChange returns the ZMQ message filters used for the attribute
+        // and for the heartbeat events at the end of the string array part of the returned DevVarLongStringArray
+        nb_endpoints -= 2;
+    }
+
+    nb_endpoints = nb_endpoints >> 1;
     size_t valid_endpoint = 0;
 
-    if (nb_endpoint != 1)
+    if (nb_endpoints != 1)
     {
-        for (valid_endpoint = 0;valid_endpoint < nb_endpoint;valid_endpoint++)
+        for (valid_endpoint = 0; valid_endpoint < nb_endpoints; valid_endpoint++)
         {
             string endpoint(ev_svr_data->svalue[valid_endpoint << 1]);
             if (check_zmq_endpoint(endpoint) == true)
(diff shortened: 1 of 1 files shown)


cppapi/client/zmqeventconsumer.cpp
```diff
@@ -1294,21 +1294,29 @@ void ZmqEventConsumer::connect_event_channel(string &channel_name,TANGO_UNUSED(D
...
 
-    if (nb_endpoint != 1)
+    if (nb_endpoints != 1)
     {
-        for (valid_endpoint = 0;valid_endpoint < nb_endpoint;valid_endpoint++)
+        for (valid_endpoint = 0; valid_endpoint < nb_endpoints; valid_endpoint++)
         {
             string endpoint(ev_svr_data->svalue[valid_endpoint << 1]);
             if (check_zmq_endpoint(endpoint) == true)
                 break;
```
mliszcz, line 1316: In the scenario you described in the original post, was channel name (last element on the list) used as zmq endpoint?
If yes, then:
1. It means that no other endpoint on the list passes the `check_zmq_endpoint` check- client won't be able to receive events anyway.
2. Maybe the `check_zmq_endpoint` should be improved to reject invalid values? I think channel name starts with 'tango://', right?

bourtemb:
> Hi @bourtemb.
> 
> I'm trying to understand better the root cause of the issue and the solution you propose.
> 
> Do you know why we are sending these extra parameters if it is safe to discard them as you did (I'm assuming this is event topic and channel name as included at the very end of `DServer::zmq_event_subscription_change`)? Can't we just not send them at all?

This is indeed event topic and channel name. We really need them to solve potential issues when using network aliases for instance.
This is required to ensure the client and the server are using the same ZMQ channel names (filters/topics).

I am not discarding them completely in this proposal.
They are used somewhere else before (in ZmqEventConsumer::initialize_received_from_admin() which is called from EventConsumer::connect_event

cppapi/client/zmqeventconsumer.cpp
bourtemb: > In the scenario you described in the original post, was channel name (last element on the list) used as zmq endpoint?

Yes

> If yes, then:
> 
>     1. It means that no other endpoint on the list passes the `check_zmq_endpoint` check- client won't be able to receive events anyway.

Indeed

>     2. Maybe the `check_zmq_endpoint` should be improved to reject invalid values? I think channel name starts with 'tango://', right?

The problem is that we pass values which are not endpoints. My proposal is to avoid that.
We have a way to know whether the 2 last values are endpoints or not thanks to the remote server tango lib version also returned as result of ZmqEventSubscriptionChange command.

If we improve check_zmq_endpoint, we could improve it so endpoint containing hostnames would be handled properly, because as it is now, check_zmq_endpoint is expecting an IP address in the given end point.

PR #717, opened 2020-05-08
.travis/gcc-latest/Dockerfile: Update to 10.1.0

t-b:


mliszcz:
Hi @t-b,

I re-started the job (there was some temporary failure when pulling the gcc image) and now it fails with:

```
/home/tango/src/cpp_test_suite/asyn/asyn_write_cb.cpp: In function 'int main(int, char**)':
/home/tango/src/cpp_test_suite/asyn/asyn_write_cb.cpp:321:30: error: suggest parentheses around assignment used as truth value [-Werror=parentheses]
  321 |   assert ( cb.faulty_attr_nb = 1);
      |            ~~~~~~~~~~~~~~~~~~^~~
/home/tango/src/cpp_test_suite/asyn/asyn_write_cb.cpp:352:22: error: suggest parentheses around assignment used as truth value [-Werror=parentheses]
  352 |   assert( cb.nb_attr = 1);
      |           ~~~~~~~~~~~^~~
/home/tango/src/cpp_test_suite/asyn/asyn_write_cb.cpp:370:22: error: suggest parentheses around assignment used as truth value [-Werror=parentheses]
  370 |   assert( cb.nb_attr = 1);
      |           ~~~~~~~~~~~^~~
```

Looks like we need to fix the tests before upgrading GCC.

t-b:
@mliszcz Thanks, fixed.

bourtemb:
> Note that for some reason push job is failing (github / travis issue? Tried restarting but didn't help).

Very strange.
It looks like Travis is trying to get upgrade-to-latest-gcc branch from tango-controls instead of from t-b fork.

t-b:
I pushed to tango-controls repo first and then deleted the branch again. Therefore the invalid push job.

PR #721, opened 2020-05-28
Fix endpoints in event connect channel [forwardport of #716]

t-b:
Since cppTango 9.3.0, ZmqEventSubscriptionChange is returning 2 extra
argout string values after the list of endpoints.
This patch proposes to take this correctly into account in
ZmqEventConsumer::connect_event_channel() method
This missing part was preventing some clients to successfully subscribe to
events in some specific network configurations on the server side (with server tango version >= 9.3.0).
For instance when the ORBendPoint was used on the server side using a host string value in the definition of the ORBendPoint (e.g. giop:tcp:myhostname:).
The bug was not observed when the host part of ORBendPoint was specified with
an IP address (In this case check_zmq_endpoint() was working properly).

### DIFF
--- a/cppapi/client/zmqeventconsumer.cpp
+++ b/cppapi/client/zmqeventconsumer.cpp
@@ -1261,21 +1261,29 @@ void ZmqEventConsumer::connect_event_channel(std::string &channel_name,TANGO_UNU
 // If the server has returned several possible ZMQ endpoints (because several NIC boards on server host), check which
 // one is correct
 //
+    size_t nb_endpoints = ev_svr_data->svalue.length();
 
-    size_t nb_endpoint = ev_svr_data->svalue.length();
-    nb_endpoint = nb_endpoint >> 1;
+    Tango::DevLong server_tango_lib_version = ev_svr_data->lvalue[0];
+    if(server_tango_lib_version >= 930)
+    {
+        // ZmqEventSubscriptionChange returns the ZMQ message filters used for the attribute
+        // and for the heartbeat events at the end of the string array part of the returned DevVarLongStringArray
+        nb_endpoints -= 2;
+    }
+
+    nb_endpoints = nb_endpoints >> 1;
     size_t valid_endpoint = 0;
 
-    if (nb_endpoint != 1)
+    if (nb_endpoints != 1)
     {
-        for (valid_endpoint = 0;valid_endpoint < nb_endpoint;valid_endpoint++)
+        for (valid_endpoint = 0; valid_endpoint < nb_endpoints; valid_endpoint++)
         {
             std::string endpoint(ev_svr_data->svalue[valid_endpoint << 1]);
             if (check_zmq_endpoint(endpoint) == true)
                 break;
         }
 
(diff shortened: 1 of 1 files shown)


mliszcz:
Hi @t-b. All travis jobs are failed due to the following issue:
```
xports.def : warning LNK4022: cannot find unique match for symbol '__' [C:\projects\cppTango\tango.vcxproj]
exports.def : warning LNK4002: __@@_PchSym_@00@UkilqvxghUxkkgzmtlUxkkzkrUhvieviUhvieviPlyqvxghOwriUivovzhvUkxsPhvieviPlyqvxghOlyq@4B2008FD98C1DD4 defined in C:\projects\cppTango\cppapi\server\server_objects.dir\Release\pch_server_objects.obj [C:\projects\cppTango\tango.vcxproj]
exports.def : warning LNK4002: __@@_PchSym_@00@UkilqvxghUxkkgzmtlUxkkzkrUxorvmgUxorvmgPlyqvxghOwriUivovzhvUkxsPxorvmgPlyqvxghOlyq@4B2008FD98C1DD4 defined in C:\projects\cppTango\cppapi\client\client_objects.dir\Release\pch_client_objects.obj [C:\projects\cppTango\tango.vcxproj]
exports.def : error LNK2001: unresolved external symbol __ [C:\projects\cppTango\tango.vcxproj]
C:/projects/cppTango/Release/tango.lib : fatal error LNK1120: 1 unresolved externals [C:\projects\cppTango\tango.vcxproj]
Command exited with code 1
```
I have no idea why we see this error. We have special handling in PCH builds to prevent this issue:
https://github.com/tango-controls/cppTango/blob/tango-9-lts/cppapi/client/CMakeLists.txt#L66

I don't have permissions to restart the job. Can you do this? Other PRs are building fine on appveyor.

t-b:
@mliszcz  s/travis/appveyor ?

This looks like some fallout from #688, maybe you need TANGO_BUILD_SHARED  in that if clause instead of BUILD_SHARED_LIBS?

I can't restart those btw.

mliszcz:
Hi @t-b 
Sure, you are right, I meant appveyor jobs.

It may be related to #688 indeed. I noticed one strange thing. It looks like we are using old (pre-#688) appveoyr.yml (e.g. it prints `md debug_build` instead of `md static_build`) but new CMakeLists files (e.g. it prints `Tango library is 'tango'` which was added in #688).

mliszcz:
Not confirmed but think this is how appveyor 'pr' jobs are working:
* first we run init step from the yaml file,
* then we clone merged branch:
  ```
  cd "C:\projects\cppTango"
  git clone -q https://github.com/tango-controls/cppTango.git C:\projects\cppTango
  git fetch -q origin +refs/pull/721/merge:
  git checkout -qf FETCH_HEAD
  ```
* then we continue with install step from the original yaml file

I think that re-basing the PR source branch on top of the target branch should help.

t-b:
@mliszcz I've rebased the PR.

mliszcz (approved):
Thanks! LGTM now.

bourtemb (approved):

PR #723, opened 2020-06-11
Testing github workflows

t-b:


.github/workflows/c-cpp.yml
```diff
@@ -0,0 +1,23 @@
+name: C/C++ CI
+
+on:
+  push:
+    branches: [ master ]
```
bourtemb, line 5: I think you need to enable that for some other branches too

PR #729, opened 2020-06-24
Use std::chrono in polling calculations

### LINKED ISSUE #675: Archive periodic events no longer received after attribute polling restart

mliszcz:
We have five different types to represent time in Tango (time_t, timeval, long (ms), double (s), Tango::TimeVal) and we convert between them a lot in the whole codebase. Sometimes we also need to shift timestamps by DELTA_T to avoid overflows. All this is error-prone and requires a lot of duplicated code.

This changes polling thread and related classes to operate on std::chrono `duration`s and `time_point`s .

Follow-up of #704.
Fixes #675.

### DIFF
--- a/cpp_test_suite/cpp_test_ds/DevTest.cpp
+++ b/cpp_test_suite/cpp_test_ds/DevTest.cpp
@@ -1,5 +1,6 @@
 #include <tango.h>
 #include <DevTest.h>
+#include <tango_clock.h>
 
 #ifdef WIN32
 #include <sys/timeb.h>
@@ -1848,30 +1849,13 @@ void DevTest::read_slow_actuator(Tango::Attribute &att)
 	slow_actua++;
 	att.set_value(&slow_actua);
 
-	struct timeval now;
-#ifdef WIN32
-	struct _timeb before_win;
-	_ftime(&before_win);
-	now.tv_sec = (unsigned long)before_win.time;
-	now.tv_usec = (long)before_win.millitm * 1000;
-#else
-	gettimeofday(&now,NULL);
-#endif
-	long delta;
 	if (slow_actua_write.tv_sec != 0)
 	{
+		auto now = std::chrono::system_clock::now();
+		auto delta = now - Tango::make_system_time(slow_actua_write);
 
-#define COMPUTE_TIME_DIFF(RESULT,BEFORE,AFTER) \
-long bef = ((BEFORE.tv_sec - 1095000000) * 1000) + (BEFORE.tv_usec / 1000); \
-long after = ((AFTER.tv_sec - 1095000000) * 1000) + (AFTER.tv_usec / 1000); \
-RESULT = after - bef;
-
-		COMPUTE_TIME_DIFF(delta,slow_actua_write,now);
-
-#undef COMPUTE_TIME_DIFF
-
-		cout << "Delta time = " << delta << std::endl;
-		if (delta > 3000)
+		cout << "Delta time = " << Tango::duration_ms(delta) << std::endl;
+		if (delta > std::chrono::milliseconds(3000))
 		{
 			att.set_quality(Tango::ATTR_VALID);
 			slow_actua_write.tv_sec = 0;
--- a/cpp_test_suite/cpp_test_ds/DevTest.cpp
+++ b/cpp_test_suite/cpp_test_ds/DevTest.cpp
@@ -1860,7 +1860,16 @@ void DevTest::read_slow_actuator(Tango::Attribute &att)
 	long delta;
 	if (slow_actua_write.tv_sec != 0)
 	{
+
+#define COMPUTE_TIME_DIFF(RESULT,BEFORE,AFTER) \
+long bef = ((BEFORE.tv_sec - 1095000000) * 1000) + (BEFORE.tv_usec / 1000); \
+long after = ((AFTER.tv_sec - 1095000000) * 1000) + (AFTER.tv_usec / 1000); \
+RESULT = after - bef;
+
 		COMPUTE_TIME_DIFF(delta,slow_actua_write,now);
+
+#undef COMPUTE_TIME_DIFF
+
 		cout << "Delta time = " << delta << std::endl;
 		if (delta > 3000)
 		{
--- a/cpp_test_suite/old_tests/size.cpp
+++ b/cpp_test_suite/old_tests/size.cpp
@@ -1,4 +1,4 @@
-/* 
+/*
  * Check size of user classes/structure (For compatibility prurpose)
  *
  * This test suite has to be run on a redhate4 host
@@ -31,17 +31,17 @@ int main(int argc, char **argv)
 	check_size("Attr",sizeof(Attr),128, &isOK);  // Was 52 in V4
 	check_size("SpectrumAttr",sizeof(SpectrumAttr),144, &isOK);	// Was 60 in V4
 	check_size("ImageAttr",sizeof(ImageAttr),160, &isOK);	// Was 68 in V4
-		
+
 	check_size("DeviceImpl",sizeof(DeviceImpl),1760, &isOK);
 	check_size("Device_2Impl",sizeof(Device_2Impl),1776, &isOK);
-	
+
 	check_size("DeviceClass",sizeof(DeviceClass),416, &isOK);  	// Was 48 in V4
-	check_size("Util",sizeof(Util),1616, &isOK);	// Was 68 in V4
+	check_size("Util",sizeof(Util),1624, &isOK);	// Was 68 in V4
 
 	check_size("Attribute",sizeof(Attribute),2536, &isOK);		// Was 208 in V4
 	check_size("WAttribute",sizeof(WAttribute),3200, &isOK);	// Was 252 in V4
 	check_size("MultiAttribute",sizeof(MultiAttribute),80, &isOK);
-	
+
 	check_size("Command",sizeof(Command),72, &isOK);
 	check_size("TemplCommand",sizeof(TemplCommand),112, &isOK);
 	check_size("TemplCommandIn",sizeof(TemplCommandIn<int>),136, &isOK);
@@ -61,7 +61,7 @@ int main(int argc, char **argv)
 	check_size("DbServer",sizeof(DbServer),32, &isOK);
 	check_size("DbDatum",sizeof(DbDatum),56, &isOK);
 	check_size("DbData",sizeof(DbData),24, &isOK);
-	
+
 	check_size("CmdDoneEvent",sizeof(CmdDoneEvent),48, &isOK);
 	check_size("AttrReadEvent",sizeof(AttrReadEvent),48, &isOK);
 	check_size("AttrWrittenEvent",sizeof(AttrWrittenEvent),40, &isOK);
--- a/cppapi/client/Connection.h
+++ b/cppapi/client/Connection.h
@@ -30,7 +30,9 @@
 #ifndef _CONNECTION_H
 #define _CONNECTION_H
 
+#include <tango_optional.h>
 
+#include <chrono>
 
 /****************************************************************************************
  * 																						*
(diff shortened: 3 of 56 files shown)


mliszcz:
Hi @bourtemb and @t-b 
The PR is ready for review. I pushed it a bit further than I initially planned and removed all references to `gettimeofday` from `cppapi` (but tests still use it). Most of the changes are straightforward and compiler should catch any issues. Still, I recommend to review commit-wise, especially 59c2a07 which is quite large.

We still have quite a lot of `time(NULL)` calls, especially for detecting if we still have registered event subscribers. I did not change them because it is a lot of work and there is little (if any) benefit of using chrono instead of operating directly on integral number of seconds (especially that we are not doing any conversions, just calculate deltas).

@andygotz you may be interested in this PR as well as it adds nanosecond precision to timestamps reported when you read an attribute or receive an event (time.tv_nsec is nonzero if operating system provides that). Polling thread now operates internally on nanoseconds as well (but you still need to configure polling period for your devices using milliseconds only).

t-b (commented):
46bc9544 (Add tango_optional<T> class, 2020-06-24)

Nice!

200d7397 (Make sleep time optional in pollthread, 2020-06-24)

- The comment of get_command() talks about the function returning true/false. I don't think this is the case.

```
-               cout5 << "Sleep for : " << sleep << std::endl;
+        if (sleep.has_value())
+        {
+            cout5 << "Sleep for : " << *sleep << std::endl;
+        }
+        else
+        {
+            cout5 << "Sleep for : -1 (undefined)" << std::endl;
+        }
```

That code asks for an output operator for tango_optional, something like

```
template <typename T>
std::ostream & operator<<(std::ostream& out, const tango_optional<T> &opt)
{
  if(opt.has_value())
  {
    out << *opt;
  }
  else
  {
    out << "(undefined)";
  }

  return out;
}
```

maybe? 


d52503db (Add poll_clock.h, 2020-06-24)

- I think "detail" is enough, names starting with two underscores are reserved
  (https://stackoverflow.com/a/228797).
- timeval/TimeVal maybe pass by const reference?
- make_TimeVal 1000.0 -> 1000 or add an explicit cast on the result if you want to do floating point division

59c2a07d (Use std::chrono in time calculations in poll thread, 2020-06-25)

That is a tough one! I would say it looks good!

```
#ifdef _TG_WINDOWS_
	#include <sys/types.h>
	#include <sys/timeb.h>
#endif
```

can go or?

-I think code like

```
std::chrono::nanoseconds(local_cmd.new_upd).count() / 1e6 
```

is a bit difficult to grasp. IIUIC it means give me the duration in milliseconds but as double with fraction.
Would we benefit having dedicated helpers for that? If you are doing that please in new commit. This one is really huge already.

4d4cef47 (Fix indentation a bit in pollthread.h and pollring.h, 2020-06-25)

Good!

fc9067ff (Merge PollThread::auto_upd and auto_name into pair, 2020-06-25)

Nice!

30ebfe6e (Add function for calculating event reporintg period, 2020-06-25)

- Typo in commit message

Nice cleanup!

82623416 (Change LockThread to use std::chrono, 2020-06-25)

```
#ifndef _TG_WINDOWS_
#include <sys/time.h>
#endif
```
can go?

c7f4b807 (Remove unused constants, 2020-06-25)

Cool!

4fcaa929 (Simplify some calculations in pollthread.cpp, 2020-06-25)


Why does

```
    auto n = std::uint64_t((next_work - prev_obj_work) / ite_prev->update);
    next_prev = prev_obj_work + (n * ite_prev->update);
```

use uint64_t instead of letting auto figure out the type?

b1210bf0 (Further simplify time calc in pollthread.cpp, 2020-06-25)

`after - next < DISCARD_THRESHOLD` -> `(after - next) < DISCARD_THRESHOLD)

dfd0f555 (Update size test, 2020-06-25)

Obviously correct according to CI.

111981f3 (Increase precision reported in TimeVal to nanoseconds, 2020-06-25)

Good!

e712daf1 (Make poll_clock.h more robust, 2020-06-26)

Perfect!

efb86cbb (Rename poll_clock.h to tango_clock.h, 2020-06-26)

Fine, in principle it is nice to merge with the original commit. But git has
lousy conflict resolution in this case, so this can stay.

f4ca96fd (Calculate some timestamps using chrono, 2020-06-26)

Looks good. But how about adding make_TimeVal_Now()?

e51329f0 (Change zmqeventsupplier to use chrono, 2020-06-26)

`now - cc.date > std::chrono::seconds(500)` -> `(now - cc.date) > std::chrono::seconds(500)` 

807c97cd (Use chrono in ping delay calculation, 2020-06-26)

Goodie!

be477d38 (Use chrono for DeviceProxy reconnection, 2020-06-26)

Nice to see!

d9a577cc (Port w_attribute to chrono, 2020-06-26)

Nothing to muse about.

dc09b082 (Port pipe.h/.cpp to chrono, 2020-06-26)

+1

17704207 (Port BlackBox to chrono, 2020-06-26)

+1!

I was not able to figure out how TimeAttrData/TimedCmdData is used as nobody is using it.

Are we already there for general nanosecond support? Would we need more than overloads which currently take a timeval/timeb?

mliszcz:
Thanks @t-b for a detailed review!

> [200d739](https://github.com/tango-controls/cppTango/commit/200d7397e80c9768847a2fa1445a480ab9bb22ee) (Make sleep time optional in pollthread, 2020-06-24)
>     * The comment of get_command() talks about the function returning true/false. I don't think this is the case.

Corrected in https://github.com/tango-controls/cppTango/pull/729/commits/595db5c8f3e45a6c42f41b3a2c70f8d6edeb3527

> [d52503d](https://github.com/tango-controls/cppTango/commit/d52503dbd1f92e8e792c4a551678584654040f13) (Add poll_clock.h, 2020-06-24)
>     * I think "detail" is enough, names starting with two underscores are reserved
>       (https://stackoverflow.com/a/228797).

Thanks! Didn't know about that. Btw according to the link this rule only applies to the names in global scope (which is not true for Tango::__detail). Anyway, I changed it to detail as it indeeds reads better.

>     * timeval/TimeVal maybe pass by const reference?

Small types (2x 64bit or less) like timeval should be passed by value, because they are passed via CPU registers. The same is true for types like std::string_view or std::span. Also, value semantics are default in C++, putting extra references and consts does not improve readability. 

As for TimeVal, it has 3 components (sec + usec + nsec), so it is passed via stack. I don't know which version will be faster (we'd need to benchmark, and it probably depends on what you have available at the call site). I've changed it to use const reference as you suggested.

>     * make_TimeVal 1000.0 -> 1000 or add an explicit cast on the result if you want to do floating point division

Corrected in subsequent commits.
 
> ```
> #ifdef _TG_WINDOWS_
> 	#include <sys/types.h>
> 	#include <sys/timeb.h>
> #endif
> ```
> 
> can go or?

Thanks for this suggestion. I removed such includes in the latest commit.

> 
> -I think code like
> 
> ```
> std::chrono::nanoseconds(local_cmd.new_upd).count() / 1e6 
> ```
> 
> is a bit difficult to grasp. IIUIC it means give me the duration in milliseconds but as double with fraction.
> Would we benefit having dedicated helpers for that? If you are doing that please in new commit. This one is really huge already.

I added two helper functions for converting duration to either seconds or milliseconds (as double).

> [4fcaa92](https://github.com/tango-controls/cppTango/commit/4fcaa9298a3bec8cf8090c54b7ec62cd088e233b) (Simplify some calculations in pollthread.cpp, 2020-06-25)
> 
> Why does
> 
> ```
>     auto n = std::uint64_t((next_work - prev_obj_work) / ite_prev->update);
>     next_prev = prev_obj_work + (n * ite_prev->update);
> ```
> 
> use uint64_t instead of letting auto figure out the type?

Because we need to wait for the whole number of update periods in the polling thread algorithm. This is described in a comment few lines above. Without casting to integer, the type of `n` may be deduced to double (because `duration` can use double as its `rep` type). Without rounding we will always calculate `next_prev = next_work` (just try substituting the first line into the second one and do the math).

> [b1210bf](https://github.com/tango-controls/cppTango/commit/b1210bf06cd3f48099837f2d692223335cab28b3) (Further simplify time calc in pollthread.cpp, 2020-06-25)
> 
> `after - next < DISCARD_THRESHOLD` -> `(after - next) < DISCARD_THRESHOLD)

Fixed, here and in a few other places.

> [dfd0f55](https://github.com/tango-controls/cppTango/commit/dfd0f555eda9bb55b4de73b488962976ce151899) (Update size test, 2020-06-25)
> 
> Obviously correct according to CI.

IMO we should remove this test. Probably author wanted it to be some sort of ABI compatibility checker but of course it is incorrect (e.g. reorder the members and the size stays the same).

> [f4ca96f](https://github.com/tango-controls/cppTango/commit/f4ca96fdf0850902882a74f468551ca39668979d) (Calculate some timestamps using chrono, 2020-06-26)
> 
> Looks good. But how about adding make_TimeVal_Now()?

It's not that simple, because current time (now) depends on what clock you're looking at. So we'd either need `make_TimeVal_system_now()` or `make_TimeVal_now<std::chrono::system_clock>()`.

> [d9a577c](https://github.com/tango-controls/cppTango/commit/d9a577cc4749edc2825e14238cdfa7664e5da8fe) (Port w_attribute to chrono, 2020-06-26)
> 
> Nothing to muse about.

I removed that macro in https://github.com/tango-controls/cppTango/pull/729/commits/d091b366e94929b7f6ec5c8f95930df7a2c6361e

> I was not able to figure out how TimeAttrData/TimedCmdData is used as nobody is using it.

No idea what it is but it is part of public API (documented in Doxygen) so I assumed it is better not to change anything there.

> Are we already there for general nanosecond support? Would we need more than overloads which currently take a timeval/timeb?

For attribute read timestamps, yes we are. There is `Attribute::set_date(TimeVal)` which has nanosecond precision and Attribute uses this `TimeVal` to store readout time internally. So if you call that instead of `timeval` overload, you'll get nanoseconds.

For polling periods, no. We internally operate on nanosecond numbers but you configure attribute's polling period as a whole number of milliseconds (this is stored in the db). Probably there is no point in changing that as we won't go below 1ms period due to performance reasons anyway.

mliszcz:
Regarding the proposal to add operator<< for optional, I think it is a good idea in general to have that, but I changed the log message to clearly indicate that we are not going to sleep at the next iteration of the loop. I propose we add this operator later when we need it (in case we'd need the same log message regardless of the presence of the value).

t-b (dismissed):
@mliszcz Thanks for the reply and the fixes. Looks good now!

t-b (approved):

PR #730, opened 2020-06-25
Automate abi api compliance check

t-b:
Backport of #662.


.travis/check-ABI-API-compliance.sh
```diff
@@ -0,0 +1,59 @@
...
+function generate_info() {
+  local prefix=$1
+  local revision=$2
+
+  git worktree remove $prefix-branch || true
+  git -c advice.detachedHead=false worktree add ${prefix}-branch ${revision}
+  mkdir ${prefix}-branch/build
+  cd ${prefix}-branch/build
+  cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_TESTING=OFF -DUSE_PCH=ON ..
+  cmake --build .
```
mliszcz, line 34: Consider adding `--target tango` to avoid building `tango-static` (and other things we may add in the future).

mliszcz (commented):
LGTM, but I'm wondering if we are not missing anything here. Our report looks like:
image
But other people have more information here (headers, sources, non-zero number of symbols):
* https://serf.apache.org/abi/compat_report/serf/0.1.2/0.2.0/59321/abi_compat_report.html
* https://clusterlabs.org/pacemaker/abi/1.0.10_to_1.0.11/compat_report.html

According to the tool's documentation (https://lvc.github.io/abi-compliance-checker/):
> Without header files it is impossible to determine public symbols in ABI and data type definitions. Without shared libraries it is impossible to determine exported symbols in the ABI of the target library and also impossible to detect added/removed symbols. 

Did you try to break ABI intentionally and check if the script detects that?

t-b:
@mliszcz From the documentation

```
Usage with ABI Dumper
This new way is based on the analysis of the debug-info from binary objects. It's more reliable, faster and simple way.

The analyzed library should be compiled with "-g -Og" GCC options to contain DWARF debug info.

Create ABI dumps for both library versions first using the ABI Dumper tool:

abi-dumper OLD.so -o ABI-0.dump -lver 0

abi-dumper NEW.so -o ABI-1.dump -lver 1

And then compare ABI dumps to create report:

abi-compliance-checker -l NAME -old ABI-0.dump -new ABI-1.dump 
```

and that is exactly what I'm doing here.

I'll change it to only compile the tango target.

t-b:
@mliszcz I could create a ABI breakage locally. But you can also just try it out with CI and tango-9-lts.

mliszcz:
@t-b I just checked the chrono PR which breaks ABI (for instance, changes PollThCmd, changes size of Util class, etc ...): https://github.com/tango-controls/cppTango/pull/729
https://github.com/tango-controls/cppTango/suites/844720906/artifacts/9566101
And the generated report says that we have 100% compatibility. I think something is missing here. Maybe tool's documentation is just incomplete.

t-b:
@mliszcz Thanks for checking. I'll look into it.

t-b:
I'm closing this one for now. And reopening once we have something useful in tango-9-lts.

PR #732, opened 2020-06-26
Require C++14 standard

mliszcz:
As decided in last Tango Kernel meeting, we now require C++14 to compile cppTango.

https://github.com/tango-controls/tango-kernel-followup/blob/master/2020/2020-06-25/Minutes.md#recent-developments

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,8 @@ cmake_minimum_required(VERSION 3.7 FATAL_ERROR)
 
 project(cppTango)
 
-set(CXX_STANDARD_REQUIRED 11)
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 include(CTest)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,7 @@ cmake_minimum_required(VERSION 3.7 FATAL_ERROR)
 
 project(cppTango)
 
-# C++17 and higher support is currently not possible as omniorb uses
-# throw specifications and these are not supported anymore.
+# omniORB>=4.2.4 is required for compilation in C++17 (or better) mode.
 if (NOT DEFINED CMAKE_CXX_STANDARD)
     set(CMAKE_CXX_STANDARD 14)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ project(cppTango)
 # throw specifications and these are not supported anymore.
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
 
 include(CTest)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,9 @@ project(cppTango)
 
 # C++17 and higher support is currently not possible as omniorb uses
 # throw specifications and these are not supported anymore.
-set(CMAKE_CXX_STANDARD 14)
+if (NOT DEFINED CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,9 @@ cmake_minimum_required(VERSION 3.7 FATAL_ERROR)
 
 project(cppTango)
 
-set(CMAKE_CXX_STANDARD 11)
+# C++17 and higher support is currently not possible as omniorb uses
+# throw specifications and these are not supported anymore.
+set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 include(CTest)
--- a/configure/cmake_pch_linux.cmake
+++ b/configure/cmake_pch_linux.cmake
@@ -50,6 +50,7 @@ function(tango_add_pch pch_target ref_target pch_extra_cxx_flags)
         ${pch_compile_opts}
         ${pch_extra_cxx_flags}
         -x c++-header
+        -std=c++${CMAKE_CXX_STANDARD}
         -c
         -o ${tango_pch_file})
 
--- a/configure/CMakeLists.txt
+++ b/configure/CMakeLists.txt
@@ -193,14 +193,6 @@ if(${FAILED})
 endif()
 
 if(NOT WIN32)
-    include(CheckCXXCompilerFlag)
-    # C++17 and higher support is currently not possible as omniorb uses
-    # throw specifications and these are not supported anymore.
-    CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX14)
-    if(COMPILER_SUPPORTS_CXX14)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14")
-    endif()
-
     if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
         message("found GNU compiler ...")
         if(CMAKE_BUILD_TYPE STREQUAL "Release")


mliszcz:
@t-b and @bourtemb 
Ready for review, although I already see some conflicts with #725. We need to decide how to calculate appropriate `-std` flags passed to the compiler. For simplicity, I propose to not set any flags by hand but let CMake to figure them out by setting CMAKE_CXX_STANDARD to the minimum version we require (14). If user wants to compile with higher version, it would be his responsibility to ensure that his omniORB is new enough and to pass e.g. `-DCMAKE_CXX_STANDARD=20` during build.

t-b:
@mliszcz I think letting the user choose CMAKE_CXX_STANDARD is the way to go.

mliszcz:
@t-b then I propose to proceed that way:
1. I change this PR to set it conditionally
    ```cmake
    if (NOT DEFINED CMAKE_CXX_STANDARD)
        set(CMAKE_CXX_STANDARD 14)
    endif()
    ```
2. You remove setting of `-std` flag from #725. If you see any benefit in doing so, you may also bail out if CMAKE_CXX_STANDARD is not 14 and omniORB version is lower than 4.3.0. But I assume if user is tampering with cmake/compilation flags, he knows what she is doing.

configure/cmake_pch_linux.cmake
```diff
@@ -50,6 +50,7 @@ function(tango_add_pch pch_target ref_target pch_extra_cxx_flags)
         ${pch_compile_opts}
         ${pch_extra_cxx_flags}
         -x c++-header
+        -std=c++14
```
t-b: @mliszcz That should not be done as that contradicts that we allow the user to set the standard.

mliszcz: I couldn't come up with anything better than checking supported compiler flags, value of CXX_STANDARD, and hoping CMake chooses the same flag. We cannot ask CMake what flag it juses for given CXX_STANDARD, see linked issue in commit description. So what we have covers the 'default' case. If user sets different CXX_STANDARD and enables PCH at the same time, the PCH will not be used at all and the user will just get warning like (pch compiled with -std=c++14, source with -std=c++17):
```
main.cpp:1:17: warning: foo.h.gch: not used because `__cpp_static_assert' defined as ` 201411L' not ` 200410L' [-Winvalid-pch]
```
We already got similar problem reported in #734, therefore I opened #733 to remove our hand-made PCH implementation. But this requires upgrading CMake, thus needs further discussions.

The only way I see to fix this is to no use CXX_STANDARD and manually calculate `-std` flag, keep it up to date after each C++ release, handle corner-cases like c++1y, 2a, ..., - basically re-implement what CMake already does.

Maybe you have another idea what to do with those flags?

mliszcz: Maybe `-std=c++${CMAKE_CXX_STANDARD}` will be a bit more compatible?

bourtemb: > 
> 
> Maybe `-std=c++${CMAKE_CXX_STANDARD}` will be a bit more compatible?


That sounds like a nice compromise while waiting for the move to newer CMake version and the PCH support directly from CMake.

CMakeLists.txt
```diff
@@ -2,7 +2,13 @@ cmake_minimum_required(VERSION 3.7 FATAL_ERROR)
 
 project(cppTango)
 
-set(CXX_STANDARD_REQUIRED 11)
+# C++17 and higher support is currently not possible as omniorb uses
+# throw specifications and these are not supported anymore.
```
bourtemb: This comment is actually no longer true when using omniORB >= 4.2.4

t-b: @bourtemb I'm fixing that in #725 already.

configure/cmake_pch_linux.cmake
t-b: > Maybe -std=c++${CMAKE_CXX_STANDARD} will be a bit more compatible?

Yes, let's do that!

CMakeLists.txt
mliszcz: Maybe I'll just remove the comment in this case? Or at least change it to *You need omniORB>=4.2.4 for C++17*. Btw is it 4.2.4 or 4.3?

configure/cmake_pch_linux.cmake
mliszcz: Done, I modified the commit that introduced this change.

mliszcz: @bourtemb what do we do about #733 and CMake's built-in PCH support? If this is something we want to have, I can work on a PR and remove our custom PCH implementation.

CMakeLists.txt
t-b: @mliszcz Change it to `You need omniORB>=4.2.4 for C++17.` as both 4.2.2 and 4.3 work with `C++17`.

mliszcz: @t-b done! I also linked your conversation with omniORB maintainer in the commit message.

t-b: @mliszcz Nice!

t-b:
@mliszcz You could also remove `TANGO_CONSTEXPR` I introduced in c15e1c4e (cmake: Rework C++ compiler feature detection, 2020-01-20) as we now should not need it anymore.

Edit: Wait don't do that, our oldest one, see https://travis-ci.org/github/tango-controls/cppTango/jobs/706070146 does not support that as in `-- Check if the compiler supports C++14 relaxed constexpr: FALSE`.

mliszcz:
@t-b  I suggest to remove it anyway. The only function where we use `TANGO_CONSTEXPR` also compiles with `constexpr` on gcc 4.9.2. See: https://godbolt.org/z/R4cruo. I don't think relaxed constexpr is required in this case. Also, having such macro does not help with anything IMO (just encourages writing more macros / #ifdefs). If we have functions which are 'optionally constexpr' then we really cannot use them unconditionally in constexpr contexts. Consider this example:
```cpp
inline TANGO_CONSTEXPR int test(int a)
{
    if (a > 0) return 5;
    else return 3;
}

template <int X>
int foo()
{
    return X;
}

// foo<test(5)> does not compile without relaxed constexpr
<source>:55:15: error: call to non-constexpr function 'int test(int)'
     foo<test(5)>();
               ^
```
If you look at `constexpr` as a potential optimization that would allow compiler to evaluate some code at compilation time, then this is simply not true and compiler does not need the function to be marked as `constexpr` to evaluate it at compilation time during optimization phase. Just replace the constexpr with inline in the code on godbolt and you'll get the same assembly generated.

t-b:
@mliszcz Yes of course you can not rely on TANGO_CONSTEXPR meaning constexpr.

mliszcz:
So, if we cannot rely on TANGO_CONSTEXPR meaning constexpr, maybe we should just remove the this macro, as it redundant? I can do that if it's ok for you.

t-b:
@mliszcz It is not redundant. It allows to have functions constexpr with one compiler and a normal function with another one.

mliszcz:
But then any other code depending on that function should be wrapped in ifdef and we'd need to provide fallback code.

```cpp
TANGO_CONSTEXPR int bar(int);
template <int X> foo();

#ifdef TANGO_CXX_HAS_RELAXED_CONSTEXPR
  foo<bar(1)>()
#else 
  // call foo in some other way
#endif 
```

This raises the question why should we have two different solutions for the same thing, if the 'fallback' can cover both cases just fine. You still need to figure out how to do things without `bar(1)` being constexpr. And compiler can optimize the call out anyway.

Can you provide some example where this is actually useful and makes a difference?

t-b:
@mliszcz 
```
:~/devel/cppTango (master) $ git grep TANGO_CONSTEXPR
cppapi/server/tango_const.h.in:  #define TANGO_CONSTEXPR constexpr
cppapi/server/tango_const.h.in:  #define TANGO_CONSTEXPR
cppapi/server/tango_const.h.in:inline TANGO_CONSTEXPR const char * data_type_to_string(int type)
```

For these kind of helper functions, the current approach works. But feel free to remove  `TANGO_CONSTEXPR` as we can use `constexpr` instead as well. But please in a new PR.

mliszcz:
@t-b 
Thanks for your comment! I'm not pushing for removal of TANGO_CONSTEXPR. We may keep it of course. My point was to highlight two things: 1. constexpr does not improve anything in terms of runtime performance, 2. to actually benefit from this function being constexpr we need #ifdef, which is bad for maintainability and readability (recall all those HAS_MAP_AT, HAS_LAMBDA_FUNC we had at some point and for each of them we needed to provided backwards-compatible solution).

t-b (approved):
LGTM!

bourtemb (approved):
Looks good to me! Thanks!

PR #737, opened 2020-07-09
Restore event subscriptions after device server rename and restart

### LINKED ISSUE #679: Event reconnection issue after renaming a device server instance
Let's consider the `TangoTest/test` device server instance exporting `sys/tg_test/1` device.
If a client subscribes to `sys/tg_test/1/ampli` attribute events and if the device server instance `TangoTest/test` is restarted, the client will reconnect and will continue to receive events for `sys/tg_test/1/ampli` attribute.
BUT if the device server instance is renamed, for instance from `TangoTest/test` to `TangoTest/test_reconnection_pb`, the client will no longer receive events and will never reconnect.
This is due to the fact that the `ZmqEventSubscriptionChange` command is sent to the admin device of `sys/tg_test/1` when a reconnection should happen. 
`sys/tg_test/1` admin device name was `dserver/TangoTest/test` before the device server instance renaming and it becomes `dserver/TangoTest/test_reconnection_pb` after the device server instance renaming.
Tango is using a DeviceProxy pointing to the old admin device name when sending `ZmqEventSubscriptionChange` command, so the client never reconnects in this specific use case.

mliszcz:
If event channel is renamed (admin device name is changed), we now
detect this, create new Device Proxy to the admin device and update
internal structures accordingly.

Fixes #679.

### DIFF
--- a/cpp_test_suite/new_tests/cxx_server_event.cpp
+++ b/cpp_test_suite/new_tests/cxx_server_event.cpp
@@ -12,6 +12,27 @@
 #undef SUITE_NAME
 #define SUITE_NAME ServerEventTestSuite
 
+template <typename TEvent>
+struct EventCallback : public Tango::CallBack
+{
+    EventCallback()
+        : num_of_all_events(0)
+        , num_of_error_events(0)
+    {}
+
+    void push_event(TEvent* event)
+    {
+        num_of_all_events++;
+        if (event->err)
+        {
+            num_of_error_events++;
+        }
+    }
+
+    int num_of_all_events;
+    int num_of_error_events;
+};
+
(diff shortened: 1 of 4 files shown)


mliszcz:
Hi @t-b and @bourtemb 
The PR is ready for review. The correction is quite complex and touches multiple parts of the code because access to the event maps is done in many places in EventConsumer and EventKeepAliveThreads. I think it is possible to refactor this to have more localized updates of these maps (maybe we can wrap all three of them in a dedicated class), but I did nothing in that direction as I tried not to introduce any changes in ABI to allow for simple porting to 9.3.

t-b (approved):
LGTM.

PR #741, opened 2020-07-13
log4tango: Use BUILD_TESTING option

t-b:
We don't want to determine that by the platform. Accompanying fix to
a81c1584 (CMakeLists.txt: Rework test suite handling, 2020-01-23).


mliszcz (approved):
LGTM!

bourtemb (approved):

PR #742, opened 2020-07-13
Function name and line number in logs and exceptions

mliszcz:
This PR adds code location information to log messages and origin field in DevFailed exception.

Example log message:
```
1594734929 [139911612307200] DEBUG dserver/DevTest/test (/home/michal/Documents/cppTango/cppapi/server/subdev_diag.cpp:80) SubDevDiag::set_associated_device() entering ...
1594734929 [139911612307200] DEBUG dserver/DevTest/test (/home/michal/Documents/cppTango/cppapi/server/device_3.cpp:1881) Leaving Device_3Impl::write_attributes_34
```

Example exception (`Except::print_exception` output):
```
30: Tango NamedDevFailedList exception
30:    Exception for object Long64_attr_rw
30:    Index of object in call (starting at 0) = 0
30:        Severity = ERROR
30:        Error reason = DevTest_WriteAttrHardware
30:        Desc : DevFailed from write_attr_hardware
30:        Origin : virtual void DevTest::write_attr_hardware(std::vector<long int>&) at (/home/michal/Documents/cppTango/cpp_test_suite/cpp_test_ds/DevTest.cpp:1244)
```

The log message is updated automatically if you use macros like ERROR_STREAM or cout5.

To have exception origin field filled automatically, please use `TANGO_THROW_EXCEPTION(reason, desc)` macro (there is also TANGO_RETHROW_EXCEPTION and corresponding macros for _API_EXCEPTION).

See also: https://github.com/tango-controls/cppTango/pull/564#discussion_r450654487


mliszcz:
Large commit "Define constants for API exception reasons" (23bba98) was done mostly automatically:
```bash
rm cppapi/server/tango_const.h.in
for e in `rg '"API_.*?"' -o -I -N | sort | uniq | sed 's/.$//' | cut -c 2-`; do
  find . -name '*.cpp' -o -name '*.tpp' -o -name '*.h' | xargs -I{} sed -i 's|"'$e'"|'$e'|g' {}
done
git checkout cppapi/server/tango_const.h.in
rg 'API_[A-Z][a-z][a-zA-Z]+' -o -I -N | sort | uniq | xargs -I{} echo 'const char* const {} = "{}";'
# replace definitions in tango_const.h.in with output of the above command
# fix broken compilation where needed
```

mliszcz:
Large commit "Replace Except::throw_exception with macro" (63eaac7) was done mostly automatically with following scripts:

```bash
for f in `rg -l -e '::throw_exception' -e '::re_throw_exception'`; do
  echo "patching $f"
  python ../rename.py $f > $f.tmp
  mv $f.tmp $f
done

# to remove extra newlines (forgot about that in rename script
for f in `rg TANGO_THROW_ -l`; do sed -i -e '/TANGO_THROW_/{n;d}' $f; done
# same for RETHROW
```

Changes to these were reverted:
```
cppapi/client/helpers/TangoExceptionsHelper.h
cppapi/server/except.cpp
cppapi/server/except.h
cppapi/client/apiexcept.h
cpp_test_suite/cpp_test_ds/SigThrow.cpp
cppapi/client/helpers/PogoHelper.h
```

Changes in these required manual corrections:
```
cppapi/client/devapi.h
patching cppapi/server/attrprop.h
cppapi/client/helpers/DeviceProxyHelper.h
cppapi/server/device.cpp
cppapi/server/dserver.cpp
cppapi/server/pollext.h
cppapi/server/w_attrsetval.tpp
cppapi/server/attribute.h
cppapi/server/pipe.h
cppapi/client/devapi_base.cpp
cppapi/client/dbapi_base.cpp
cppapi/server/attribute.cpp
```

Rename script:
```python
import re
import sys

SEARCH_THROW = '::throw_exception'
SEARCH_RETHROW = '::re_throw_exception'

FLUSH = object()

def Parser():
    buff = ""
    while True:
        c = yield
        if (c is FLUSH):
            print(buff, end='')
            buff = ""
            continue

        buff += c

        if c.isspace():
            print(buff, end='')
            buff = ""
            continue

        if buff.endswith(SEARCH_THROW) or buff.endswith(SEARCH_RETHROW):

            api_obj, _method = buff.rsplit("::", maxsplit=1)
            api = not bool(re.match(r"^(::)?(Tango)?(::)?Except", api_obj))
            rethrow = buff.endswith(SEARCH_RETHROW)
            buff = ""

            arg_buff = ""
            args = []
            yield # opening brace
            while True:
                c = yield
                if c == ";":
                    args.append(arg_buff[:-1]) # closing brace
                    break
                if c == ",":
                    args.append(arg_buff)
                    arg_buff = ""
                    continue
                if c == '"':
                    arg_buff += c
                    while True:
                        c = yield
                        arg_buff += c
                        if c == '"':
                            break
                    continue
                arg_buff += c

            args = [s.replace("(const char *)", "") for s in args]
            args = [s.replace("(const char*)", "") for s in args]
            args = [s.strip() for s in args]

            if rethrow:
                assert len(args) == 4
                if api:
                    print("TANGO_RETHROW_API_EXCEPTION({}, {}, {}, {});".format(
                        api_obj, args[0], args[1], args[2]))
                else:
                    print("TANGO_RETHROW_EXCEPTION({}, {}, {});".format(
                        args[0], args[1], args[2]))
            else:
                assert len(args) == 3
                if api:
                    print("TANGO_THROW_API_EXCEPTION({}, {}, {});".format(
                        api_obj, args[0], args[1]))
                else:
                    print("TANGO_THROW_EXCEPTION({}, {});".format(
                        args[0], args[1]))

parser = Parser()
parser.send(None)

with open(sys.argv[1]) as file:
    for line in file:
        for char in line:
            parser.send(char)
    parser.send(FLUSH)
```

mliszcz:
Hi @t-b and @bourtemb, the build is passed and PR is ready for review. There are two lengthy commits generated with a script (attached above for reference). I suggest you to just pick some random files and check that, because the whole diff is huge.
* "Define constants for API exception reasons" (23bba98)
* "Replace Except::throw_exception with macro" (63eaac7)

t-b (changes requested):
Very nice addition.

Once that is merged, we can close https://github.com/tango-controls/cppTango/pull/366 IMHO.

Two small changes would be nice:

f3fa387d (Add macro for detecting current function name, 2020-07-11)

Can you add a `#undef BOOST_CURRENT_FUNCTION` so that we don't have that name floating around?

b79d92d7 (Include current file and line number in log messages, 2020-07-11)

Can you add

```
#undef TANGO_QUOTE
#undef TANGO_QUOTE_
```

to the bottom of tango_current_function.h.

mliszcz:
> 
> Can you add a `#undef BOOST_CURRENT_FUNCTION` so that we don't have that name floating around?
> 
> Can you add
> 
> ```
> #undef TANGO_QUOTE
> #undef TANGO_QUOTE_
> ```
> 
> to the bottom of tango_current_function.h.

It's not that simple. Seems like macros are expanded lazily. See: https://gcc.gnu.org/onlinedocs/cpp/Argument-Prescan.html#Argument-Prescan
> After substitution, the entire macro body, including the substituted arguments, is scanned again for macros to be expanded.

```cpp
#define A 1
#define B A
#undef A

int foo() { return B; }
```

```
<source>: In function 'int foo()':
<source>:2:11: error: 'A' was not declared in this scope
    2 | #define B A
      |           ^
<source>:5:20: note: in expansion of macro 'B'
    5 | int foo() { return B; }
      |                    ^
Compiler returned: 1
```

But I think we should remove `BOOST_CURRENT_FUNCTION` because there may be collision if user includes proper boost header. I'll just replace `BOOST_CURRENT_FUNCTION` with `TANGO_CURRENT_FUNCTION` if that's ok with you.

`TANGO_QUOTE` must stay for the same reason.

t-b:
@mliszcz Thanks, learned something today. 

> I'll just replace BOOST_CURRENT_FUNCTION with TANGO_CURRENT_FUNCTION if that's ok with you.

Jeep, that is good!

t-b (approved):

sergirubio:
Hi all,
If you agree, I can do the review for this PR
is it ok?
Sergi

mliszcz:
Thanks @sergirubio ! I just added you as a reviewer of this PR.

sergirubio:
Hi Michal,

I already build your branch and I am going to test it using a modified TangoTest device compiled against it with -v4 flag on it.

I have just one question ... to see the information in the DevFailed exception from a client, the client should be also using the same branch or it could be a tango 9.3.4 python client?

Sergi

mliszcz:
Hi Sergi,
Thanks for checking this PR.

If the exception comes from the server, then you don't need the code in this branch on the client side. Location information is just appended to the origin field which is transferred from the server.

For exceptions throw from the client code, you'll not have code location information if you use 9.3.4.

Issue #735, opened 2020-06-28
Public/private header separation and stable ABI proposal


mliszcz:
I recently started looking into what can be done not to expose all system internals in tango.h.

The goals are to:
* define stable API and ABI for Tango which is not impacted by internal refactoring and bugfixes
* have the same header layout in source tree as when installed on production system (to prevent issues like in #720)
* improve compilation times
  * currently any change in most of the header files requires recompilation of all translation units because everything is transitively included from tango.h
  * there should be no need for hacks like PCHs if #includes were used correctly
* not expose libraries used internally by Tango
  * this will allow to include boost (or whatever) without forcing users to install it
  * after breaking API a bit, it should be possible to hide omniORB and ZMQ as well (allows to have the same API for Tango v10)

It looks like a lot of work but it is really mostly renaming and moving stuff around, which can be automated to some extent with refactoring tools and IDEs. I also do not expect to introduce much conflicts with pending PRs because we do not rename files and we change mostly function declaration lines (like `Attribute::do_something` to `AttributePrivate::do_something`).

The basic idea is to change the classes we modify frequently to use PImpl idiom (a bit modified to support inheritance). Here is a blog-post describing how QT (much larger than Tango) is doing this: https://wiki.qt.io/D-Pointer .

My proposal is to have something similar in Tango. We may proceed as follows:
1. PImpl all relevant classes, the procedure is like:
   * Rename Attribute to AttributePrivate
   * Create Attribute class in `<repo>/include/tango/attribute.h`
     * stores owning pointer to AttributePrivate
	 * has only methods which are currently defined in @publicsection of Attribute
	 * forwards all calls to AttributePrivate
	 * has no protected data members (users do not inherit from this class anyway)
     * has additional protected constructor taking AttributePrivate (to support WAttribute inheritance)
   * Rename WAttribute to WAttributePrivate (WAttributePrivate inherits from AttributePrivate)
   * Create WAttribute inheriting from Attribute
     * in constructor, creates WAttributePrivate and passes it to Attribute using special constructor
   * Change MultiAttribute to store Attribute instead of AttributePrivate
   * Pass Attribute/WAtttribute again to device's read/write callbacks
   * Proceed in similar way with DeviceImpl, Device_5Impl, ...
     * Except that DeviceImpl does not inherit from any class (even CORBA servant), but wraps DeviceImplPrivate which inherits from the servant
[...]

PR #744, opened 2020-07-25
PImpl Attribute classes

mliszcz:
This PR changes Attribute and derived classes to use PImpl idiom.
We have two class hierarchies now:
* `AttributePrivate <- WAttributePrivate <- FwdAttributePrivate`, which are just renamed Attribute, WAttribute and FwdAttribute (as we knew them before this PR),
* `Attribute <- WAttribute <- FwdAttribute`, which contain only public methods from the corresponding Private classes. Additionally Attribute contains owning pointer to AttributePrivate.

In cppTango library code we try to use normal (public) classes when possible, but in places where we need to break encapsulation, we can call Attribute::get_impl() to access private implementation.

The new classes (and their dependencies) are declared in headers located in `cppapi/include`. In long term only this directory should be visible for the users (and copied to their machines as part of install target).

See also #735.


mliszcz:
Hi @t-b and @bourtemb, the PR is ready for review now. I recommend commit-wise approach as there are few large commits that either do mass-rename or copy some files.

I'd also like to get your opinion on few things I am not completely sure about:
1. Should we have a copyright note in the new files I added? If yes, maybe we could define a template somewhere in contributing guidelines. What we have now seems to be outdated and inconsistent between different files. For now I have not added any copyright notes to the new files (also applies to some other PRs I recently submitted), as all this information can be found in other places (git log, LICENSE file, etc...). I cannot comment from legal point though.
2. We have two `attribute.h` files, one in `cppapi/server` containing `AttributePrivate` and other in `cppapi/include/server` containing `Attribute` definition. This is ok as long as we are careful enough about what we pass to the compiler include path with `-I`. I'd rename the one with `AttributePrivate` to `attribute_private.h` (to resolve name collision and to follow the rule that file is named after main class it contains), but I'm afraid that it will break any pending PRs which touch Attribute code. What do you think?
3. TangoTest device is using private (undocumented) methods `WAttribute::get_w_dim_x/y()`. For now I just exposed them in the public class without documenting them in API documentation. Maybe we should include them in the public API by documenting them? Maybe we should also review other undocumented methods and decide if we want to expose them or not, because when we release this PR and start compiling device servers in the wild, for sure more issues like this one will pop up.

PR #745, opened 2020-07-26
.travis/gcc-latest/Dockerfile: Update to 10.2.0

t-b:


mliszcz (approved):
LGTM! Maybe we should just use `10` tag to avoid changing it every minor/patch upgrade (which should not break our codebase).

t-b:
@mliszcz It's a 1 minute job for me to change that. I also prefer to have fixed versions instead of moving targets.

bourtemb (approved):

PR #746, opened 2020-08-06
Event unsubscription from DeviceProxy destructor

### LINKED ISSUE #353: Exception on unsubscription when a client has two proxies for the same attribute
The example code below creates two attribute proxies to the same attribute, subscribes both of them to the same event, then unsubscribes the first one and destroys its proxy. This apparently unsubscribes the other callback and a Tango exception is thrown during `pr2->unsubscribe_event(ev2);`. No exception is thrown when this line is moved before the line with `delete pr1;`.

Happens with Tango 9.2.5a.

```
#include <tango.h>

class MyCallBack : public Tango::CallBack
{
   protected:
   void push_event(Tango::AttrConfEventData *) final
   {
   }
};

int main(void)
{
   MyCallBack clb1, clb2;
   int ev1, ev2;
   Tango::AttributeProxy *pr1 = new Tango::AttributeProxy("sys/tg_test/1/long_scalar");
   Tango::AttributeProxy *pr2 = new Tango::AttributeProxy("sys/tg_test/1/long_scalar");

   try
   {
      ev1 = pr1->subscribe_event(Tango::ATTR_CONF_EVENT, &clb1);
      ev2 = pr2->subscribe_event(Tango::ATTR_CONF_EVENT, &clb2);
      pr1->unsubscribe_event(ev1);
      delete pr1;
      pr2->unsubscribe_event(ev2);
      delete pr2;
   }
   catch(const Tango::DevFailed &ex)
   {
      Tango::Except::print_exception(ex);
   }
   return 0;
}
```

mliszcz:
The pointer to device proxy used for event subscription is now stored in
EventSubscribeStruct (which is unique entry per subscription) instead of
EventCallBackStruct (which is shared between all subscriptions to the
same event).

Ability to associate subscriptions with device proxies is needed to
correctly unsubscribe from events inside device proxy destructor.

Fixes #353

### DIFF
--- a/cpp_test_suite/new_tests/cxx_dserver_misc.cpp
+++ b/cpp_test_suite/new_tests/cxx_dserver_misc.cpp
@@ -394,6 +394,38 @@ cout << "str = " << str << endl;
         TS_ASSERT_EQUALS(3, callback2.num_of_all_events);
         TS_ASSERT_EQUALS(0, callback2.num_of_error_events);
     }
+
+    void test_unsubscription_multiple_subscriptions_with_single_proxy()
+    {
+        EventCallback<Tango::EventData> callback1{};
+        EventCallback<Tango::EventData> callback2{};
+
+        const std::string attribute_name = "event_change_tst";
+
+        auto proxy = std::make_unique<DeviceProxy>(device1_name);
+
+        TS_ASSERT_THROWS_NOTHING(proxy->subscribe_event(
+            attribute_name, Tango::USER_EVENT, &callback1));
+        TS_ASSERT_THROWS_NOTHING(proxy->subscribe_event(
+            attribute_name, Tango::USER_EVENT, &callback2));
+
+        TS_ASSERT_THROWS_NOTHING(device1->command_inout("IOPushEvent"));
+        Tango_sleep(1);
+        TS_ASSERT_EQUALS(2, callback1.num_of_all_events);
+        TS_ASSERT_EQUALS(0, callback1.num_of_error_events);
+        TS_ASSERT_EQUALS(2, callback2.num_of_all_events);
+        TS_ASSERT_EQUALS(0, callback2.num_of_error_events);
+
+        proxy.reset(nullptr);
+
+        TS_ASSERT_THROWS_NOTHING(device1->command_inout("IOPushEvent"));
+        Tango_sleep(1);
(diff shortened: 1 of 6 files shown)


t-b (dismissed):
LGTM. It took me a while to understand that `callback_list[0]` is always valid. But in https://github.com/tango-controls/cppTango/blob/d067b589985c701ae49541f3a27366ceaab0f196/cppapi/client/event.cpp#L1745 we always add one entry.

mliszcz:
Thanks @t-b for this comment. I agree it is a bit fishy. Maybe I'll add assertion to all the places to clearly indicate that the presence of at least one entry is guaranteed? What do you think?
```cpp
assert(!epos->second.callback_list.empty());
// access callback_list[0]
```

t-b:
@mliszcz An assertion yes, but maybe only once? Can you add a wrapper function which returns `callback_list[0]` and then only that wrapper has the assertion? But I'm also okay with as it is.

t-b:
Automatic unsubscription in DeviceProxy dtor was added in 36281159 (- Add automatic event un-subscription in DeviceProxy class dtor, 2016-06-10).

mliszcz:
HI @t-b can you re-check the last commit (the one changing access to `callback_list[0]`)? It dismissed your previous review.

t-b (dismissed):

t-b (dismissed):
LGTM

bourtemb (commented):
Thanks @mliszcz for the work on this PR.

It looks like there is one case which is not well handled here.
When a user subscribes to the same event (same attribute and same event type) several times using the same DeviceProxy but using different callbacks.

If the DeviceProxy object is deleted, it should unsubscribe all the subscriptions done with this DeviceProxy object.
With the current solution in this PR, during DeviceProxy deletion, it unsubscribes only for the 1st callback in the use case I described it seems and some callbacks are still called when they should not after DeviceProxy deletion.

mliszcz:
Thanks @bourtemb for catching this.There was indeed one place where I forgot to initialize the new DeviceProxy* field in event subscription struct. It is fixed now and test is added in the latest commit.

t-b (approved):

bourtemb (approved):
Thanks @mliszcz !
It seems to work as well for the use case I was describing now.
Thanks for the new test for that too!
Looks good to me!

PR #752, opened 2020-08-13
Remove conditional constexpr define

t-b:
We now allow a couple of C++14 features so we can remove the conditional
handling for C++14 style constexpr. As pointed out by Michal Liszcz
conditional constexpr also does not allow us internally to rely on
functions being constexpr.

@mliszcz IIRC conclusion was to make this change or?

### DIFF
--- a/configure/CMakeLists.txt
+++ b/configure/CMakeLists.txt
@@ -214,14 +214,6 @@ if(WARNINGS_AS_ERRORS)
   endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


mliszcz (approved):

bourtemb (approved):

mliszcz:
> 
> @mliszcz IIRC conclusion was to make this change or?

This is something I proposed during a review of another PR. But since Reynald also already approved, I think we can conclude that this is something what we want.

PR #755, opened 2020-08-27
Bugfix/fix abi checking

t-b:
So here is my stab at fixing the ABI checking.

This PR includes a commit which breaks the ABI, just to see that it is working better now. Judging from the open issues on the involved projects I doubt that this is foolproof, but at least I could get it working now. If we decide that this approach is too kludgy we can also just remove it again.

I'm getting the following now when I include the breaking commit

```diff
--- old.dump    2020-08-27 21:25:02.180190259 +0200
+++ new.dump    2020-08-27 21:25:02.296190257 +0200
@@ -6743,7 +6743,7 @@
                                             'Class' => '1040135',
                                             'Header' => undef,
                                             'Line' => '115',
-                                            'MnglName' => '_ZN5Tango7ApiUtil18get_asynch_repliesEl',
+                                            'MnglName' => '_ZN5Tango7ApiUtil18get_asynch_repliesEi',
                                             'Param' => {
                                                          '0' => {
                                                                   'name' => 'this',
@@ -6751,7 +6751,7 @@
                                                                 },
                                                          '1' => {
                                                                   'name' => 'call_timeout',
-                                                                  'type' => '18757'
+                                                                  'type' => '18744'
                                                                 }
                                                        },
                                             'Return' => '1',
@@ -237394,7 +237394,7 @@
                                                   '_ZN5Tango7ApiUtil14device_to_attrERKNS_15DeviceAttributeERNS_16AttributeValue_4E' => 1,                                                                                                                                      
                                                   '_ZN5Tango7ApiUtil14get_ip_from_ifERSt6vectorINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESaIS7_EE' => 1,                                                                                                             
                                                   '_ZN5Tango7ApiUtil15set_sig_handlerEv' => 1,
-                                                  '_ZN5Tango7ApiUtil18get_asynch_repliesEl' => 1,
+                                                  '_ZN5Tango7ApiUtil18get_asynch_repliesEi' => 1,
                                                   '_ZN5Tango7ApiUtil18get_asynch_repliesEv' => 1,
                                                   '_ZN5Tango7ApiUtil19attr_to_device_baseINS_16AttributeValue_4EEEvPKT_PNS_15DeviceAttributeE' => 1,                                                                                                                            
                                                   '_ZN5Tango7ApiUtil19attr_to_device_baseINS_16AttributeValue_5EEEvPKT_PNS_15DeviceAttributeE' => 1,                                                                                                                            
ABI/API breakages detected!
```

and no complaint when I revert the ABI breaking commit.

This approach does give quite some false positives, but I don't see how I can get the original approach working.

### DIFF
--- a/.travis/check-ABI-API-compliance.sh
+++ b/.travis/check-ABI-API-compliance.sh
@@ -23,7 +23,7 @@ function generate_info() {
   local prefix=$1
   local revision=$2
 
-  git worktree remove $prefix-branch || true
+  git worktree remove --force $prefix-branch || true
   git -c advice.detachedHead=false worktree add ${prefix}-branch ${revision}
   mkdir ${prefix}-branch/build
   cd ${prefix}-branch/build
--- a/.travis/check-ABI-API-compliance.sh
+++ b/.travis/check-ABI-API-compliance.sh
@@ -12,11 +12,8 @@ then
   export CMAKE_BUILD_PARALLEL_LEVEL=$(grep -c ^processor /proc/cpuinfo)
 fi
 
-function print_abi_api_breakages() {
-   echo "ABI breakages detected:"
-   cat compat_reports/libtango/${old_revision}_to_${new_revision}/abi_affected.txt | c++filt
-   echo "API breakages detected:"
-   cat compat_reports/libtango/${old_revision}_to_${new_revision}/src_affected.txt | c++filt
+function exit_on_abi_api_breakages() {
+   echo "ABI/API breakages detected!"
    exit 1
 }
 
@@ -54,6 +51,6 @@ generate_info "old" ${old_revision}
 generate_info "new" ${new_revision}
 
 # Compare results
-abi-compliance-checker -l libtango -old ${base}/libtango-old-${old_revision}.dump \
-                                   -new ${base}/libtango-new-${new_revision}.dump \
-                                   -list-affected || print_abi_api_breakages
+sed "s/${old_revision}/fixed/g" ${base}/libtango-old-${old_revision}.dump > old.dump
+sed "s/${new_revision}/fixed/g" ${base}/libtango-new-${new_revision}.dump > new.dump
+diff -Nur old.dump new.dump || exit_on_abi_api_breakages
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -48,8 +48,3 @@ jobs:
       run: .travis/install_tango_idl.sh
     - name: Check ABI/API compliance
       run: docker exec -w /home/tango/src cpp_tango .travis/check-ABI-API-compliance.sh || true
-    - name: Upload compliance report
-      uses: actions/upload-artifact@v1
-      with:
-        name: ABI-API compliance report
-        path: compat_reports


t-b:
https://github.com/tango-controls/cppTango/runs/1038310880#step:7:551 shows the breakage.

t-b:
@mliszcz @bourtemb So here is the fixed version now.

mliszcz:
@t-b have you tried using `abi-dumper -compare OLD.dump NEW.dump` instead of plain `diff`? Maybe it will be a bit more robust (if if works). https://manpages.debian.org/unstable/abi-dumper/abi-dumper.1.en.html

Edit: Actually these dumps look like serialized perl data structures and the dumper tool seems to treat them as such: https://github.com/lvc/abi-dumper/blob/master/abi-dumper.pl#L6578-L6634
So if `abi-dumper -compare` won't work we may try to write a perl one-liner that would load both dumps, pass them into https://metacpan.org/pod/Data::Diff and print the result?

bourtemb:
> @t-b have you tried using `abi-dumper -compare OLD.dump NEW.dump` instead of plain `diff`? Maybe it will be a bit more robust (if if works). https://manpages.debian.org/unstable/abi-dumper/abi-dumper.1.en.html

@t-b I think you didn't reply to this question from @mliszcz

t-b:
@bourtemb THanks for the reminder. I will have a look, altough I think that we could also just use what we have and see if it works in practice.

.travis/check-ABI-API-compliance.sh
```diff
@@ -54,6 +51,6 @@ generate_info "old" ${old_revision}
 generate_info "new" ${new_revision}
 
 # Compare results
-abi-compliance-checker -l libtango -old ${base}/libtango-old-${old_revision}.dump \
-                                   -new ${base}/libtango-new-${new_revision}.dump \
-                                   -list-affected || print_abi_api_breakages
+sed "s/${old_revision}/fixed/g" ${base}/libtango-old-${old_revision}.dump > old.dump
+sed "s/${new_revision}/fixed/g" ${base}/libtango-new-${new_revision}.dump > new.dump
+diff -Nur old.dump new.dump || exit_on_abi_api_breakages
```
bourtemb, line 56: diff -r option (to recursively compare subdirectories) does not hurt but is not needed here.

t-b:
@mliszcz @bourtemb 
My earlier version used `abi-dumper` but as I said in the commit message, this had way too many bugs (false positives). How about we merge that now, given it does work, and later on improve it if required.

mliszcz (approved):
Hi @t-b, fine with me! This is still better than what we already have.

.travis/check-ABI-API-compliance.sh
bourtemb, line 56:
```suggestion
diff -Nu old.dump new.dump || exit_on_abi_api_breakages
```

bourtemb (approved):
@t-b thanks for this PR.
We'll see in practice how it behaves.
I just added a small suggestion to remove a useless (I think, unless this option is useful on some operating system versions?) option in the diff but this option should not cause any harm in any case so it can be merged like that too.

t-b:
@bourtemb Thanks for the review. Yes the 'r' to diff is not needed, I would still merge as is, as it does not hurt and we can finally get this merged.

Issue #756, opened 2020-08-27
Fix some sonarcloud complaints


t-b:
The lists at [1] and [2] show the top issues sonarcloud found. Most of these look really like we should fix them.

[1]: https://sonarcloud.io/project/issues?id=org.tango-controls%3Acpp-tango%3Atango-9-lts&resolved=false&severities=BLOCKER
[2]: https://sonarcloud.io/project/issues?id=org.tango-controls%3Acpp-tango%3Atango-9-lts&resolved=false&severities=CRITICAL&types=VULNERABILITY

t-b:
The merge of #757 let us jump from D to B in security, see https://sonarcloud.io/project/activity?custom_metrics=security_rating&graph=custom&id=org.tango-controls%3Acpp-tango%3Atango-9-lts.

t-b:
And if I understand sonarcloud correctly if we want to go from B to A in security we have to remove strcat and strcpy invocations.

I'm all for that, but these are a couple of more hits.

```
$git grep -P "(strcpy|strcat)" cppapi cpp_test_suite | wc -l
187
```

PR #757, opened 2020-08-28
Replace sprintf with snprintf

t-b:
See #756.

Most of the replacements were done mechanically with 
```
for i in `git grep -l sprintf`; do sed -i "s/:*sprintf(\([a-zA-Z_]*\),/std::snprintf(\1, sizeof(\1),/g" $i; done
```
 and then manually checked. Blackbox was done manually.


bourtemb (approved):
LGTM. Thanks @t-b

PR #759, opened 2020-08-28
Cleanup log4tango

### LINKED ISSUE #712: Remove log4tango/threading utilities
I propose to remove all the code we use to support threads in log4tango on various platform and replace it with corresponding classes from standard library:
* [ ] Remove all code from (+ corresponding implementation files) https://github.com/tango-controls/cppTango/tree/59931b9fbc92e375ae2e6c99f1adf779b97df975/log4tango/include/log4tango/threading
* [ ] Replace with standard classes
  * Mutex -> std::mutex
  * ScopedLock -> std::lock_guard
  * RecursiveMutex -> std::reentrant_mutex
  * ThreadLocalDataHolder -> thread_local storage class
* [ ] Remove flags
  * HAVE_PTHREAD_ATTR_GET_NP 1
  * HAVE_PTHREAD_GETATTR_NP 1
  * HAVE_PTHREAD_CREATE 1
  * HAVE_PTHREAD_NP_H 1
  * LOG4TANGO_USE_PTHREADS 1
  * LOG4TANGO_USE_MSTHREADS
  * LOG4TANGO_HAVE_THREADING

### LINKED ISSUE #713: Remove custom ostringstream implementation
What we currently have is invalid C++ code and shall be removed.
http://eel.is/c++draft/namespace.std
> Unless otherwise specified, the behavior of a C++ program is undefined if it adds declarations or definitions to namespace std or to a namespace within namespace std.

These files (and LOG4TANGO_HAVE_SSTREAM flag) should be removed:
* log4tango/include/log4tango/Portability.hh
* log4tango/src/PortabilityImpl.hh
* log4tango/src/PortabilityImpl.cpp

### LINKED ISSUE #728: Cleanup log4tango build-system
- [x] Remove autotools/automake
- [x] Remove VS projects

```
drwxr-xr-x    - michal michal 29 paź  2019 -- ms32vc10/
drwxr-xr-x    - michal michal 29 paź  2019 -- ms64vc9/
drwxr-xr-x    - michal michal 29 paź  2019 -- ms64vc10/
drwxr-xr-x    - michal michal 29 paź  2019 -- ms64vc11/
drwxr-xr-x    - michal michal 29 paź  2019 -- ms64vc12/
drwxr-xr-x    - michal michal 29 paź  2019 -- msvc8/
drwxr-xr-x    - michal michal 29 paź  2019 -- msvc9/
```

### LINKED ISSUE #765: log4tango: Fix the check for int64_t
We are currently doing
https://github.com/tango-controls/cppTango/blob/40c05aef2fe4988511ccedfe949889f701b1618f/log4tango/config/config.cmake#L57-L67

 but the presence of int64_t does have nothing to do with the platform or the architecture.

See https://stackoverflow.com/a/48460091

t-b:
Close #728. 
Close #713.
Close #712.


log4tango/include/log4tango/threading/StandardThreads.hh
```diff
@@ -1,37 +1,39 @@
...
+
+namespace threading {
 
-#include <log4tango/Portability.hh>
+std::string get_thread_id();
 
-    
+using Mutex = std::mutex;
+using RecursiveMutex = std::recursive_mutex;
+using ScopedLock = std::lock_guard<std::mutex>;
```
mliszcz: I suggest to remove these definitions and use 'std' classes directly. Or we may adapt a rule that we always use these aliases instead of 'std' classes. This is to avoid possible confusion due to mixing different names.

* Mutex is referenced only once, from: AppenderAttachable.hh
* RecursiveMutex is never used
* ScopedLock is used in a few places in AppenderAttachable.cpp and Logger.cpp.

log4tango/config/config.cmake
```diff
@@ -27,32 +27,10 @@ endmacro()
...
-
-#check threads
-check_library_exists(pthread pthread_attr_get_np "" HAVE_PTHREAD_ATTR_GET_NP)
-check_library_exists(pthread pthread_getattr_np "" HAVE_PTHREAD_GETATTR_NP)
-check_library_exists(pthread pthread_create "" HAVE_PTHREAD_CREATE)
-check_include_file("pthread.h;pthread_np.h" HAVE_PTHREAD_NP_H)
-
 if(CMAKE_THREAD_LIBS_INIT)
     set(LOG4TANGO_HAVE_THREADING "/**/")
 endif()
```
mliszcz: https://cmake.org/cmake/help/latest/module/FindThreads.html
> The thread library to use. This may be empty if the thread functions are provided by the system libraries and no special flags are needed to use them.

This check is probably bad for portability. Maybe we should check for `Threads_FOUND` instead?

Also, is it possible to use Tango without threads? After a quick check of omniORB source code it seems that they require threads (posix or ms or vxworks) and do not have any fallback implementation (compilation errors out with "Implementation header file incomplete"). And since log4tango is not standalone library, I suggest to do the following:
* require Threads_FOUND to be set here
* remove LOG4TANGO_HAVE_THREADING, remove DummyThread.hh/.cpp
* move log4tango/threading/StandardThreads.hh to log4tango/StandardThreads.hh if you decide to keep that file (I'd also rename to e.g. 'ThreadingUtils.hh)
* remove entire log4tango/threading directory

mliszcz (changes requested):
LGTM! Very good initiative and I think we can use this opportunity to remove even more stuff.

log4tango/include/log4tango/threading/StandardThreads.hh
t-b: Good point. Done.

log4tango/config/config.cmake
t-b: Fair point. I've done that. And also reworked the `int64_t` check.

mliszcz (approved):
LGTM! FYI in #765 I added a comment related to int64_t check because you are now resolving this one as well.

log4tango/config/config.cmake
t-b: @mliszcz I've created https://github.com/tango-controls/cppTango/issues/793 I think we should tackle that one in a separate PR. I don't see an issue with backwards compatibility.

Issue #740, opened 2020-07-13
Problems when cross compiling for ARM


JeanLucPons:
Hello,

I've got several problem when trying to cross compile Tango for ARM platform:
gcc version 4.7.3 (Sourcery CodeBench Lite 2013.05-40) 

My cmake build:
```
[debian9-devel.pons] > cmake .. -DCMAKE_CXX_COMPILER:FILEPATH=/segfs/linux/xilinx_sdk/arm-toolchain/bin/arm-xilinx-linux-gnueabi-gcc -DIDL_BASE=/users/pons/tango/tango-idl/install -DOMNI_BASE=/segfs/tango/ORB/omniORB4.2.1_p1/arm_xilinx -DZMQ_BASE=/segfs/tango/transport/zmq4.0.5/arm_xilinx -DIDL_BASE=/users/pons/tango/tango-idl -DTANGO_JPEG_MMX=OFF -DBUILD_TESTING=OFF -DCMAKE_INSTALL_PREFIX=/segfs/tango/release/arm_xilinx/9.3.4-rc6
System TANGO_HOST=acudebian7:10000
No build type specified - default is Release
Using IDL_BASEIDL_BASE=/users/pons/tango/tango-idl
Using OMNI_BASE=/segfs/tango/ORB/omniORB4.2.1_p1/arm_xilinx
Using ZMQ_BASE=/segfs/tango/transport/zmq4.0.5/arm_xilinx
-- Looking for C++ include zmq.hpp
-- Looking for C++ include zmq.hpp - found
-- Check if zmq::socket has a disconnect method: FALSE
Verifying /segfs/tango/ORB/omniORB4.2.1_p1/arm_xilinx/bin/omniidl
omniidl version 1.0
found GNU compiler ...
setup for release build ...
Remove tango.h, tangoSK.cpp and tangoDynSK.cpp from idl if exist
Generate tango.h, tangoSK.cpp and tangoDynSK.cpp from idl
Using OMNIIDL_PATH=/segfs/tango/ORB/omniORB4.2.1_p1/arm_xilinx/bin/
Using IDL=/users/pons/tango/tango-idl/include
-- OMNIIDL_PATH: /segfs/tango/ORB/omniORB4.2.1_p1/arm_xilinx/bin/
-- CMAKE_CURRENT_BINARY_DIR: /users/pons/tango/cppTango/build/cppapi/server/idl
[...]

PR #763, opened 2020-09-10
Allow turning shared library building off

t-b:


mliszcz (approved):
LGTM!

t-b:
@mliszcz Thanks for the review. I had to unforunately fix the PR as the tests were correctly not passing.

bourtemb:
@t-b, as you mentioned yesterday in the Kernel meeting, after the merge of this PR, only the shared library or the static library will be built, no longer both together.
So, in the past, we used to build both and then both were installed together.
If we want to continue to provide/install both static and shared cppTango libraries with the Tango Source Distribution, we will have to adapt the TangoSourceDistribution to do 2 cppTango builds and 2 installations steps, then?

t-b:
> If we want to continue to provide/install both static and shared cppTango libraries with the Tango Source Distribution, we will have to adapt the TangoSourceDistribution to do 2 cppTango builds and 2 installations steps, then?

I don't think you really need two different builds, but yes something like that needs to be done.

If that is not what we want, `BUILD_SHARED_LIBS` could also only enable/disable shared library building and we always keep building the static library.

lorenzopivetta:
> If that is not what we want, `BUILD_SHARED_LIBS` could also only enable/disable shared library building and we always keep building the static library.

Hi @t-b may I suggest the opposite? We always build the shared lib and only build the static library if BUILD_STATIC_LIB is defined? Based on the assumption most of the users just need the shared libs... that's our case, at least.

t-b:
@lorenzopivetta 

> I suggest the opposite? We always build the shared lib and only build the static library if BUILD_STATIC_LIB is 
> defined? Based on the assumption most of the users just need the shared libs... that's our case, at least.

This would break @JeanLucPons use case as he requested in https://github.com/tango-controls/cppTango/issues/740#issue-655658461:

> I don't have omniORB and zmq compiled in shared, so the compilation of tango SHARED failed and prevent the 
> build process to end properly. 

So I think the ability to choose one of them is key to flexibility.

lorenzopivetta:
Agreed. So, do wee need two defines, e.g. BUILD_STATIC_LIB and BUILD_SHARED_LIB, at least one mandatory?

t-b:
> Agreed. So, do wee need two defines, e.g. BUILD_STATIC_LIB and BUILD_SHARED_LIB, at least one mandatory?

My proposal is to have `BUILD_SHARED_LIBS,` which is the variable CMake suggests. If true the shared library is built, if false the static library.

lorenzopivetta:
This means that you either build the static or the shared lib. Fine with me. Which the default will be?

t-b:
> This means that you either build the static or the shared lib. Fine with me. Which the default will be?

Yes, default will be shared.

.travis/test.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 
+if [ $RUN_TESTS = "OFF" ]
```
bourtemb, line 3: I know there is no problem in our current usage because RUN_TESTS should always be initialized, but just in case:
```suggestion
if [ "$RUN_TESTS" = "OFF" ]
```

The same for the COVERALLS test a few lines below.

bourtemb (approved):
Looks good to me.
Just a small suggestion to improve a test in a bash script (Could be done in another PR because I think there are some other locations having similar code)

PR #766, opened 2020-09-10
Keep source tree clean and use more targets

t-b:
This also makes executing cmake faster.


configure/cmake_pch_linux.cmake
```diff
@@ -73,6 +73,6 @@ function(tango_add_pch pch_target ref_target pch_extra_cxx_flags)
         MAIN_DEPENDENCY ${tango_h_file})
 
     add_custom_target(${pch_target}
-        DEPENDS ${tango_pch_h_file} ${tango_pch_file})
+        DEPENDS ${tango_pch_h_file} ${tango_pch_file} idl_objects)
```
mliszcz, line 76: Since you are interested in optimizing the build steps: I don't think such a strong dependency is necessary as we really only need to depend on existence of `tango.h`, without compiling `tangoSK.cpp` (which may be done later / in parallel). Having said that, I'm also ok with what what you currently did. Thanks!

mliszcz (approved):
LGTM

configure/cmake_pch_linux.cmake
t-b: I've tried with depending on `tango.h` only earlier, but that did not work.

mliszcz: Maybe you need a full path: `${CMAKE_BINARY_DIR}/cppapi/server/idl/tango.h`. But I haven't tried that.

PR #770, opened 2020-09-11
Fix and document windows build

t-b:


INSTALL.md
```diff
@@ -205,3 +205,51 @@ TANGO_TEST_CASE_SKIP_FIXTURE=1 ctest -V -R ds_cache
...
 docker stop tango_cs mysql_db
 ```
+
+# Building on windows
+
+For the majority of users using the prebuilt binaries from the release page is
+easier. The following documentation is targeted for developers.
+
+We assume Windows 10 and a Visual Studio 2017 development environment. In
+addition python 3.7 must be installed (this is required by omniidl), get it
```
mliszcz, line 215: Why we require exactly python 3.7? What about 3.8 and 3.9? README.win32.txt in omniORB-4.2.3 says only that: *omniidl requires Python 2.7 or 3.x.*

mliszcz (approved):
LGTM! I just left a comment regarding required python version.

INSTALL.md
t-b: The omniorb binaries we are using are built in https://github.com/tango-controls/omniorb-windows-ci/ and these are currently built against libpython-3.7.dll. It's a kludge to install python manually to fullfill the dependency. Ideally we would compile omniorb without python support.

mliszcz: I didn't know about that, thanks for clarification!

PR #776, opened 2020-10-08
Add more classes to documentation

t-b:
Found in https://github.com/tango-controls/cppTango-docs/issues/3.


mliszcz (approved):
LGTM, but I have not checked if we include everything now (if not, I'm assuming tango-controls/cppTango-docs#3 will come back)

t-b:
> LGTM, but I have not checked if we include everything now (if not, I'm assuming tango-controls/cppTango-docs#3 will come back)

Yes, if something is still missing we do that.

PR #781, opened 2020-10-18
Remove tango has log4tango define

t-b:
Should be merged after #759.


bourtemb:
@taurel and @nleclercq , I think we would need your expertise here to know whether we should keep this possibility to compile/run cppTango without log4tango or drop it.
Do you know whether someone is using cppTango without log4tango?
Do you think we should keep this possibility?
Since cppTango 9.3.3, liblog4tango has been merged into libtango.

nleclercq:
I would say that the logging service is part of the Tango DNA. IHMO, it makes no sense to keep the possibility to disable the logging mechanism. The  TANGO_HAS_LOG4TANGO is there for pure historical reasons. The log4tango based logging implementation used to be a proposal. It seems that we finally adopted it  :-)

t-b:
@nleclercq Thanks. So we can do that change.

PR #782, opened 2020-10-18
configure/CMakeLists.txt: Remove bogus omniidl check

t-b:
Since 40ffd174 (cppapi/server/idl/CMakeLists.txt: Rework completely, 2019-11-13)
we search for the omniidl executable in
cppapi/server/idl/CMakeLists.txt and abort if we could not find it.

So let's remove that duplicated part here.


mliszcz (approved):
LGTM

bourtemb (approved):
LGTM
Thanks @t-b

Issue #617, opened 2019-11-05
Remove internal jpeg implementation


t-b:
In cppapi/server/jpeg we have our own modified fork of libjpeg. Instead of that we should remove that code and use libjpeg-turbo.

- [ ] This dependency should be optional, by default we require it but allow the user to not-require it.
- [ ] Add tests which uses jpeg encoded attributes reading and writing

PR #783, opened 2020-10-18
cppapi/server/jpeg_mmx/jpeg_dct_mmx.cpp: Remove calling GCC internal functions

t-b:
The JPEG code has a MMX assembler routines which have optimized code for
32bit OS.

This code uses GCC internals like __i686.get_pc_thunk.bx, which loads
the position of the code into a register. This is required for position
independent code [1].

The name of that function changed in the GCC repository in a3330c9d
(morestack.S (__i686.get_pc_thunk.bx): Rename to..., 2011-05-04).

So let's use the new name.

This bug was already reported earlier [2], but the fix never made it
into the code.

[1]: https://stackoverflow.com/questions/6679846/what-is-i686-get-pc-thunk-bx-why-do-we-need-this-call
[2]: https://sourceforge.net/p/tango-cs/bugs/811/

### DIFF
--- a/cppapi/server/jpeg_mmx/jpeg_dct_mmx.cpp
+++ b/cppapi/server/jpeg_mmx/jpeg_dct_mmx.cpp
@@ -251,17 +251,10 @@ ALIGN8 short __jpmm_row_tab_frw[] = {  // forward_dct coeff table
 	"pmulhw	"#REG1", QWORD PTR [esp]	\n" \
 	"add	esp,8	\n"
 
-#if __GNUC__ > 3
-#define lea_addr_in_reg(REG1) \
-	"call	__i686.get_pc_thunk.bx	\n" \
-	"add    ebx,_GLOBAL_OFFSET_TABLE_	\n" \
-	"lea	"#REG1",__jpmm_row_tab_frw@GOTOFF	\n"
-#else
 #define lea_addr_in_reg(REG1) \
 	"call	__tg_get_pc	\n" \
 	"add    ebx,_GLOBAL_OFFSET_TABLE_	\n" \
 	"lea	"#REG1",__jpmm_row_tab_frw@GOTOFF	\n"
-#endif
 
 #else /* __PIC__ */
 
@@ -1060,21 +1053,12 @@ void jpeg_fdct_mmx( short *block )
   );
 
 #ifdef __PIC__
-#if __GNUC__ > 3
- __asm__(
- 	"push 	%ebx		\n"
-	"call	__i686.get_pc_thunk.bx	\n"
-	"add    $_GLOBAL_OFFSET_TABLE_, %ebx	\n"
-	"lea	__jpmm_row_tab_frw@GOTOFF(%ebx),%ebx	\n"
-   );
-#else
  __asm__(
  	"push 	%ebx		\n"
 	"call	__tg_get_pc	\n"
 	"add    $_GLOBAL_OFFSET_TABLE_, %ebx	\n"
 	"lea	__jpmm_row_tab_frw@GOTOFF(%ebx),%ebx	\n"
    );
-#endif
 #endif /* __PIC__ */
 
   // Rows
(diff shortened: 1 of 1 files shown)


mliszcz (dismissed):
LGTM, but at this point I think we should simply consider removing any MMX code we have (maybe after doing some local benchmarking with and without). Using MMX is probably a pessimization and if one is interested in SIMD and perfornamce, better results can be achieved letting the compiler to generate the assembly (and use SSE, AVX or whatever the CPU supports, not limiting ourselves to MMX).

Not to mention the limited maintainability of assembly code compared to C++.

https://stackoverflow.com/questions/44299401/difference-between-mmx-and-xmm-register
> Nowadays they are just a historical oddity, I don't think anybody actually uses MMX anymore, as it has been completely superseded by the various SSE extensions. Edit: as Peter Cordes points out in the comments, there is still quite some MMX code around. 

https://stackoverflow.com/questions/31490853/are-different-mmx-sse-and-avx-versions-complementary-or-supersets-of-each-other
> If CPU-compatibility was a non-issue, the legacy-SSE versions of vector instructions would be truly obsolete, like MMX is now.

t-b:
@mliszcz I would go one step further and use a third-party libjpeg library. What about that? See also #617.

So we would use this patch only as a first step.

mliszcz:
@t-b Yes, but I'd make such dependency optional, because probably not everyone is using that. If library is not found, we simply do not provide functions to encode/decode JPEG images into DevEncoded. IMO it should be device server's responsibility to provide data encoding functions for whatever format it is using.

cppapi/server/jpeg_mmx/jpeg_dct_mmx.cpp
```diff
@@ -253,7 +253,7 @@ ALIGN8 short __jpmm_row_tab_frw[] = {  // forward_dct coeff table
 
 #if __GNUC__ > 3
 #define lea_addr_in_reg(REG1) \
-	"call	__i686.get_pc_thunk.bx	\n" \
+	"call	__x86.get_pc_thunk.bx	\n" \
```
bourtemb: When looking at the commit @t-b referenced in his commit message:
https://github.com/gcc-mirror/gcc/commit/a3330c9d558720697db4f6f767252335eefe89a3
it seems like this commit is part of gcc releases >= 4.7.0.
So what happens for 4.0.0 <= gcc versions < 4.7.0?
This is not so important for tango-9-lts since it requires recent gcc versions for the C++11 features, but this info could be useful if we backport the fix into 9.3-backports.
For tango-9-lts, we could also remove at some point the #ifdef sections when `__GNUC__` <= 4 I think.

bourtemb:
> IMO it should be device server's responsibility to provide data encoding functions for whatever format it is using.

I think the original goal was to simplify the device server programmer job by providing some methods to optimize image transfer using the jpeg format.
This is quite a common use case. This is why it was added.
It is even documented:
https://tango-controls.readthedocs.io/en/latest/development/advanced/transferring-images.html?#transferring-images
Of course this is something that can be rediscussed.

I think it is not stupid to add an optional dependency to an external jpeg encoding library. There must be some very reliable and maintained libraries around.

cppapi/server/jpeg_mmx/jpeg_dct_mmx.cpp
t-b: Valid point. How about removing the `__GNUC__ > 3` branch and always use the fallback? The fallback code uses the exact same code as the GCC code.

See
https://github.com/tango-controls/cppTango/blob/8d0bd12e18aa7ea5dad4642790e68499705623b7/cppapi/server/jpeg_mmx/jpeg_dct_mmx.cpp#L1974-L1990

vs

 https://github.com/gcc-mirror/gcc/commit/a3330c9d558720697db4f6f767252335eefe89a3#diff-cb9d1aa158c2bbb17eb20b4b216cfd3f93dd41d65332c9d70991a829b14aaccaL463

t-b: @mliszcz @bourtemb I went that route and removed the GCC specific code.

mliszcz (approved):
LGTM!

t-b:
@bourtemb +1 or -1?

PR #794, opened 2020-10-22
Use chrono to represent log4tango timestamps

mliszcz:
With this PR LOG4TANGO_HAVE_INT64_T is no longer needed and can be removed.


t-b (dismissed):
Nice!

t-b:
@mliszcz And this also kills the only users of `LOG4TANGO_HAVE_GETTIMEOFDAY` and `LOG4TANGO_HAVE_GETTIMEOFDAY` thus making this obsolete as well. I've removed that as well in https://github.com/t-b/cppTango/commit/9f54f053aada92e2f2d58b2291218eeb010b7244. Feel free to grab and munge into your commit.

mliszcz:
Thanks @t-b. Won't this introduce any conflicts with #759? I was thinking that we merge this one and then #759 will clean up the build system.

t-b:
@mliszcz It interferes anyway and I also don't fear conflicts. I'm also in favour of merging this one here first and then #759. The main reason for my patch above is that we have everything in the right place. And here you are removing the users of these defines, so I think here is also the right place to remove setting them.

mliszcz:
@t-b done. I reworded your commit message because it was just 'changes'. Please re-review.

t-b (approved):

PR #808, opened 2020-11-17
Search docker for tests

t-b:
I had that issue when I played around with cppTango on my mac. We now warn when building the tests and docker is not present, and also error out when running the tests and docker is missing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,6 +53,12 @@ add_subdirectory("log4tango")
 add_subdirectory("cppapi")
 
 if(BUILD_TESTING)
+
+  find_program(DOCKER_BINARY docker)
+  if(NOT DOCKER_BINARY)
+      message(WARNING "The tests can not be run as docker is missing.")
+  endif()
+
   if(BUILD_SHARED_LIBS)
     add_subdirectory("cpp_test_suite")
   else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,7 +57,7 @@ if(BUILD_TESTING)
     add_subdirectory("cpp_test_suite")
   else()
     message(WARNING "Not building the tests, because that is currently supported only when BUILD_SHARED_LIBS is ON")
-    SET(BUILD_TESTING OFF)
+    set(BUILD_TESTING OFF)
   endif()
 endif()
 
--- a/cpp_test_suite/environment/run_with_fixture.sh.cmake
+++ b/cpp_test_suite/environment/run_with_fixture.sh.cmake
@@ -8,6 +8,11 @@ if [ $# -lt 1 ]; then
     exit 1
 fi
 
+if ! hash docker 2>/dev/null; then
+  echo "Can not run the tests as docker is missing"
+  exit 1
+fi
+
 tc_program="$1"
 tc_run_name="$(basename "$tc_program")_$(date '+%Y%m%d.%H%M%S.%N')"
 shift 1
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -53,11 +53,16 @@ add_subdirectory("log4tango")
 add_subdirectory("cppapi")
 
 if(BUILD_TESTING)
-  if(BUILD_SHARED_LIBS)
+  find_program(DOCKER_BINARY docker)
+
+  if(NOT DOCKER_BINARY)
+      message(WARNING "Not building the tests, because docker is not present.")
+      set(BUILD_TESTING OFF)
```
mliszcz: Will this handle following scenario without the need to recompile the whole library (server_objects + client_objects)?:
1. `cmake ..`
2. `make`
3. install docker
4. `ctest`

I would just print the warning that the test won't run without docker, without disabling any flags. Lack of docker during configuration time does not prevent us from compiling the tests and device servers. And has nothing to do with presence of docker later, when one may want to run the tests.

Also, we may want to compile individual tests or device server for e.g. debugging or demo purposes and just start the database manually on the host.

If failures with obscure errors are your concern, I suggest to check for docker presence from the test runner script and then print a message and fail the test. What do you think?

t-b: @mliszcz Thanks for the input. I'll keep the warning but still compile the tests. And I'll fail test execution early if docker is missing.

t-b: @mliszcz Done.

CMakeLists.txt
```diff
@@ -53,11 +53,15 @@ add_subdirectory("log4tango")
...
+  if(NOT DOCKER_BINARY)
+      message(WARNING "The tests can not be run as docker is missing.")
+  elseif(BUILD_SHARED_LIBS)
     add_subdirectory("cpp_test_suite")
   else()
     message(WARNING "Not building the tests, because that is currently supported only when BUILD_SHARED_LIBS is ON")
-    SET(BUILD_TESTING OFF)
+    set(BUILD_TESTING OFF)
   endif()
 endif()
```
mliszcz: I think these if-elseif-else are a bit hard to read, partly because the 'usual' execution path is in the middle elseif. I'd print both warning messages even if we don't have docker and if we are doing static build because these two things are really unrelated.

Also note that currently if we don't have docker and are doing a static build build, we will not clear the BUILD_TESTING option, which may be a problem if we attempt to check this option later or in some other cmake file.

I suggest to do a minimal change. Either check for docker binary just after `add_subdirectory(...)`:
```cmake
if (BUILD_TESTING)
  if (BUILD_SHARED_LIBS)
    add_subdirectory("cpp_test_suite")

    # added
    find_program(DOCKER_BINARY docker)
    if (NOT DOCKER_BINARY)
      message()
    endif()
  endif()
endif()
```
Or if you don't like double-nesting, after we check for static build:
```cmake
if (BUILD_TESTING)
  if (BUILD_SHARED_LIBS)
    add_subdirectory("cpp_test_suite")
  else()
    set(BUILD_TESTING OFF)
  endif()
endif()

# added
if (BUILD_TESTING)
  find_program(DOCKER_BINARY docker)
  if (NOT DOCKER_BINARY)
    message()
  endif()
endif()
```

mliszcz: Actually docker check can go straight into the first BUILD_TESTING check. User intention was to run the tests so I think we can keep the warning even if we disabled tests for any other reason.

```cmake
if (BUILD_TESTING)
  # added
  find_program(DOCKER_BINARY docker)
  if (NOT DOCKER_BINARY)
    message()
  endif()

  if (BUILD_SHARED_LIBS)
    add_subdirectory("cpp_test_suite")
  else()
    set(BUILD_TESTING OFF)
  endif()
endif()
```

t-b: @mliszcz Jeep. I've used the last suggestion. Thanks.

mliszcz (approved):
LGTM

PR #817, opened 2020-11-26
Fix API_EventTimeOut

### LINKED ISSUE #686: Unsubscribe in push_event leads to API_EventTimeout
We encountered this issue when investigating https://github.com/tango-controls/pytango/issues/292.

The architecture is as following. We have one DS called `DeviceServer` which
exports in total two devices, each of a different class. These two classes are:
* `Device1` with one attribute `attr1`
* `Device2` with two attributes `attr1` and `attr2`

What we will do:
1. Subscribe to `attr1` configuration events of the `Device1` device.
2. Subscribe to `attr1` change events of the `Device2` device.
3. Subscribe to `attr2` change events of the `Device2` device
   with a callback which will unsubscribe from the previously subscribed
   `attr1` configuration events of the `Device1` device.  

This leads to the situation where the client stops receiving events and just get's `API_EventTimeout`.

Thanks to the example provided by @schooft we could also reproduce it with C++ client.

More details and the exemplary code can be found in https://github.com/reszelaz/test-tango-unsub.

t-b:
Forward port #699.

Close #686.

### DIFF
--- a/cppapi/client/zmqeventconsumer.cpp
+++ b/cppapi/client/zmqeventconsumer.cpp
@@ -698,6 +698,9 @@ bool ZmqEventConsumer::process_ctrl(zmq::message_t &received_ctrl,zmq::pollitem_
 {
     bool ret = false;
 
+    // static variable to count the number of ZMQ_DELAY_EVENT requests currently in progress:
+    static int nb_current_delay_event_requests = 0;
+
 //
 // For debug and logging purposes
 //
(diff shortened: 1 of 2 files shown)


cppapi/client/zmqeventconsumer.cpp
```diff
@@ -698,6 +698,9 @@ bool ZmqEventConsumer::process_ctrl(zmq::message_t &received_ctrl,zmq::pollitem_
 {
     bool ret = false;
 
+    // static variable to count the number of ZMQ_DELAY_EVENT requests currently in progress:
+    static int nb_current_delay_event_requests = 0;
```
mliszcz: Can you change this to a member variable? This was discussed already during the review of the original PR but postponed due to ABI compatibility. Adding more global state is going to be a problem if we at some point decide to delete or recreate instance of this class.

t-b: @bourtemb Although I created the PR, can I pass that onto you?

mliszcz: ~~Sure I'll do that.~~
EDIT: Sorry! I thought this comment was addressed to me.

mliszcz (approved):

t-b:
@bourtemb Are we good to go? Although you did the code, as I've opened the PR, it is you who has to review.

bourtemb (approved):

Issue #816, opened 2020-11-20
Limit of six device classes in a device server - FileDatabase::DbGetClassProperty bug


DrewDevereux:
After experiencing "unidentifiable C++ exception" when trying to add a device to pytango's `DeviceTestContext`, I dug into the code, and eventually found that a device server can have any number of device instances of **up to six device classes**. But as soon as a seventh device class is added, the device server cannot run. The following python code illustrates this problem. The device server runs successfully if `include_seventh_device` is `False`. If set to `True`, then the device server fails to run.
```
from tango import Database
from tango.server import Device, run
from tango.test_context import get_host_ip

class A(Device): pass
class B(Device): pass
class C(Device): pass
class D(Device): pass
class E(Device): pass
class F(Device): pass
class G(Device): pass

# Write device info to text file database
with open("db_file", "w") as f:
    f.write("Debug/debug/DEVICE/A:test/a/1\n")
    f.write("Debug/debug/DEVICE/B:test/b/1\n")
    f.write("Debug/debug/DEVICE/C:test/c/1\n")
    f.write("Debug/debug/DEVICE/D:test/d/1\n")
    f.write("Debug/debug/DEVICE/E:test/e/1\n")
    f.write("Debug/debug/DEVICE/F:test/f/1\n")
    f.write("Debug/debug/DEVICE/G:test/g/1\n")

include_seventh_device = False

# run server
run(
    [A, B, C, D, E, F] + [G]*include_seventh_device,
    args=[
        "Debug",
        "debug",
        "-ORBendPoint",
        f"giop:tcp:{get_host_ip()}:57061",
        "-file=db_file"
    ]
)
```

bourtemb:
> 
> 
> I wild guess would be that `i` -> `j` in this line.

At first sight, I would replace 
`(*data_out)[index] = Tango::string_dup((*data_in)[i + 1].in());index++;`
with 
`(*data_out)[index] = Tango::string_dup((*data_in)[j].in());index++;`

as it is done in FileDatabase::DbGetDeviceProperty  so `i+1` -> `j` 

But deeper investigations are required because there is no doxygen documentation in the code listing what these methods are taking as input and what they are returning.
We need to refer to the equivalent commands in the Tango Database server...

bourtemb:
If we refer to DbGetClassProperty command in the Tango Database server, here are the argin and argout expected arguments:

```
Argin description:
Str[0] = Tango class
Str[1] = Property name
Str[2] = Property name
```

[...]

PR #818, opened 2020-12-01
Fix a bug in FileDatabase::DbGetClassProperty (#816)

bourtemb:
Add documentation to FileDatabase::DbGetClassProperty method
Translate some comments from Italian to English

### DIFF
--- a/cppapi/client/filedatabase.cpp
+++ b/cppapi/client/filedatabase.cpp
@@ -1610,7 +1610,6 @@ CORBA::Any*   FileDatabase :: DbGetClassProperty(CORBA::Any& send)
 	const Tango::DevVarStringArray* data_in = NULL;
 	char num_prop_str[256];
 	char num_vals_str[256];
-	unsigned int num_prop = 0;
 	unsigned int num_val = 0;
 
 	cout4 << "FILEDATABASE: entering DbGetClassProperty" << endl;
@@ -1624,7 +1623,7 @@ CORBA::Any*   FileDatabase :: DbGetClassProperty(CORBA::Any& send)
 
 	data_out->length(2);
 	(*data_out)[0] = Tango::string_dup((*data_in)[0]); index++;
-	num_prop = data_in->length() - 1;
+	const unsigned int num_prop = data_in->length() - 1;
 	sprintf(num_prop_str,"%ud",num_prop);
 	(*data_out)[index] = Tango::string_dup(num_prop_str); index++;
 
@@ -1644,7 +1643,6 @@ CORBA::Any*   FileDatabase :: DbGetClassProperty(CORBA::Any& send)
 				{
 					if (equalsIgnoreCase((*data_in)[j].in(), m_server.classes[i]->properties[m]->name))
 					{
-						num_prop++;
 						num_val = m_server.classes[i]->properties[m]->value.size();
 						seq_length = seq_length + 2 + num_val;
 						(*data_out).length(seq_length);
--- a/cppapi/client/filedatabase.cpp
+++ b/cppapi/client/filedatabase.cpp
@@ -1580,7 +1580,30 @@ CORBA::Any*   FileDatabase :: DbDeleteDeviceAttributeProperty(CORBA::Any& send)
 
 }
 
-
+/*
+ * FileDatabase :: DbGetClassProperty(CORBA::Any& send)
+ * @param send CORBA::Any Argin representing a DevVarStringArray
+ * Argin description:
+ * Str[0] = Tango class
+ * Str[1] = Property #1 name
+ * Str[2] = Property #2 name
+ * ...
+ * @return DevVarStringArray as CORBA::Any
+ * Argout description:
+ * Str[0] = Tango class
+ * Str[1] = Number of properties
+ * Str[2] = Property #1 name
+ * Str[3] = Property #1 value size (number of value elements, 1 for scalar case, 0 if class property not found)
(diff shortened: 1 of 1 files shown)


cppapi/client/filedatabase.cpp
```diff
@@ -1612,8 +1635,8 @@ CORBA::Any*   FileDatabase :: DbGetClassProperty(CORBA::Any& send)
 	{
 		if (equalsIgnoreCase((*data_in)[0].in(), m_server.classes[i]->name))
 		{
-			// m_server.classes[i] e' la classe che cerchiamo
-			for (unsigned int j = 1; j < (*data_in).length(); j++)  // in 0 c'e' il nome della classe e poi a seguire le proprieta
+			// m_server.classes[i] is the class we are looking for
+			for (unsigned int j = 1; j < (*data_in).length(); j++)  // at index 0 is the name of the class, property names are following
 			{
 				unsigned long nb_prop_defined = m_server.classes[i]->properties.size();
```
mliszcz, line 1640: Consider adding a variable that would hold property name for the current loop iteration.
```cpp
auto const &prop_name = (*data_in)[j];
```
This `(*data_in)[j]` is repeated three times below. Not only this is unreadable and hard to maintain but also was the source of this bug.

cppapi/client/filedatabase.cpp
```diff
@@ -1642,11 +1665,13 @@ CORBA::Any*   FileDatabase :: DbGetClassProperty(CORBA::Any& send)
 
 				if (m == nb_prop_defined)
 				{
+					// The requested property does not exist in the specified class
 					seq_length = seq_length + 2;
 					data_out->length(seq_length);
-					(*data_out)[index] = Tango::string_dup((*data_in)[i + 1].in());index++;
+					// The requested property name is returned, followed by a 0,
+					// meaning the length of this property value is 0 ( <=> class property not found )
+					(*data_out)[index] = Tango::string_dup((*data_in)[j].in());index++;
```
mliszcz, line 1671: This `index++` scattered all over the place is not only ugly but dangerous if one forgets to add it. There are few options that we could use to avoid that, at least these come to my mind:

1. Have a lambda that increases this index for us. But this could only be done in 9.4.
    ```cpp
    auto const push_data_out = [index = 0u](auto const &elem) mutable
    {
        (*data_out)[index] = Tango::string_dup(elem);
        index++;
    };
    ```
    Here we could simply call:
    ```cpp
    push_data_out(prop_name);
    push_data_out("0");
    ```

2. Collect all data in a vector and then copy that into the sequence. Note that we want to copy only the pointers. And it's not exception-safe but this is non-throwing code.
    ```cpp
    std::vector<char*> output{};
    output.push_back(Tango::string_dup("whatever"));

    auto *result = new DevVarStringArray{};
    result.length(output.size());
    std::memcpy(result.get_buffer(), output.data(), output.size() * sizeof(char*));
    ```

3. Write some utility classes which will provide a sane API for CORBA sequences. Maybe push_back like vector has, iteration support like in std;:span, etc. But this is something that we may rather consider for 9.4, not for backports.

mliszcz (commented):
I have some general comments regarding the implementation of this function. Maybe we could use this opportunity to do a small cleanup here. Github does not allow to put comments on lines not touched by the PR, so I'll leave them here:

https://github.com/tango-controls/cppTango/blob/1ddda45603e02ac279e9da6dd88e05da95542b61/cppapi/client/filedatabase.cpp#L1624
Unneeded and also incorrect because earlier we assign `num_prop = data_in->length() - 1;`. This variable should be declared const.

https://github.com/tango-controls/cppTango/blob/1ddda45603e02ac279e9da6dd88e05da95542b61/cppapi/client/filedatabase.cpp#L1629-L1633
Multiple problems here but main one is that `value` is a vector, c++ requires `size()` to return unsigned integer, printing with `%d` is at least suspicious. I haven't checked printf implementation but I guess there is a reason why they have `%u`. I'd replace this with a simple:
```
sprintf(num_vals_str, "%lu", static_cast<unsigned long>(num_val))
```

t-b:
@mliszcz This PR targets 9.3-backports so I would favour the smallest possible bugfix. Your cleanup is IMHO good for tango-9-lts. Regrading the snprintf thing I have something in https://github.com/tango-controls/cppTango/pull/792/files#diff-bdc1f56f72d7a12798a63b803aebf5a6b49233a15c0cdefb705572c03d23f566 which deals with that. The PR is not yet done though.

mliszcz:
@t-b sure, these are just comments, Reynald may not agree with them or drop or postpone them, that's how reviews work. I put them here mainly to allow us re-consider these changes when forwardport PR comes. Otherwise I'd probably forget what I learnt about this code today.

bourtemb:
> I have some general comments regarding the implementation of this function. Maybe we could use this opportunity to do a small cleanup here. Github does not allow to put comments on lines not touched by the PR, so I'll leave them here:
> 
> https://github.com/tango-controls/cppTango/blob/1ddda45603e02ac279e9da6dd88e05da95542b61/cppapi/client/filedatabase.cpp#L1624
> 
> Unneeded and also incorrect because earlier we assign `num_prop = data_in->length() - 1;`. This variable should be declared const.

Indeed. Thanks @mliszcz . These suggestions should be implemented in 1bd6d38f902a8dd6195870dd3042d0fdfea2911f

t-b (approved):

mliszcz (approved):

Issue #805, opened 2020-11-17
Add CI job with clang tidy


t-b:
This can already be done locally

```
cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=true ..
clang-tidy $(find ../cppapi -iname "*.cpp")
```

which gives 
[warnings.zip](https://github.com/tango-controls/cppTango/files/5553188/warnings.zip)

and manually filtered

```
$grep "warning:" warnings.log | grep -oP "\[clang.*\]" | sort | uniq -c
     18 [clang-analyzer-core.CallAndMessage]
     30 [clang-analyzer-core.NonNullParamChecker]
     71 [clang-analyzer-core.NullDereference]
      8 [clang-analyzer-core.UndefinedBinaryOperatorResult]
      4 [clang-analyzer-core.uninitialized.Assign]
     12 [clang-analyzer-core.uninitialized.UndefReturn]
      2 [clang-analyzer-cplusplus.NewDelete]
      2 [clang-analyzer-cplusplus.NewDeleteLeaks]
     26 [clang-analyzer-deadcode.DeadStores]
     22 [clang-analyzer-optin.cplusplus.VirtualCall]
    135 [clang-analyzer-security.insecureAPI.strcpy]
      3 [clang-analyzer-unix.MismatchedDeallocator]
      3 [clang-diagnostic-absolute-value]
      2 [clang-diagnostic-instantiation-after-specialization]
      4 [clang-diagnostic-string-plus-int]
```

mliszcz:
> This can already be done locally

Still I think it would be useful to consider:
* adding .clang-tidy config file with more checks enabled (especially checks related to code readability) and some other disabled (like maybe the one for non-optimal padding, which generates a lot of noise)
* running multiple tidy instances in parallel with run-clang-tidy.py script or gnu parallel or something else (related: https://reviews.llvm.org/D54141)
* ensure that we don't see duplicates coming from the same header included in multiple TUs
* ensure that warnings coming from third-party libraries are not reported.

For instance I remember seeing NullDereference reported multiple times (probably false positive) in omniORB's stringtypes.h and seqTemplatedecls.h when running clang-tidy on our attribute.cpp. I could not get rid of this neither with `-line-filter` nor with `-header-filter`.

PR #820, opened 2021-01-03
Gitlab CI migration

mliszcz:
This PR adds configuration for Gitlab CI/CD pipelines and introduces necessary test code changes to make them pass on Gitlab.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -106,7 +106,6 @@ abi-api-compliance-check:
         .travis/check-ABI-API-compliance.sh || true
   after_script:
     - docker stop cpp_tango
-    - gzip -r compat_reports
   artifacts:
     when: always
     paths:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -57,6 +57,7 @@ services:
     - test ${SONAR_SCANNER} = "ON" && .travis/sonar.sh || true
   after_script:
     - docker stop cpp_tango
+    - find build -name 'core.*' -exec gzip "{}" \;
     - mkdir -p build/cpp_test_suite/test_results
     - gzip -r build/cpp_test_suite/test_results
   artifacts:
@@ -137,6 +138,7 @@ debian10:
     TANGO_ENABLE_COVERAGE: "ON"
     OS_TYPE: debian10
   after_script:
+    - find build -name 'core.*' -exec gzip "{}" \;
     - mkdir -p build/cpp_test_suite/test_results
     - gzip -r build/cpp_test_suite/test_results
     - mkdir coverage
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -75,6 +75,7 @@ abi-api-compliance-check:
   image: debian:10
   before_script:
     - apt-get update && apt-get install -y git docker.io
+    - git fetch origin "$CI_MERGE_REQUEST_TARGET_BRANCH_NAME"
     - .travis/before_install.sh
     - docker build -t cpp_tango .travis/gcc-latest
     - >
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -150,15 +150,15 @@ debian10:
     - mkdir coverage
     - >
         docker exec --workdir /home/tango/src cpp_tango
-        gcovr --filter '^cppapi/' --filter '^log4tango/' -j$(nproc)
+        gcovr --filter '^cppapi/' --filter '^log4tango/(?!tests/)' -j$(nproc)
         --xml --output coverage.xml
     - >
         docker exec --workdir /home/tango/src cpp_tango
-        gcovr --filter '^cppapi/' --filter '^log4tango/' -j$(nproc)
+        gcovr --filter '^cppapi/' --filter '^log4tango/(?!tests/)' -j$(nproc)
         --html-details --output coverage/coverage.html
     - >
         docker exec --workdir /home/tango/src cpp_tango
-        gcovr --filter '^cppapi/' --filter '^log4tango/' -j$(nproc)
+        gcovr --filter '^cppapi/' --filter '^log4tango/(?!tests/)' -j$(nproc)
     - tar czf coverage.tar.gz coverage
     - docker stop cpp_tango
   artifacts:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -14,6 +14,13 @@ variables: &variables
 services:
   - docker:20.10.1-dind
 
+# See: https://docs.gitlab.com/ce/ci/yaml/README.html#workflowrules-templates
+workflow:
+  rules:
+    - if: $CI_MERGE_REQUEST_IID
+    - if: $CI_COMMIT_TAG
+    - if: $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH
+
 .job-template: &job-template
   image: debian:10
   before_script:
@@ -70,8 +77,9 @@ abi-api-compliance-check:
   variables:
     <<: *variables
     RUN_TESTS: "OFF"
-  only:
-    - merge_requests
+  rules:
+    # This job runs only for merge requests.
+    - if: $CI_MERGE_REQUEST_IID
   image: debian:10
   before_script:
     - apt-get update && apt-get install -y git docker.io
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,203 @@
+variables: &variables
+  DEBIAN_FRONTEND: noninteractive
+  DOCKER_HOST: tcp://docker:2375
+  # Build parameters. We later override them at job level if needed.
+  SONAR_SCANNER: "OFF"
+  COVERALLS: "OFF"
+  RUN_TESTS: "ON"
+  STOCK_CPPZMQ: "ON"
+  USE_PCH: "ON"
+  CMAKE_BUILD_TYPE: "Debug"
+  TANGO_USE_USING_NAMESPACE: "ON"
+  BUILD_SHARED_LIBS: "ON"
+
+services:
+  - docker:20.10.1-dind
+
+.job-template: &job-template
+  image: debian:10
+  before_script:
+    - ulimit -c unlimited
+    - echo "core.%e.%p.%t" > /proc/sys/kernel/core_pattern
+    - apt-get update && apt-get install -y git wget unzip docker.io
+    - .travis/before_install.sh
+    - docker build -t cpp_tango .travis/${OS_TYPE}
+    - >
+        docker run
+        --rm
+        --name cpp_tango
+        --user root
+        -e BINTRAY_API_KEY=${CI_BINTRAY_API_KEY}
+        -e BINTRAY_USER_NAME=tango-ci
+        -e COVERALLS_REPO_TOKEN=${COVERALLS_REPO_TOKEN}
+        -e SONAR_TOKEN=${SONAR_TOKEN}
+        -e TANGO_HOST=${TANGO_HOST}
+        -e TRAVIS_BRANCH=${TRAVIS_BRANCH}
+        -e TRAVIS_JOB_ID=${TRAVIS_JOB_ID}
+        -e DOCKER_HOST=tcp://$(getent hosts docker | awk '{ print $1 }'):2375
+        -v `pwd`:/home/tango/src
+        -v `pwd`/.sonar:/home/tango/.sonar
+        -v `pwd`/idl:/home/tango/idl
+        -v `pwd`/cppzmq:/home/tango/cppzmq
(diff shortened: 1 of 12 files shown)


mliszcz:
Note that for now I disabled two tests (one testcase in CXX::cxx_dserver::misc and event::change_event). I do not know how to correct them.

mliszcz:
Also note that the same branch is pushed to the test gitlab repository: https://gitlab.com/tango-controls/test_gh_import/cppTango/-/tree/gitlab-ci-migration

mliszcz:
And FYI we've just hit the dockerhub pull limit:
```
toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit

The command "docker build --build-arg APP_UID=$(id -u) --build-arg APP_GID=$(id -g) -t cpp_tango .travis/${OS_TYPE}
```

andygotz:
Hi @mliszcz - HNY!

How to solve this? Should we upgrade to a TEAM licence or should we host our own docker images? I see the limit is 200 / 6 hours. How many pull requests do we have per build? We could host the docker images on gitlab in the future: https://docs.gitlab.com/ee/user/packages/container_registry/

mliszcz:
Hi Andy!

I think for now we don't need to do anything. As you've said, we are just blocked for six hours. I don't know how many pulls we need. @t-b calculated that some time ago and if I recall we were below the current limits. I exhausted this limit only due to extensive use of CI for testing the new Gitlab CI pipeline for cppTango.

Yes, we can move the images to Gitlab container registry if necessary, this is a good idea IMO. We can also keep the images in dockerhub and just mirror them (manually by `docker push` for now) in gitlab so that cppTango CI pipeline will not contribute to the dockerhub pull limits. This would be very simple to do.

In the long run I think we should stop using docker or any containers for tests in CI because it contributes to the execution time and it increases complexity of our setup. But this would require some changes on database ds side (mainly adding support for SQLite).

.gitlab-ci.yml
```diff
@@ -0,0 +1,203 @@
...
+        --rm
+        --name cpp_tango
+        --user root
+        -e BINTRAY_API_KEY=${CI_BINTRAY_API_KEY}
+        -e BINTRAY_USER_NAME=tango-ci
+        -e COVERALLS_REPO_TOKEN=${COVERALLS_REPO_TOKEN}
+        -e SONAR_TOKEN=${SONAR_TOKEN}
+        -e TANGO_HOST=${TANGO_HOST}
+        -e TRAVIS_BRANCH=${TRAVIS_BRANCH}
+        -e TRAVIS_JOB_ID=${TRAVIS_JOB_ID}
```
bourtemb, line 43: Who is setting these TRAVIS_BRANCH and TRAVIS_JOB_ID variables when running on gitlab?
They seem to be used by sonar.sh.
I guess this will have to be changed for sonar to work properly.
@mliszcz I remember you mentioned on slack in December that we should probably enable Sonar after the migration so we could fix sonar in a PR after the migration (or use other tools to do the same things as sonar).
Maybe we should actually forget about Sonar after the migration and concentrate on issues like https://github.com/tango-controls/cppTango/issues/805

.gitlab-ci.yml
```diff
@@ -0,0 +1,203 @@
...
+        gcovr --filter '^cppapi/' --filter '^log4tango/' -j$(nproc)
+        --html-details --output coverage/coverage.html
+    - >
+        docker exec --workdir /home/tango/src cpp_tango
+        gcovr --filter '^cppapi/' --filter '^log4tango/' -j$(nproc)
+    - tar czf coverage.tar.gz coverage
+    - docker stop cpp_tango
+  artifacts:
+    when: always
+    expire_in: 1h
```
bourtemb: isn't 1h a bit too short?

.gitlab-ci.yml
```diff
@@ -0,0 +1,203 @@
...
+    - .travis/test.sh
+    - set +e
+    - test ${SONAR_SCANNER} = "ON" && .travis/sonar.sh || true
+  after_script:
+    - docker stop cpp_tango
+    - mkdir -p build/cpp_test_suite/test_results
+    - gzip -r build/cpp_test_suite/test_results
+  artifacts:
+    when: always
+    expire_in: 1h
```
bourtemb: isn't 1h a bit too short?

.gitlab-ci.yml
```diff
@@ -0,0 +1,203 @@
...
+        -e CI_TARGET_BRANCH=${CI_MERGE_REQUEST_TARGET_BRANCH_NAME}
+        -e CMAKE_BUILD_PARALLEL_LEVEL=$(nproc)
+        cpp_tango
+        .travis/check-ABI-API-compliance.sh || true
+  after_script:
+    - docker stop cpp_tango
+    - gzip -r compat_reports
+  artifacts:
+    when: always
+    expire_in: 1h
```
bourtemb: isn't 1h a bit too short?
Default is 30 days if we don't precise this keyword I think, which seems more reasonable to me.

bourtemb (commented):
Thanks @mliszcz for this great and useful work.
I would suggest to remove the expiration delay of 1h for some artifacts (default seems to be 30 days which seems more convenient, unless I missed 1 constraint).
I would also suggest to create a Merge Request on our Gitlab Test repository to check whether the abi-compliance-check is correctly executed in this case.
By the way, it's great to have converted abi-compliance-check too.
The modifications for code coverage in CMake files will be useful too.

bourtemb:
I created the Merge Request on the Gitlab test repo to check whether the tests triggered only on Merge Request are working properly.
Here is the MR: https://gitlab.com/tango-controls/test_gh_import/cppTango/-/merge_requests/819

.gitlab-ci.yml
mliszcz: Thanks for pointing those! I haven't noticed these variables when I copy-pasted the instructions. Since it is related to sonar, I propose to leave it as it is, so when we get back to sonar integration after migration we will be able to notice that something still needs to be changed here.

I'm all in for replacing sonar with something that we can run locally and that can produce html report similar to what we have in sonar. It is never good to depend on third-party, proprietary service.

.gitlab-ci.yml
mliszcz: Right. What do you propose? Maybe 1 day would be ok? I'd rather not abuse GL-provided storage with old test logs and core dumps. Also note that artifacts from the last run of each job are kept indefinitely, regardless of this setting.

mliszcz:
Thanks for triggering the ABI/API check. It failed:
```
$ docker exec -w /home/tango/src -e CI_TARGET_BRANCH=${CI_MERGE_REQUEST_TARGET_BRANCH_NAME} -e CMAKE_BUILD_PARALLEL_LEVEL=$(nproc) cpp_tango .travis/check-ABI-API-compliance.sh || true
fatal: No such ref: 'origin/tango-9-lts'
```
Either gitlab uses different remote name or does not fetch all branches. I'll have a look right now.

.gitlab-ci.yml
bourtemb: I would try not to abuse with core files, which can be big indeed.
But I think having several days would be great (so we can launch something on Friday evening and come back on Monday or Tuesday and get the results).
I'm not sure to understand fully your sentence "Also note that artifacts from the last run of each job are kept indefinitely, regardless of this setting."
What happens if someone makes a new commit in another branch? Will this one be considered as the last run?
Or is it the artifacts from the last run from each branch which are kept?

t-b: Why not just use the default for now? We need to buy more project space anyway soon.

mliszcz: I did as Thomas suggested and removed explicit setting of expiration time.

> What happens if someone makes a new commit in another branch? Will this one be considered as the last run?
Or is it the artifacts from the last run from each branch which are kept?

Per my understanding, it keeps all artifacts from all jobs in the last pipeline of each branch, MR and tag:
https://docs.gitlab.com/ee/ci/yaml/#artifactsexpire_in
https://gitlab.com/gitlab-org/gitlab/-/issues/16267
https://gitlab.com/gitlab-org/gitlab/-/issues/229936

.gitlab-ci.yml
mliszcz: Removed

.gitlab-ci.yml
mliszcz: Removed

mliszcz:
API/API compliance check is fixed now: https://gitlab.com/tango-controls/test_gh_import/cppTango/-/jobs/956206341
By default Gitlab runner is not fetching any branches except the current one.

.gitlab-ci.yml
```diff
@@ -0,0 +1,211 @@
...
+    - >
+        docker exec
+        -w /home/tango/src
+        -e CI_TARGET_BRANCH=${CI_MERGE_REQUEST_TARGET_BRANCH_NAME}
+        -e CMAKE_BUILD_PARALLEL_LEVEL=$(nproc)
+        cpp_tango
+        .travis/check-ABI-API-compliance.sh || true
+  after_script:
+    - docker stop cpp_tango
+    - gzip -r compat_reports
```
bourtemb: If we don't gzip the compat reports, it might be possible to browse them more easily online?: https://gitlab.com/tango-controls/test_gh_import/cppTango/-/jobs/956206341/artifacts/browse/compat_reports/libtango/fe934f0ccf09980e679e1f87b0bd042573a31a9b_to_46b5052e3f4576e0f5274c5d6cbbba58bde6a897/
What do you think?

mliszcz: Yes, compression is not needed here. I guess Gitlab could display file contents directly in the artifacts UI: https://gitlab.com/gitlab-org/gitlab/-/issues/211517

I just pushed a commit that removes compression. Let's wait and see how it looks like: https://gitlab.com/tango-controls/test_gh_import/cppTango/-/jobs/960823927

INSTALL.md
```diff
@@ -71,6 +71,7 @@ PKG_CONFIG_PATH="/usr/local/libzmq:/usr/local/omniORB" cmake ..
 | `CPPZMQ_BASE`                |                                        | cppzmq installed path
 | `IDL_BASE`                   |                                        | tangoidl installed path
 | `OMNI_BASE`                  |                                        | omniORB4 installed path
+| `TANGO_ENABLE_COVERAGE`      | `OFF`                                  | Instrument code for coverage analysis (Requres CMake 3.13+)
```
bourtemb:
```suggestion
| `TANGO_ENABLE_COVERAGE`      | `OFF`                                  | Instrument code for coverage analysis (Requires CMake 3.13+)
```

INSTALL.md
```diff
@@ -283,3 +284,26 @@ cmake -G "Visual Studio 15 2017 Win64" -DCMAKE_INSTALL_PREFIX=install -DCMAKE_BU
...
+   ```
+2. Run any relevant tests that contribute to the code coverage:
+   ```
+   ctest ...
+   ```
+3. Generate the report with a tool of choice, e.g. `gcovr` can generate reports
+   in many formats. Note that usually one is interested only in the library
+   code and may want to exclude any test code from the report.
+   ```
+   gcovr --filter '^cppapi/' --filter '^log4tango/' --html-details --output coverage.html
```
bourtemb: We should probably suggest to exclude log4tango/tests directory too in the coverage report.

.gitlab-ci.yml
mliszcz: We can indeed browse html reports directly in the gitlab UI. Thanks for this comment.

https://tango-controls.gitlab.io/-/test_gh_import/cppTango/-/jobs/960823927/artifacts/compat_reports/libtango/fe934f0ccf09980e679e1f87b0bd042573a31a9b_to_a0e273c48570abc5c7820c3ca3dc4b05c3594e29/compat_report.html

INSTALL.md
mliszcz: Done in 13fbbd3e04a0604763866b2ac36394744d63923d

t-b (approved):
Looks good!

I think we should the cleanup (ditching travis in paths and names, remove coveralls/bintray?/sonar? support) once we have migrated to gitlab.

mliszcz:
Thanks @t-b! For cleanup tasks I created an issue: https://github.com/tango-controls/cppTango/issues/821
Please use that one to report any things we should remember about after the migration.

bourtemb (approved):
Looks good to me too!
Thanks @mliszcz

mliszcz:
Thank both of you for the review. Today or tomorrow I'm going to port these changes to 9.3-backports and open a similar PR.

mliszcz:
@t-b, @bourtemb should I merge that one? Or after the migration? Note that it is disabling two tests.

bourtemb:
@mliszcz
For me, it's fine to merge it now.

PR #830, opened 2021-01-20
Prepare for archiving

t-b:
@mliszcz @bourtemb Before archiving we should tell visitors where we can now be found.

```
$cat Readme.md 
⛵ We have moved to a new home, see https://gitlab.com/tango-controls/cppTango.
```


mliszcz:
Why do you want to remove all the code? Isn't it enough to just put a note on top of the readme and archive the repo?

t-b:
@mliszcz Because people should not checkout that repo anymore. By making their build fail, they are not stuck with an old version which will never be updated.

bourtemb:
But it's easy to change the remote on git.
I guess we can still allow them to clone it and simply inform on the Readme that they should update their remote git repos on their local copies.

t-b:
> But it's easy to change the remote on git.
> I guess we can still allow them to clone it and simply inform on the Readme that they should update their remote git 
> repos on their local copies.

They can still clone it. The latest commit will just not hold the code anymore. Everyone should change the remote in git. But by removing the code we enforce that rather than just tell people.

t-b:
@bourtemb Take for example your conda package. The next time it is built it would still be built from the old repo without you noticing when we just add a note. But with my approach we would fail the build and the user automatically will notice.

Building from tags and commits does still work as I don't rewrite history.

mliszcz:
@t-b I'm not sure... It may break links pointing to the files in this repo, it makes browsing the old code less convenient (few clicks are needed to see previous commit), etc.. I see no (technical) reason for preventing people from using the old repo. The change was well communicated. Whoever wants to switch to the new repo, can do so (change their CI pipeline or whatever). Those who do not want or do not care about the new code will be forced to do something.

It's just my personal opinion. I think that how this repo (and all other repos under the tango-controls organization, for the sake of consistency) should look like after the migration is a topic for discussion in the kernel meeting.

lorenzopivetta:
IMHO, I'd definitely prefer not to keep two copies of the code. My 2c.

t-b:
@mliszcz We don't want to have users being stuck on old versions or? There is also no reason to use this repo and/or github site anymore.

mliszcz:
> @mliszcz We don't want to have users being stuck on old versions or? There is also no reason to use this repo and/or github site anymore.

As long as they are not coming with bug reports and support requests for the code that is no longer maintained, I am not much concerned about who uses which version.

I do not have any strong opinion what should be done with this repo. I'm also ok with removing the code. Maybe consider bringing this topic to the kernel meeting. I do not feel in position to judge how this repo should look like post-migration. But if code is removed, I think it should be done consistently across all repos.

bourtemb:
> 
> 
> @bourtemb Take for example your conda package. The next time it is built it would still be built from the old repo without you noticing when we just add a note. But with my approach we would fail the build and the user automatically will notice.
> 
> Building from tags and commits does still work as I don't rewrite history.

Well, the conda packages are usually building from tags.
But I understand your position and indeed the users which are relying on GitHub repo for some of their automatic tasks will notice changes faster with your approach.
I personally didn't think about deleting the files. It didn't come to my mind because I thought that a good Readme and good communication should be enough and all the discussions I had on that migration topic didn't mention that, I think (but my memory is not always reliable).
I remember at some point it was said that we could eventually in the future setup GitHub to be a read-only mirror repo of the official repo because it is "important" to be on Github. Having something on Github with only a Readme file visible on the main is a bit strange and does not give a wonderful impression about the project at first sight but well, might be acceptable if the Readme is clear.

cpascual:
Personally I think it is better to not change the README nor remove anything **in the `tango-9-lts` branch**, but instead  
create a **new branch** called "moved-to-gitlab" (containing only the README with the migration message) and make it the default branch in github just before archiving the project.

In this way, people cloning the default branch will get the clear message, but people depending on GH infrastructure for whatever reason will not be forced to act immediately.

Then, after some time, of course the tango-9-lts branch could be removed from GH to finally force the change.

IMHO this is cleaner and also more flexible

tjuerges:
In my opinion @t-b Thomas makes a valid and a very good point.  The problem is that the authoritative repository has moved to gitlab but it did not move clear cut with a new release but in between releases.  Just changing the readme will not do a good job because human beings are creatures of habit.  They will clone the code from the well established github repo or they will find the github repo link somewhere on the internet.  It just takes time to update all the links.

I know that gitlab can mirror github repos.  Hoping that github also has this feature, a solution could be to make the github repo read-only and mirror it from the repo on gitlab. By this the old links would not turn invalid, people would have time to make the migration, too, and we would have time to modify all the links floating around.

beenje:
As some mentioned, I think the best would be to keep GitHub as a read-only mirror. It's possible to do that from GitLab: https://docs.gitlab.com/ee/user/project/repository/repository_mirroring.html#setting-up-a-push-mirror-from-gitlab-to-github

beenje:
A few extra information if we go the mirror way:
- the repository won't be read-only. Members who have access could still push to the repo (we should only keep the tango-controls owners). It's their responsibility to not do that.
- If someone pushes to GitHub, the next sync from GitLab will fail, so we'll notice.
- Issues can be disabled in GitHub: https://docs.github.com/en/github/managing-your-work-on-github/disabling-issues
Note that it will hide the "issues" tab. Existing issues won't be available anymore (even with the url). I don't think that's a problem as issues will be migrated to GitLab.
- I don't think PRs can be disabled. It would just be a matter of redirecting the user to GitLab in the comments. This could be automated with a bot if there are too many, but I guess it's unlikely.

As an example I have this personal repo https://github.com/beenje/gidgetlab set as a mirror of https://gitlab.com/beenje/gidgetlab

bourtemb:
> 
> As an example I have this personal repo https://github.com/beenje/gidgetlab set as a mirror of https://gitlab.com/beenje/gidgetlab

I'm currently doing a test with https://gitlab.com/tango-controls/test_gh_import/cppTango being mirrored to https://github.com/bourtemb/cppTango-mirror-test

I selected the option to mirror only the protected branches. Only tango-9-lts branch seems to be mirrored.

We will have to be careful to update the protected branches setting on gitlab after the migration because we would like 9.3-backports branch to be protected too there. The protection branches setting does not seem to be automatically migrated.

The main branch in the mirror got created when I created the repo before on Github.

cpascual:
>Issues can be disabled in GitHub: Note that it will hide the "issues" tab. Existing issues won't be available anymore (even with the url). I don't think that's a problem as issues will be migrated to GitLab.

This may be annoying because it would break existing links in other places (e.g., another project linking a comment in a cppTango issue). For this alone I would prefer the "archive" option (which leaves the issues still visible but read-only).

beenje:
> > Issues can be disabled in GitHub: Note that it will hide the "issues" tab. Existing issues won't be available anymore (even with the url). I don't think that's a problem as issues will be migrated to GitLab.
> 
> This may be annoying because it would break existing links in other places (e.g., another project linking a comment in a cppTango issue). For this alone I would prefer the "archive" option (which leaves the issues still visible but read-only).

True. Issues don't have to be disabled if we mirror the repo, but they won't be read-only, which might be annoying.
Archiving has some benefits.

Anyway, I think the code should not disappear from GitHub. I'd vote for mirroring or archiving.

t-b:
I think we should not use mirroring. As we've decided to move we should move in a clear manner. There is no point in tango people still using github for whatever reason when we are not active on github anymore.

My suggestion:
- Apply a PR like this one to all active branches (tango-9-lts, 9.3-backports, master)
- Archive the repository thus making it read only

The workflow-breaking part here is *on* purpose as we want people to notice ASAP so that they can find us at our new home.

This approach has also the advantage that we don't need to invest any resources in github anymore.

tjuerges:
> The workflow-breaking part here is _on_ purpose as we want people to notice ASAP so that they can find us at our new home.

This is a very good point, @t-b Thomas.  Would it make sense to align this with a or better the next release if that is feasible?

lorenzopivetta:
Since we decided to move to GL, I cannot see any benefit on maintaining the code on GH... except some additional work, and, maybe, some unforeseen issues from time to time because people tend to be lazy and forget the changes. Chances are we'll have to spend time understanding questions or bug reports just because someone's using the wrong repo. And we do not have the time, nor the manpower for this in the Tango community. Thus, my vote goes to completely removing tango-controls from GH.

t-b:
@tjuerges 
> Would it make sense to align this with a or better the next release if that is feasible?

We will have a 9.3.5 release from 9.3-backports branch soon. But I actually wanted to make that from gitlab already.
For the tango-9-lts branch the plan is to have a beta release in the summer.

andygotz:
@t-b , @tjuerges and @lorenzopivetta I agree with your proposal to remove the projects from GH once tey have been moved successfully to GL to avoid confusion and not have to maintain two repositories.

bourtemb:
I think I would prefer going for the archive option on Github with a clear README.
People won't be able to create/comment Pull Requests nor issues any longer on Github and will be notified that the repo is read only. They can't even push to it when it is in archive mode, even if they are owner.
It's true that the suggestion from @t-b to also remove the code in the latests commits and to archive afterwards is probably the most effective to make people aware of the change.
It might be a bit brutal for some people who might need some time to do some updates on their side?
If we decide to remove the code as Thomas is suggesting before archiving, I would do that in several steps.
First I would archive the repo as it is now.
Then I would do the migration to Gitlab.
We can wait for some time to give time to people to update links here and there and to update their automated routines still relying on this github repo code.
Then I would unarchive the github repo temporarily, just the time to merge a PR like this one.
Then I would archive the github repo again and forever.

cpascual:
@lorenzopivetta 
> removing tango-controls from GH

If you mean actually removing the repos, I do not think it is good, because it would break links in forums, other projects, etc. Plus it does not save any effort compared to leave it archived.

So I definitely would go for the archiving option.

Regarding whether pushing a "git rm" on the first day (as @t-b ) proposes or wait a bit (as  @bourtemb) suggests, I would be ok with either, but my preference is in the intermediate solution that I already proposed: leave (for a while) the `tango-9-lts` *as it is* and create a new branch called `moved-to-gitlab` containing only the README with the migration message) **and make it the default branch in github** just before archiving the project.

lorenzopivetta:
@cpascual

> If you mean actually removing the repos, I do not think it is good, because it would break links in forums, other projects, etc.

Do we really want any reference in place in forums, projects etc. to keep pointing to the old GH repo? IMHO, that's really inconvenient. I mean, regardless to what we decide to do during the transition phase, I still see no point in keeping the old stuff there, just to bounce people to a different site. And, if I understand it correctly, you still have to go through the README to understand what's happening...

bourtemb:
> > If you mean actually removing the repos, I do not think it is good, because it would break links in forums, other projects, etc.
> 
> Do we really want any reference in place in forums, projects etc. to keep pointing to the old GH repo? IMHO, that's really inconvenient. I mean, regardless to what we decide to do during the transition phase, I still see no point in keeping the old stuff there, just to bounce people to a different site. And, if I understand it correctly, you still have to go through the README to understand what's happening...

@lorenzopivetta , I think there will always be a place (forum, websites we do not own, e-mails, publications, ...) somewhere pointing to github so I think it is good to keep something here just to bounce people to the new site. We will not edit the old messages on the forum for instance.

I think a good compromise would be to do the following:
1/ Archive the Github repository just the time of the migration (few minutes).
2/ Once the repository is correctly migrated to Gitlab, unarchive temporarily the Github repository and create a new branch called moved-to-gitlab containing only the README with the migration information message. Make this moved-to-gitlab branch the default branch on github (so people cloning the repo will see this branch by default and people coming on the main github repo web page will see this README). We could also adapt the README in the main branches (tango-9-lts, 9.3-backports and master to put the migration information message at this stage).
3/ Archive the Github repository again
4/ Send an e-mail on the Tango mailing list informing about the migration and warn about the fact that the code on the Github repository is no longer maintained there and is no longer the official one. Inform people relying on github repo for some automatic things to use Gitlab instead. Give a deadline (2 months?) for them to adapt their scripts, links in documentation, etc...
5/ After this deadline, unarchive temporarily Github repo and do a git rm in the main branches, which will force people still relying on github repo to move to gitlab. And finally archive definitely the Github repo, which will be there only to bounce people to the new repository location.

cpascual:
@lorenzopivetta , 

> Do we really want any reference in place in forums, projects etc. to keep pointing to the old GH repo? IMHO, that's really inconvenient.

I suspect that you are thinking on *code*, but I am talking also about issues, PRs and their comments. When we migrate, the project internal cross-references will also be migrated, but any link from another project will still be pointing to the GH issue. Think, for example, of a doc from an institute mentioning a workaround that is explained in comment of an issue of cppTango: I see a lot of value (and no harm) in not breaking that link. IMHO, the only thing that we should (eventually) break is code pulls (to give a very clear warning that the up-to-date code is in GL), and this can be done by archiving and pushing a commit that removes everything but the README.

ajoubertza:
I like @cpascual 's suggestion of the "moved-to-gitlab" default branch.   And archiving the repo on github.  I think that is the simplest solution that doesn't break anything, and makes the move obvious to anyone landing on the default github page.

cpascual:
According to today's decision of the Tango kernel meeting, the migration will be done Tomorrow with [this generic procedure](https://github.com/tango-controls/gitlab-migration-tools#add_moved_branchsh)

